PR #2, opened 2017-03-15
Make charges floats. Propagate rxn name to molecules for ie_rxns.

loriab:


dgasmith:
Whoops, so this is fine. Can you update the hashes because we went to floats?

dgasmith:
Just for the python testing which will be different from psinet. Run 'py.test -v' to observe the required hashes. 

-Daniel Smith

PR #24, opened 2018-08-21
CI: Updates conda-envs to mostly from default channels

### LINKED ISSUE #23: Conda env channel changes
Conda-forge is the primary channel used for all environments; this is not needed as Anaconda has pulled the majority of the packages back upstream. Also, conda-forge NumPy and Psi4 has conflicts as cf uses OpenBLAS. This does not appear to be an issue, but no reason to have multiple BLAS libraries in the possible conflict.

dgasmith:
## Description
As described, env builds should pull from either default or Psi channels. Fixes #23.

## Status
- [x] Ready to go

### DIFF
--- a/devtools/conda-envs/dask.yaml
+++ b/devtools/conda-envs/dask.yaml
@@ -1,7 +1,6 @@
 name: test
 channels:
-  - conda-forge
-  - psi4/label/dev
+  - psi4
 dependencies:
     # Dask testing environment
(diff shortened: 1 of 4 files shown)


dgasmith:
@loriab Can you look this over?

A browse through `conda list` shows defaults + psi channel for everything which is excellent. Also cut down ~2 minutes on the build time.

PR #32, opened 2018-08-28
Crank: Track optimizations

dgasmith:
## Description
Correctly tracks all optimizations procedures spawned by a Crank service and reduce the cost of the test by an order of magnitude by using RDKit.

## Status
- [X] Ready to go


qcfractal/interface/orm/crank_orm.py
```diff
@@ -1,12 +1,13 @@
 """
-A ORM for Crank
+A ORM for TorsionDrive
```
Lnaden: If there is a `torsiondrive_orm.py` file, does that make this one redundant?

dgasmith: It does, fixed the history so that it correctly shows a name change and removes the redundant file.

PR #35, opened 2018-08-29
Db update

### LINKED ISSUE #33: Database Update Option
Currently, once a database is added from the Client, there is no way to update the database. The `Database.save` member function does have a `force` option that does not currently work. `force` should be fixed, and a test added.

dgasmith:
## Description
Allows client Databases to overwrite data on the server. Closes #33.

@dnascimento13 

## Status
- [x] Ready to go

### DIFF
--- a/qcfractal/db_sockets/mongo_socket.py
+++ b/qcfractal/db_sockets/mongo_socket.py
@@ -492,7 +492,7 @@ class MongoSocket:
 
 ### Mongo database functions
 
-    def add_database(self, data):
+    def add_database(self, data, overwrite=False):
         """
         Adds a database to the database.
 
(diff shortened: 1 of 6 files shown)


dgasmith:
We "overwrite" the database so that the table ID stays the same, but all other data is changed (effectively, a clear and reset).

Lnaden:
Do you think it would be helpful to add a formal reset function of some kind?

dgasmith:
In this context, what are would reset mean?

Lnaden:
Just a thought that if we are going through the process of clear and reset, then would it make sense to provide a formal reset function for the socket? It may also be such a rare thing you would only want to do while also adding a new database that its fine to leave it as is.

dgasmith:
Ah so reset would be delete in this context? Delete is something I haven't formally figure out yet. Somewhere there is a line where it should be possible, but not easy I think.

Lnaden:
> Somewhere there is a line where it should be possible, but not easy I think.

I agree. I think a reason I asked here is this seemed very easy to do, but still has to be intentional. This is fine for now.

> Ah so reset would be delete in this context? 

Yes, sorry. I did mean "delete" instead of "reset"

PR #40, opened 2018-08-31
Add initial Example test

Lnaden:
## Description
This adds the bash script to run the Fireworks example, and then adds
a pytest file to execute and validate. Its a bit crude, but I am open
to suggestions on fixes.

## Questions
Known issues:
- [ ] I'm not sure we want to run the examples on Travis all the time, they
  can be a bit slow, and I imagine as the examples get more complex,
  they will. I have not added the travis hooks to do this yet for that
  reason
- [ ] The py.test script relies on subprocess output and very output
  sensitive regex matching to assert values are returned correctly
- [ ] Cleaning up the background process jobs is fragile, but it was either
  that or never clean them up and/or flood the STDOUT with MongoDB and
  Fireworks spam.

I'm open to suggestions for these issues, but I can confirm it works
locally.
## Status
- [ ] Ready to go

### DIFF
--- a/examples/fireworks_server/README.md
+++ b/examples/fireworks_server/README.md
@@ -5,7 +5,7 @@
 This example assumes you have a install configure to the setup in the
 [fireworks.yaml or openff.yaml conda env](../../devtools/conda-envs).
 
-You effectivley need MongoDB and Fireworks installed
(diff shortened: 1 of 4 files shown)


Lnaden:
You are okay with the examples not running on Travis?

PR #43, opened 2018-09-04
Merges service result with procedures

dgasmith:
## Description
We previously had three types of compute pathways:
 - results, started in the queue, and were deposited in the results table.
 - procedures, started in the queue, and deposited their results in both the results and procedures table (geometry optimizations as an example).
 - services which iteratively created new tasks for the queue.

This above had several complexities:
 - The result and procedures table only contained complete results while the services table could contain both complete and incomplete results.
- The services and procedures table could have nearly identical jobs (geometry optimizations run all on node, or new geometries spawned from the server as a service)

This PR makes the following changes:
 - Changed the queue names to `task_queue` and `services_queue` to better reflect their purpose.
 - `services_queue` now deposits their results into the procedures 

This PR also adds geometry optimization ORM.

## Status
- [x] Ready to go

### DIFF
--- a/qcfractal/db_sockets/mongo_socket.py
+++ b/qcfractal/db_sockets/mongo_socket.py
@@ -718,6 +718,12 @@ class MongoSocket:
             modified_count += result.modified_count
         return (match_count, modified_count)
 
+    def del_services(self, values, index="id"):
+
+        index = _translate_id_index(index)
+
+        return self._del_by_index("service_queue", values, index=index)
+
 ### Mongo queue handling functions
 
     def queue_submit(self, data, tag=None):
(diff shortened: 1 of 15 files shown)


dgasmith:
This pull request **introduces 1 alert** when merging 01ea130835190111572e397028814681b6fb2d3b into 42cf530355145d00b87d02484f01e1a6a9f61271 - [view on LGTM.com](https://lgtm.com/projects/g/MolSSI/QCFractal/rev/pr-bf1f966a8fb391c43344e18e26b0da953cc8ffb2)

**new alerts:**

* 1 for Unused import

---

*Comment posted by [LGTM.com](https://lgtm.com)*

dgasmith:
Yet to be enumerated, I think we need a dual on-node and as-a-service example to delineate the processes.

PR #46, opened 2018-09-06
Fixes codecov reporting stats

dgasmith:
## Description
Something changed so that `codecov` now looks at omited coverage lines as well. This PR should ignore these files.

## Status
- [ ] Ready to go


Lnaden:
Do you also want to add versionneer to the ignored files?

dgasmith:
The base `versioneer` is not hit according to codecov. Id like to leave it out unless it becomes an issue.

PR #47, opened 2018-09-06
Skipif for Mongo

### LINKED ISSUE #18: Tests require active Mongod
The testing structure currently assumes there is an active Mongod on the standard `localhost:27017`. This assumption could be problematic as testing takes ~30 seconds per Mongo access before the query times out resulting in a potential ~ 30-minute test failure. Several options are available:
 - Create a new fixture that checks for a live mongo instance and wraps all mongo tests.
 - Fail if no DB backend is detected.
 - Spawn a temporary Mongod instance if no live instance is identified.

dgasmith:
## Description
Skips tests that require an active mongo connection. `pytest.mark.skipif` does not apply to fixture so an explicit call is required.

Closes #18.

## Questions
- [x] We skip 40/53 tests currently without a Mongo connection. Not sure if we should fail or not.

@dnascimento13 `py.test -v` now passes with many skips in ~5 seconds as opposed to many minutes previously.

## Status
- [x] Ready to go

### DIFF
--- a/qcfractal/testing.py
+++ b/qcfractal/testing.py
@@ -7,6 +7,7 @@ import logging
 import pkgutil
 import socket
 import threading
+import pymongo
 from contextlib import contextmanager
 
 import pytest
@@ -46,7 +47,6 @@ else:
 def has_module(name):
     return _programs[name]
 
-
 # Add a number of module testing options
 using_fireworks = pytest.mark.skipif(has_module('fireworks') is False, reason=_import_message.format('fireworks'))
 using_dask = pytest.mark.skipif(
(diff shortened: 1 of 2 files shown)


Lnaden:
Why can we not add this as a fixture?

> We skip 40/53 tests currently without a Mongo connection. Not sure if we should fail or not.

I think we should fail the tests outright without another fallback storage layer. Since the package does not really function without a storage layer, then it does not make sense to test without a critical part of the package working.

dgasmith:
> Why can we not add this as a fixture?

Can you elaborate this?

Lnaden:
Whoops. I was thinking adding this as part of a `@pytest.mark.skipif(...` decorator. Its being added as part of a fixture in what you did. I missed it on first pass

dgasmith:
Not sure I completely agree that it should fail outright. It is still covers quite a bit (45% coverage) and the case where a user simply runs `py.test --pyargs qcfractal`.

PR #49, opened 2018-09-12
Fix syntax in azure yaml and make hidden

Lnaden:
## Description
As it says on the tin. If I am following correctly, this should fix the pipeline file and make the pipeline manageable from the Azure side 

## Status
- [x] Ready to go

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -5,7 +5,8 @@
 
 jobs:
 
-- job: 'Test'
+- job: 'Testing'
+  displayName: 'Unit Testing'
   pool:
     vmImage: 'Ubuntu 16.04'
   strategy:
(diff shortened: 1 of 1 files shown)


.azure-pipelines.yml
```diff
@@ -16,26 +17,32 @@ jobs:
...
     # Create test environment for package
     - conda env create -n test python=$PYTHON_VER -f devtools/conda-envs/openff.yaml
     - source activate test
-
     # Build and install package
-  - python setup.py develop --no-deps
+    - python setup.py develop --no-deps
+    displayname: "Conda Env Creation"
 
+  # Display MongoDB, Conda Environment, Python version
```
dgasmith: We need to spin up a MongoDB, I don't think this is done?

Lnaden: Its not. But I think we have to get the YAML working once first and then we can work from there.

Lnaden: And by working, I mean syntax correct, not everything passing correct.

PR #50, opened 2018-09-12
Fix more Azure script syntax issues

Lnaden:
## Description
If everything is configured correctly, this should work the pipeline system should pick up on this PR 

Might be some tinkering required and I will work with a feature branch if I have to.

## Status
- [ ] Ready to go

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -28,31 +28,30 @@ jobs:
 
     # Additional info about the build
   - script: |
-    - uname -a
-    - free -m
-    - df -h
-    - ulimit -a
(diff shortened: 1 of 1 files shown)


Lnaden:
Looks like the hook is detecting changes, which is a good start

Lnaden:
@dgasmith I think this can be merged, even though we are not using this file right now in Azure. There are some other changes we will need to make eventually to make it work on Azure (namely getting MongoDB working there), but this PR does make required syntax changes we will need eventually anyways.

We can override the Azure failure since the last change since I disabled that hook and the rest are passing.

PR #53, opened 2018-09-17
Results hash index

dgasmith:
## Description
This PR makes the following changes:
- Adds a `hash_index` to the results table to allow add_compute returns return a single index (hash) rather than this return having multiple meanings depending on context (molecule_id for single runs, hash for procedures).
 - Reworks database `get_generic` to allow arbitrary queries for internal usage.
 - Blank hooks are now correctly ignored.

## Status
- [x] Ready to go

### DIFF
--- a/qcfractal/storage_sockets/mongo_socket.py
+++ b/qcfractal/storage_sockets/mongo_socket.py
@@ -38,6 +38,9 @@ def _str_to_indices(ids):
 
 
 def _str_to_indices_with_errors(ids):
+    if isinstance(ids, str):
+        ids = [ids]
+
     good = []
     bad = []
     for x in ids:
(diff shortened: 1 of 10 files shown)


qcfractal/storage_sockets/mongo_socket.py
```diff
@@ -257,66 +256,58 @@ def _del_by_index(self, table, hashes, index="_id"):
 
         return (self._tables[table].delete_many({index: {"$in": hashes}})).deleted_count
 
-    def _get_generic_by_id(self, ids, table, projection=None):
+    def _get_generic(self, query, table, projection=None, allow_generic=False):
```
Lnaden: Can we add a test to sift through more complex queries?

dgasmith: I am not sure we should be completely happy with the current querying structure and would like to delay full tests there until we are a bit more firm that the current function is what we should be expanding on.

The names `id` and `hash` are appropriate throughout the code I believe. As `id` still refers to `table id` where `hash` refers to input hashing.

Lnaden: But are we not using the `hash` as the `table id`, or is that not in yet?

dgasmith: Not yet, I think we need to do more validation before that change.

Lnaden: Okay, but you want to get this merged in all the same to start the process and we can add in the validation later?

dgasmith: The querying structure changes and the `id->hash` conversion are two different pieces of the PR. This PR does add a `hash_index` to the results table but does not yet operationally effect anything. This is good to merge now so that we can explore hash usage IMO.

Lnaden: Good enough reasoning. I'll merge it in once I clean up the pipeline.

PR #64, opened 2018-09-27
 Ensure torsion drive completed tasks pull from their queue id

### LINKED ISSUE #62: Better handle completed torsion drive task result checking
As described in #61, the check for previous completed torsion drive scans searches through the procedure table, not the tasks table like the other jobs do. This will cause a problem since the logic for `iterate` assumes the `job_map` IDs are on the `tasks_queue` table. However, this has not come up in testing nor in production yet.

Things needed:
* Make a test to hit this code
* Improve the logic to make sure all IDs in the `job_map` are either from the same table, or all use a locator object to fetch across the MongoDB

Lnaden:
## Description
This makes sure the `job_map` for new Torsion Drive service tasks
reference the same table in the storage layer rather than different
layers.

A new bug was identified for a rare corner case where a secondary
service submits jobs after the initial query from a primary service, but
before the tasks from the primary are submitted to the queue. We have
elected not to fix that in this change, but will raise a helpful error
if someone ever actually gets it.

Fixes #62 

Credit to @dgasmith for coming up with the main fix.

Once this is merged in, we should make a pair of new issues:
1. Fix the duplicate entry bug mentioned above (easier)
2. Generalize the service submission process since we will have to do that anyways going forward so as to not have to fix this for every new service (harder)

## Status
- [x] Ready to go

### DIFF
--- a/qcfractal/testing.py
+++ b/qcfractal/testing.py
@@ -9,6 +9,7 @@ import socket
 import threading
 import pymongo
 from contextlib import contextmanager
+from collections import Mapping
 
 import pytest
 from tornado.ioloop import IOLoop
@@ -47,6 +48,7 @@ else:
 def has_module(name):
     return _programs[name]
 
+
 # Add a number of module testing options
 using_fireworks = pytest.mark.skipif(has_module('fireworks') is False, reason=_import_message.format('fireworks'))
 using_dask = pytest.mark.skipif(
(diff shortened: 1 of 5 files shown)


qcfractal/testing.py
```diff
@@ -58,6 +60,17 @@ def has_module(name):
 using_unix = pytest.mark.skipif(os.name.lower() != 'posix', reason='Not on Unix operating system, '
                                                                    'assuming Bash is not present')
 
+
+# More complex than a simple top-level merge {**x, **y} which does not handle nested dict
```
dgasmith: Can you add this line as a docstring when you get the chance.

Lnaden:
There we go, this should fix it

PR #65, opened 2018-09-29
Dataset docs

dgasmith:
## Description
While constructing dataset docs I found that the current way of obtaining collections was a bit awkward `Dataset.from_server("S22", client)`. Collections can now be pulled directly from a client `client.get_collection("dataset", "S22")`.

- Adds a string form for Collections.
- Allows collections to be built straight from clients.
- The client can now list all available collections through a `list_collections` function.
- Starts some dataset docs, but still quite a ways to go.

## Status
- [x] Ready to go


docs/source/collection-dataset.rst
```diff
@@ -0,0 +1,110 @@
...
+
+.. code-block:: python
+
+    ds = ptl.collections.Dataset("my_dataset", dtype="rxn")
+
+New reactions can be added by providing the linear combination of molecules
+required to compute the desired quantity. When the Dataset is queried these
+linear combinations area automatically combined for the caller.
+
+    ds = ptl.collections.Dataset("Atomization Energies", dtype="ie")
```
Lnaden: Missing a `..code-block::` directive here

docs/source/collection-dataset.rst
```diff
@@ -0,0 +1,110 @@
...
+    N 0.0 0.0 0.0
+    """)
+
+
+    ds.add_rxn("Nitrogen Molecule", [(N2, 1.0), (N_atom, -2.0)])
+
+A given reaction can be examined by using the ``get_rxn`` function. We store
+the ``molecule_hash`` followed by the coefficient.
+
+    json.dumps(ds.get_rxn("Nitrogen Molecule"), indent=2)
```
Lnaden: Missing a `..code-block::` directive here

docs/source/portal-molecule.rst
```diff
@@ -23,21 +23,68 @@ Molecules can be created via NumPy, Psi4-like molecule strings, and QCSchema JSO
...
+   | comment                                       | Any additional comment one would attach to the molecule.                                                                 | string                      |
+   +-----------------------------------------------+--------------------------------------------------------------------------------------------------------------------------+-----------------------------+
+   | connectivity                                  | (nat, ) list describing bonds within a molecule. Each element is a (atom1, atom2, order) tuple.                          | array[array]                |
+   +-----------------------------------------------+--------------------------------------------------------------------------------------------------------------------------+-----------------------------+
+   | provenance                                    | #/definitions/provenance                                                                                                 | object                      |
```
Lnaden: Not sure if this was suppose to link somewhere like a `:term:` directive, but it looks odd. If its placeholder, thats also fine for now

docs/source/portal-molecule.rst
```diff
@@ -23,21 +23,68 @@ Molecules can be created via NumPy, Psi4-like molecule strings, and QCSchema JSO
...
+
+A full description of the various specifications can be found below:
+
+Required Keys
++++++++++++++
+
+   +-----------------------------------------------+--------------------------------------------------------------------------------------------------------------------------+-----------------------------+
+   | Key Name                                      | Description                                                                                                              | Field Type                  |
+   +===============================================+==========================================================================================================================+=============================+
+   | symbols                                       | (nat, ) atom symbols in title case.                                                                                      | array[string]               |
```
Lnaden: Small nuance: but the "nat" took a bit for me to figure out it meant "N atom", I don't know if there is a better way to show this since "nat" is phonetically pronounceable, mabye "N-at?" Fine for now, something to think about.

qcfractal/interface/client.py
```diff
@@ -5,6 +5,10 @@
 import requests
 import yaml
 
+from collections import defaultdict
+
+from .collections import collection_factory
```
Lnaden: Well, this is technically correct, albeit a bit confusing. This might be just what we have to do to avoid name collision here.

qcfractal/interface/client.py
```diff
@@ -241,15 +245,37 @@ def add_options(self, opt_list, full_return=False):
 
     ### Collections section
 
-    def get_collections(self, collection_list, full_return=False):
+    def list_collections(self, collection_type=None):
```
Lnaden: Thinking about de-duplication, can you have this function just call `get_collection` or have both of them call a hidden function?

qcfractal/interface/collections/biofragment.py
```diff
@@ -114,3 +115,4 @@ def submit_torsion_drives(self, options_set, torsions):
                 submissions.append(ret)
         return submissions
 
+collection_utils.register_collection(BioFragment)
```
Lnaden: May as well add the empty line here so we dont have to push a change to portal later which gets built on the standalone

qcfractal/interface/collections/collection.py
```diff
@@ -33,7 +30,7 @@ def __init__(self, name, **kwargs):
         """
 
         self.client = kwargs.pop("client", None)
-        if (self.client is not None) and (not isinstance(self.client, FractalClient)):
+        if (self.client is not None) and not (self.client.__class__.__name__ == "FractalClient"):
```
Lnaden: Is this safe? Wouldn't the `isinstance` be a much safer call than any class named "FractalClient"? This would also make it so subclasses of FractalClient (should remote users chose to make one) won't work here.

qcfractal/interface/client.py
```diff
@@ -241,15 +245,37 @@ def add_options(self, opt_list, full_return=False):
...
+        payload = {"meta": {}, "data": [(collection_type.lower(), collection_name)]}
         r = self._request("get", "collection", payload)
 
         if full_return:
             return r.json()
         else:
-            return r.json()["data"]
+            # If nothing found
+            if len(r.json()["data"]):
+                return collection_factory(r.json()["data"][0], client=self)
```
Lnaden: Should this thrown an error if there are more than 1 collection? Otherwise there will be ambiguous returns which may or may not be deterministic.

qcfractal/interface/collections/dataset.py
```diff
@@ -701,3 +701,6 @@ def __getitem__(self, args):
             A view of the underlying dataframe data
         """
         return self.df[args]
+
+
+register_collection(Dataset)
```
Lnaden: blank line

qcfractal/interface/collections/openffworkflow.py
```diff
@@ -118,3 +119,4 @@ def get_data(self):
 
         return ret
 
+collection_utils.register_collection(OpenFFWorkflow)
```
Lnaden: blank line

qcfractal/tests/test_portal.py
```diff
@@ -44,15 +44,15 @@ def test_options_portal(test_server):
 
 def test_collection_portal(test_server):
 
-    db = {"collection": "TorsionDrive", "name": "Torsion123", "something": "else", "array": ["54321"]}
+    db = {"collection": "torsiondrive", "name": "Torsion123", "something": "else", "array": ["54321"]}
```
Lnaden: This brings up an interesting point. Should we make db calls case-insensitive for things like the collection name? We would have to store a map of either all Lcase or Ucase names to the class objects, but I can see this being an issue.

qcfractal/interface/client.py
dgasmith: Not optimal, but no way around without renaming collections I believe.

qcfractal/interface/client.py
dgasmith: Duplicated 2-3 lines isn't worth writing an extra function IMO. Most of the deduplication is handled in the call function.

qcfractal/interface/collections/biofragment.py
dgasmith: Can you explain more here?

qcfractal/interface/collections/collection.py
dgasmith: Not particularly safe, but would cause circular imports otherwise.

qcfractal/interface/client.py
dgasmith: It isn't currently possible to get more than one return here at the moment due to the unique requirements of the index.

qcfractal/tests/test_portal.py
dgasmith: Yea, have been toying around with this. Currently we wrap it correctly on the client side of things, but not on the mongo side. Same issue with `name` where we might want `display_name` or similar where the case is kept.

docs/source/portal-molecule.rst
dgasmith: Add a note at the bottom

qcfractal/interface/collections/biofragment.py
Lnaden: Github is complaining about the missing blank line at the end of this file. I'd like to avoid the making a change later where we push to the standalone `QCPortal` repo which just updates empty lines. Its harmless, but since we are in here anyways

qcfractal/interface/client.py
Lnaden: This is fine then!

qcfractal/interface/collections/collection.py
Lnaden: Ah. Those. Yes, lets avoid those first

qcfractal/tests/test_portal.py
Lnaden: I think this is fine fore now, but something we can think about for future releases.

qcfractal/interface/client.py
Lnaden: I was not suggesting making a change, just noting for myself mostly that this going to be the canonical way we have to handle this

dgasmith:
Ok, all comments should be fixed up.

PR #66, opened 2018-09-29
CI: Fixes python version 

dgasmith:
## Description
Previous uses of `conda env create` were not pinning a Python version so only 3.6 was ever tested. A script has been built to overcome this issue. Issue brought up by @loriab.

Python 3.7 cannot be satisfied at the moment as neither RDKit nor Psi4 are built with Py3.7 at the moment. 

## Status
- [x] Ready to go

PR #69, opened 2018-10-02
Changelog for 0.2.0a

dgasmith:
## Description
Updates the changeling for 0.2.0a, should be the list item before minting a release.

## Status
- [x] Ready to go

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -13,31 +13,53 @@ Enhancements
 Bug Fixes
 +++++++++
 
-0.2.0a / 2018-10-DD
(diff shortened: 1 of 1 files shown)


dgasmith:
Ugh, a 10MB binary from monkey type slipped in to the repo.

Lnaden:
Looks like it came in on #66. There might be a few ways we can remove it from the repo, but it would be tricky

dgasmith:
It will require a history rewrite which may play havoc with QCPortal. My thought is to leave it, not a huge issue at the moment. But could grow if we have more of these slip ups.

PR #73, opened 2018-10-17
Start process of converting to Pydantic for validation

Lnaden:
## Description
This commit starts the pydantic validation conversion by adding
the BaseModel from Pydantic to the Collections, and reworking
the collections to not rely on an abstract method + extra hidden
properties to infer data model and validation.

This is not the complete implementation, but should get it started.

## Todos
Outstanding TODOs (can be completed in separate PRs, this is stable now):
* [ ] Convert the calls from interface <-> client to handle JSON string instead of dict
* [ ] Add Server-side validation of the incoming JSON blobs (requires previous step)
* [ ] Ensure the validation of Collection DataModels is correctly specific, especially on Dict models
* [ ] Double check the OpenFFWorkflow data model for correct defaults
* [ ] Double check ALL defaults are correct
* [ ] Add Pydantic validation to Procedures

## Questions
- [ ] Is `DataModel` a good class name for the internal class?
- [ ] Are we okay with relying on the default name for the `id` field of `"local"` until its assigned by some other means? This includes a few places where I check `id` against the default to see if its been assigned by the server

## Status
- [x] Ready to go


Lnaden:
This PR is now ready to go and awaiting review. Outstanding TODOs can be handled in separate PR

qcfractal/interface/collections/collection.py
```diff
@@ -33,19 +35,34 @@ def __init__(self, name, **kwargs):
...
+        This structure validates input, allows server-side validation and data security,
+        and will create the information to pass back and forth between server and client
+
+        Subclasses of Collection can extend this class internally to change the set of
+        additional data defined by the Collection
+        """
+        name: str
+        collection: str = None
+        provenance: Dict[str, str] = {}
+        id: str = 'local'
```
dgasmith: @doaa-altarawy Is there a better thing here than `id` so that we do not get clashes with Python namespace? Not terribly worried about it and keeps it nice and consistent throughout the entire stack, but thought I would ask.

qcfractal/interface/collections/collection.py
```diff
@@ -63,18 +80,16 @@ def __str__(self):
         Collection(id='5b7f1fd57b87872d2c5d0a6d', name=`S22`, client=`localhost:8888`)
         """
 
-        uid = "local"
-        if "id" in self.data:
-            uid = self.data["id"]
+        uid = self.data.id
```
dgasmith: We can now just add this to line 91 directly.

qcfractal/interface/collections/collection.py
```diff
@@ -132,36 +147,13 @@ def from_json(cls, data, client=None):
...
                            format(class_name, data["collection"]))
 
-        # Ensure all required fields are found.
-        req_fields = cls.__base_fields | getattr(cls, "_" + cls.__name__ + "__required_fields")
-        if len(req_fields - data.keys()):
-            missing = req_fields - data.keys()
+        # Attempt to build class, ensure all fields to reconstruct from JSON are present
+        req_fields = cls.DataModel.schema()['properties'].keys()
+        missing = req_fields - data.keys()
+        if len(missing):
```
dgasmith: Can we make these required and it is autovalidated via PyDantic?

dgasmith:
Need to nuke Py35 test.

qcfractal/interface/collections/collection.py
Lnaden: > Can we make these required 

Many of the keys have default values, which is something we want in places, having a default automatically makes it `optional` from a PyDantic standpoint. However, this is a special case where we want data *fully defined* when being fed back in from the server, even if that data was default value when originally created. This preserves check traps the case of rebuilding the object from existing data where some fields were omitted.

We could remove this check and allow missing fields to have their defaults filled in. Mandatory fields will still be checked by PyDantic

>  autovalidated via PyDantic?

Types and mandatory fields will still be checked, but this is a different type of check.

qcfractal/interface/collections/collection.py
Lnaden: Added in upcoming commit

qcfractal/interface/collections/collection.py
dgasmith: Good point, can you slightly enhance the comment line slightly to describe why we are approaching it this way. Not a big issue.

Lnaden: Will do!

qcfractal/interface/collections/dataset.py
```diff
@@ -530,24 +553,27 @@ def add_rxn(self, name, stoichiometry, reaction_results=None, attributes=None, o
...
+            rxn_dict["reaction_results"]["default"] = reaction_results
         else:
             raise TypeError("Passed in reaction_results not understood.")
 
-        self.data["reactions"].append(rxn)
+        rxn = Rxn(**rxn_dict)
+
+        self.data.reactions.append(rxn)
 
+        # TODO: Should this return the dict, or the PyDantic BaseModel?
```
Lnaden: @dgasmith This is the one outstanding issue I have on this. The `self.data.reactions` is a list of `BaseModel` objects, not a list of `dict`. This leads to 2 questions:

1. Should this function, `add_rxn` return the `dict` or the `BaseModel`?
2. More broadly, this may change how users interact with the `self.data.rections`, which I imagine would eventually go to the more general Python `Dataclass` from 3.7 eventually anyways, which would behave mostly the same way. Should we expect our users to start doing this now, or should we continue to only expose the `dict` for `[key]` type access rather than by attribute?

Lnaden: On a related note, I have the `get_rxn` function return a `dict`, not a `BaseModel`, which is inconsistent and we should pick one or the other. This is more a note for me but based on the answer above, will change what I do with it.

dgasmith:
Everything looks good, please update the TODO and questions above.

Should I squash and merge or will you rebase the last 4 commits to one?

Lnaden:
I'll rebase. 

I did have an outstanding question though, see the comments in my latest review for details

qcfractal/interface/collections/dataset.py
dgasmith: 1) Should return a BaseModel I think to be a bit more class focused.
2) Can you give an example? I would imagine users will want to do something like `self.data.reactions["my_cool_reaction"]`.

Will Dataclass eventually supersede pedantic?

Lnaden: > Can you give an example? I would imagine users will want to do something like self.data.reactions["my_cool_reaction"].

Right, people would *want* to do that, but they cant. They can't do that right now to be fair.

The current (`master`, not this pr) model would be `self.data.reactions[3]['name']` to access the reaction name, the new, post-pr model would be `self.data.reactions[3]`.name` because `reactions` itself is still a list. This change may force users to alter their code a bit, but its a pretty trivial change, and one for the better I think. It does make arbitrary iteration a bit more odd since you would have to make this kind of change:

```python
my_atters = ['some', 'list', 'of', 'keys']

# old style
for reaction in data.reactions:
    result = [do_something(reaction[attr]) for atter in my_atters]

# new style
for reaction in data.reactions:
    result = [do_something(getattr(reaction, atter)) for atter in my_atters]
```
Although I imagine that being rare since I think the more common action will be:
```python
for reaction in data.reactions:
    result[reaction.name] = do_something(reaction.attr1, reaction.attr2)
```
which I think is less ambiguous and does not force the programmer/reader to internalize list comprehension and looping

> Will Dataclass eventually supersede pydantic?

No, because the `Dataclass` does not support validation and coercion, its just a container. The way I view the data class is the natural evolution the dict where:
* Keys are pure string,
* Keys are immutable
to a formal named class where the `key:value`  pair system is replaced with `Class.attribute = value`. If Python ever converted the `Dataclass` to enforce strict input types based on the type hints, then I could see pydantic loosing out eventually, but I don't think they will do that because then you may as well make variable declaration require type definition as well.

If this all makes sense to you, and you are okay with the users making this change, then I will make the following changes:
* Make `add_rxn` return the `Rxn(BaseModel)` class
* Make `get_rxn` return the `Rxn(BaseModel)` class

One small caveat is we don't protect the `Collection.data` object, or any of its values at all. There is nothing stopping users from modifying the exiting fields to whatever they want, but I don't think there is a strong need to do this.

Lnaden:
Should be good to go!

Issue #76, opened 2018-10-24
Use Mongoengine as an ODM for MongoDB


doaa-altarawy:
We would like to have the option to use ODM (Object-Document mapper, like ORM) when handling MongoDB. This will move some of the business logic and validation, currently implemented in the monogo_socket, into the Mongoengine.

This addition will require some changes in the design of the indexes and collections relations in the current design to be able to handle it efficiently in the Mongoengine layer.

dgasmith:
The `task_queue` collection needs an index for `created_on`. Currently `queue_get_next` queries become very slow as the total task lists grows due to no indices there.

doaa-altarawy:
I'll add an index on the status too.

PR #78, opened 2018-10-25
WIP: Initial models for Mongoengine (ME) classes

doaa-altarawy:
## Description
Implementing issue #76 

## Todos
  - [x] Replace Mongoengine_socket methods to use ME classes
  - [ ] Track references of any table names in the code, and only manipulate the DB through the socket methods
  - [x] Ability to run Fractal-server completely using ME
  - [ ] Integartion tests

## Status
- [ ] Ready to go


qcfractal/storage_sockets/models.py
```diff
@@ -0,0 +1,259 @@
...
+        The molecule DB collection is managed by pymongo, so far
+    """
+
+    name = db.StringField()
+    molecular_formula = db.StringField()
+    molecule_hash = db.StringField()
+
+    def create_hash(self):
+        """ TODO: create a special hash before saving"""
+        return ''
```
dgasmith: Is there any way to get the full JSON document here? Should simple to do so if so.

Also should we call this `Molecule` as there is a `Molecule` class in interface.

qcfractal/storage_sockets/models.py
```diff
@@ -0,0 +1,259 @@
...
+    """
+        Hold the result of an atomic single calculation
+    """
+
+    # uniquely identifying a result
+    program = db.StringField(required=True)  # example "rdkit", is it the same as program in options?
+    driver = db.StringField(required=True)  # example "gradient"
+    method = db.StringField(required=True)  # example "uff"
+    basis = db.StringField()
+    molecule = db.ReferenceField(Molecule, required=True)   # or LazyReferenceField if only ID is needed?
```
dgasmith: Only ID is usually needed as we typically only pull data from `return_result` or `properties`.

qcfractal/tests/test_mongoengine_socket.py
```diff
@@ -0,0 +1,74 @@
...
+    """
+        Test the use of the ME class Molecule
+
+        Note:
+            creation of a Molecule using ME is not implemented yet
+            Should create a Molecule using: mongoengine_socket.add_molecules
+    """
+
+    # don't use len(Molecule.objects), slow
+    num_mol_in_db = Molecule.objects().count()
```
dgasmith: It appears that MongoEngine using a global Mongo client here?

qcfractal/tests/test_mongoengine_socket.py
```diff
@@ -0,0 +1,74 @@
...
+    result_list = Molecule.objects(molecular_formula="H4O2")
+    assert len(result_list) == 2
+    assert result_list[0].multiplicity == 1
+
+    # Query with fields NOT in the model. works too!
+    result_list = Molecule.objects(charge=0)
+    assert len(result_list) == 2
+
+    # get unique by hash and formula
+    one_mol = Molecule.objects(molecule_hash=water_mol.molecule_hash,
```
dgasmith: I guess we can do the following in some cases as well:
```python
query = {"molecule_hash": water_mol.molecule_hash, "molecular_formula":water_mol.molecular_formula}
.query(**query).only(molecule_hash=True).limit(5)
```

The attributed chaining feels awkward at the moment compared to PyMongo-- is this something that you become more accustomed to? Is there a way to do specify this in one instead of chaining?

qcfractal/storage_sockets/models.py
doaa-altarawy: Yes, it's a to_json() method. We always want to work with the class itself unless we will send it over the REST API.

For the molecule naming, I think there should be no issue in the future. Portal won't shouldn't import or use the storage package at all. Levi's classes should be in another package used by Portal and not by Fractal.

qcfractal/storage_sockets/models.py
doaa-altarawy: Which field are you referring to?

By referencing the class we mean it's going to be pulled and used. If we use class_id only, then it means we will rarely use it or pull it. 
So

qcfractal/tests/test_mongoengine_socket.py
doaa-altarawy: Yes.

qcfractal/tests/test_mongoengine_socket.py
doaa-altarawy: chaining is the typical way, similar to Pandas (and Dash as you have mentioned).

I'll post example of how to do the query. The key here is to stop using the JSON style, lots of quotes and nested brackets :)

dgasmith:
Certainly more work to do here, but I think everything looks good! @doaa-altarawy ready to merge?

Lnaden:
Woo!

PR #83, opened 2018-10-26
Task/Service Error queries

dgasmith:
## Description
The client is now able to provide the status of any job along with possible error messages. This interface will evolve significantly over time, but presents a working example. Items added:
 - `client.check_tasks` to find status of tasks
 - `client.check_serivces` to find the status of services
 - Error message field renamed from "error" to "error_message" to be better reflect the intent.

## Status
- [x] Ready to go

### DIFF
--- a/qcfractal/interface/client.py
+++ b/qcfractal/interface/client.py
@@ -444,4 +444,18 @@ class FractalClient(object):
         else:
             return r.json()["data"]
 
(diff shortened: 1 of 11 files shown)


dgasmith:
Right now `check_tasks({})` will return the status of _every_ task in the queue. Currently, this could lead to pulling too much data.

PR #84, opened 2018-10-26
Pins QCEngine version

dgasmith:
Pins version to v0.3.1 so that we do not simply pull from the head and can increment as necessary.

## Status
- [x] Ready to go


dgasmith:
I need to get this so I can edit QCEngine.

PR #85, opened 2018-10-29
CLI Cleanup and ping variable

dgasmith:
## Description
Slightly cleans up the CLI by correctly functioning and name spacing the code. Also allows users to tune how often to ping the sever. 2 seconds is good for testing, but production is too much so moved to 15 seconds by default.

## Status
- [x] Ready to go

### DIFF
--- a/qcfractal/cli/qcfractal_manager.py
+++ b/qcfractal/cli/qcfractal_manager.py
@@ -9,46 +9,58 @@ import tornado.log
 import qcfractal
 from . import cli_utils
 
-parser = argparse.ArgumentParser(description='A CLI for the QCFractal QueueManager.')
-subparsers = parser.add_subparsers(help='QueueManager Backend Type', dest='adapter_type')
-dask_parser = subparsers.add_parser('dask', help='Dask QueueManager')
-fw_parser = subparsers.add_parser('fireworks', help='Fireworks QueueManager')
-
-# Options for Dask
-dask_parser.add_argument("--dask-uri", type=str, help="URI of the dask-server")
-dask_parser.add_argument(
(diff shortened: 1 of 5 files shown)


dgasmith:
LGTM really does not like brackets in `format` strings for any purpose other than the format itself. Fair enough.

Lnaden:
The `\{` is not an escape for the curly braces in Python, that is a legitimate error, throws this error:

```
KeyError: "'fireworks', 'dask'\\
```

Replace the escaped curly braces with a double curly brace: i.e. `\{ \}` -> `{{ }}`

dgasmith:
Ah good to know, this is the first project that I have used `format` in.

`%` has worked for the past 40 years in Python/C/C++! Get off my lawn.

Lnaden:
I prefer `format` myself because its much more clear and more powerful than just `%`, but it does require learning a mini-language and has a couple oddities like this. If it gets too confusing, (ala `{{{{x}}}}` = `{x}`, you can also just do string concatenation through `+` or Python new line string concatenation.

dgasmith:
Completely agree that `format` is the way to go in the future, why the whole project is that way. If possible I would prefer to stick to just `format` for consistency throughout the QCA software stack. We can of course bend that, but be good be consistent if possible.

Lnaden:
I have yet to see a use case where `%` is the superior option to `format`, so I think enforcing `format` throughout QCA, except where external libs are used (e.g. versioneer).

dgasmith:
Can we spin non-URI support off as an Issue to tackle later? Changing away from URI's at the moment will conflate with the mongoengine PR.

doaa-altarawy:
Ok, sounds good.

Issue #91, opened 2018-10-31
from_server does not find an existing OpenFFWorkflow


ChayaSt:
**Describe the bug**
When I try to retrieve an existing workflow by calling `collections.OpenFFWorkflow.from_server(name='chemper2_rdkit', client=client)` of a collection that I had already registered, I get the following error:
```
KeyError                                  Traceback (most recent call last)
<ipython-input-3-b4da8d7fca6e> in <module>
----> 1 wf_2 = portal.collections.OpenFFWorkflow.from_server(name='chemper2_rdkit', client=client)

~/src/molssi/QCFractal/qcfractal/interface/collections/collection.py in from_server(cls, client, name)
    121             raise KeyError("Warning! `{}: {}` not found.".format(class_name, name))
    122 
--> 123         return cls.from_json(tmp_data["data"][0], client=client)
    124 
    125     @classmethod

~/src/molssi/QCFractal/qcfractal/interface/collections/collection.py in from_json(cls, data, client)
    168         print('data: {}'.format(data))
    169         # Allow PyDantic to handle type validation
--> 170         return cls(name, client=client, **data)
    171 
    172     def to_json(self, filename=None):

~/src/molssi/QCFractal/qcfractal/interface/collections/openffworkflow.py in __init__(self, name, options, client, **kwargs)
     39         # Expand options
     40         if options is None:
---> 41             raise KeyError("No record of OpenFFWorkflow {} found and no initial options passed in.".format(name))
     42         super().__init__(name, client=client, **options, **kwargs)
     43 

KeyError: 'No record of OpenFFWorkflow chemper2_rdkit found and no initial options passed in.'
```
But when I look at `client.get_collection("OpenFFWorkflow", 'chemper2_rdkit', full_return=True)`
I see that the workflow does exist:
```
{'meta': {'errors': [],
  'n_found': 1,
  'success': True,
  'error_description': False,
  'missing': []},
 'data': [{'name': 'chemper2_rdkit',
   'collection': 'openffworkflow',
   ...
```
But since `options` is `None` [here](https://github.com/MolSSI/QCFractal/blob/master/qcfractal/interface/collections/openffworkflow.py#L22-L41) It raises a `KeyError`. 
**To Reproduce**
```
# set up a client
client = portal.FractalClient("https://localhost:7777/", verify=False)
# register workflow
with open('example_workflow.json') as f:
    wf_json = json.load(f)['example']['fragmenter']
[...]

PR #82, opened 2018-10-26
[WIP] add optimization service to openffworkflow 

ChayaSt:
## Description
This PR adds an splits up `add_fragment()` to `add_torsiondrive()` and `add_optimize()`. This will allow us to submit general and constrained geometry optimization jobs. This is needed for running limited torsion scans for rings and  double bonds.  

## Todos
Notable points that this PR has either accomplished or will accomplish.
  - [x] Changed `_required_fields` `torsiondrive_meta` and `optimization_meta` to `torsiondrive_static_options` and `optimization_static_options`. An example corresponding `workflow.json` lives [here](https://github.com/openforcefield/fragmenter/blob/workflow/examples/example_workflow.json)
  - [x] replaces `add_fragment()` with `add_torsiondrive()`
  - [x] adds `add_optimize()` to submit geometric jobs directly without `torsiondrive`
  - [ ] build out an optimization service

## Questions
- [ ] @dgasmith,  I still need to add an `optimize` service. Should I model it on [`torsiondrive_service`](https://github.com/MolSSI/QCFractal/blob/master/qcfractal/services/torsiondrive_service.py)?

## Status
- [ ] Ready to go

PR #92, opened 2018-10-31
OpenFF: Fix OpenFFWorkflow class signature

ChayaSt:
## Description
This PR addresses https://github.com/MolSSI/QCFractal/issues/91

## Todos
  - [x] Changed OpenFFWorkflow class signature to:
```
def __init__(self, name, client=None, **kwargs):
    super().__init__(name, client=client, **kwargs)
```
  - [x] Added `torsiondrive_input` to DataModel which was inadvertently removed with https://github.com/MolSSI/QCFractal/pull/82.

## Status
- [ ] Ready to go

### DIFF
--- a/devtools/conda-envs/dask.yaml
+++ b/devtools/conda-envs/dask.yaml
@@ -28,5 +28,5 @@ dependencies:
     # Pip depends
   - pip:
     - pydantic
-    - git+git://github.com/MolSSI/QCEngine@v0.3.1#egg=qcengine
+    - git+git://github.com/MolSSI/QCEngine@v0.4.0#egg=qcengine
     - codecov
--- a/devtools/conda-envs/fireworks.yaml
+++ b/devtools/conda-envs/fireworks.yaml
@@ -24,7 +24,7 @@ dependencies:
     # Pip depends
   - pip:
     - pydantic
-    - git+git://github.com/MolSSI/QCEngine@v0.3.1#egg=qcengine
+    - git+git://github.com/MolSSI/QCEngine@v0.4.0#egg=qcengine
 
     # Fireworks
     - fireworks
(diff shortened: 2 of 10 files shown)


ChayaSt:
Yes, I'll add the defaults to this PR. For reference, an example of the structure and defaults is [here](https://github.com/openforcefield/fragmenter/blob/workflow/examples/example_workflow.json)

ChayaSt:
I also see there's a KeyError 
https://travis-ci.org/MolSSI/QCFractal/jobs/449103349#L1168-L1170

There seems to be something weird going on because when I print out what `optimization_meta` [here](https://github.com/MolSSI/QCFractal/blob/master/qcfractal/interface/collections/openffworkflow.py#L198-L200) it's just 2 empty dictionaries:
```
{'optimization_meta': {}, 'qc_meta': {}}
```
but `torsion_meta` does keep the options specified in the workflow JSON and when I print it it has them:
```
{'torsiondrive_meta': {}, 'optimization_meta': {'program': 'geometric', 'coordsys': 'tric'}, 'qc_meta': {'driver': 'gradient', 'method': 'UFF', 'basis': '', 'options': 'none', 'program': 'rdkit'}}
```

dgasmith:
I believe the signature in the tests will need to be changed to:
```
wf = portal.collections.OpenFFWorkflow("Workflow1", client=client, **openff_workflow_options)
```

ChayaSt:
The tests should pass now. 
@Lnaden, I added the default `kwargs`.

dgasmith:
I think this is ready to go.

Lnaden:
I think this is good. Thanks for adding the defaults and better typing to the OpenFFWorkflow, @ChayaSt !

PR #94, opened 2018-11-02
OpenFF: workaraound to get optimized molecules

ChayaSt:
## Description
Small fix that adds final molecules from optimization jobs to `list_final_molecules`. 

## Status
- [ ] Ready to go

### DIFF
--- a/qcfractal/interface/collections/openffworkflow.py
+++ b/qcfractal/interface/collections/openffworkflow.py
@@ -356,7 +356,13 @@ class OpenFFWorkflow(Collection):
             tmp = {}
             for k, v in self.data.fragments[frag].items():
                 if v["hash_index"] in self._torsiondrive_cache:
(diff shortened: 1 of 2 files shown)


qcfractal/interface/collections/openffworkflow.py
```diff
@@ -356,7 +356,13 @@ def list_final_molecules(self, fragments=None, refresh_cache=False):
             tmp = {}
             for k, v in self.data.fragments[frag].items():
                 if v["hash_index"] in self._torsiondrive_cache:
-                    tmp[k] = self._torsiondrive_cache[v["hash_index"]].final_molecules()
+                    obj = self._torsiondrive_cache[v["hash_index"]]
```
dgasmith: Looks good, can you add a test that pulls molecules after an optimization?

PR #96, opened 2018-11-05
Collections: Fixes JSON serialization

dgasmith:
Tiny fix in the JSON serialization and dumping.

## Status
- [x] Ready to go

### DIFF
--- a/qcfractal/interface/collections/collection.py
+++ b/qcfractal/interface/collections/collection.py
@@ -177,11 +177,12 @@ class Collection(abc.ABC):
         ret : dict
             A JSON representation of the Collection
         """
+        data = self.data.dict()
         if filename is not None:
             with open(filename, 'w') as open_file:
-                json.dump(self.data, open_file)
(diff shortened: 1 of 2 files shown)


Lnaden:
Do we want to dump the dict of the data, or allow the natural `self.data.to_json()` function to handle it (would cast it as PyDantic Json, but should be pretty close to natural `json` since its not a schema output, but would match the PyDantic Schema)

dgasmith:
I think once we get a handle on the MongoEngine `to_json/to_mongo/to_python` and the Pydantic `dict/json` functions we will have a much better idea of how we want to handle the serializers. This is here to fix an immediate bug when dealing with dumping these files to disk.

Lnaden:
> This is here to fix an immediate bug when dealing with dumping these files to disk.

Fair enough, there might be a new rare bug introduced in going from this output to python object, but we can fix that on a second pass if it becomes an issue

Issue #99, opened 2018-12-07
Optimize DB indexes


doaa-altarawy:
After the new DB design, create appropriate DB indexes to support and speed up current queries. 

**Note**

Keep an index for hash_inx field in Procedure table.

dgasmith:
The intent of the PR has been accomplished. We will re-review after the SQL changes.

PR #100, opened 2018-12-07
Optimize DB indexes

doaa-altarawy:
Implementing issue: Optimize DB indexes #99

## Status
- [x] Ready to go

### DIFF
--- a/qcfractal/storage_sockets/models.py
+++ b/qcfractal/storage_sockets/models.py
@@ -264,7 +264,7 @@ class TaskQueue(db.DynamicDocument):
 
     meta = {
         'indexes': [
-            '-created_on',
+            'created_on',
             'status',
             'manager',
             # {'fields': ("status", "tag", "hash_index"), 'unique': False}
--- a/qcfractal/storage_sockets/models.py
+++ b/qcfractal/storage_sockets/models.py
@@ -132,12 +132,8 @@ class Result(BaseResult):
     properties = db.DynamicField()  # accept any, no validation
     return_result = db.DynamicField()  # better performance than db.ListField(db.FloatField())
     provenance = db.DynamicField()  # or an Embedded Documents with a structure?
-        #  {"creator": "rdkit", "version": "2018.03.4",
-        # "routine": "rdkit.Chem.AllChem.UFFGetMoleculeForceField",
-        # "cpu": "Intel(R) Core(TM) i7-8650U CPU @ 1.90GHz", "hostname": "x1-carbon6", "username": "doaa",
-        # "wall_time": 0.14191770553588867},
 
-    schema_name = db.StringField() #default="qc_ret_data_output")
+    schema_name = db.StringField()  # default="qc_ret_data_output"??
     schema_version = db.IntField()  # or String?
 
     meta = {
@@ -195,7 +191,7 @@ class Procedure(BaseResult):
 
 class OptimizationProcedure(Procedure):
     """
-        An Optimization  procedure
+        An Optimization  procedure (not used so far)
     """
 
     procedure = db.StringField(default='optimization', required=True)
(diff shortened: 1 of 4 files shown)


qcfractal/storage_sockets/mongoengine_socket.py
```diff
@@ -457,14 +457,15 @@ def get_molecules(self, molecule_ids, index="id"):
...
+        if index == "id":
             molecule_ids, bad_ids = _str_to_indices_with_errors(molecule_ids)
 
         # Project out the duplicates we use for top level keys
         proj = {"molecule_hash": False, "molecular_formula": False}
 
         # Make the query
-        data = self._tables["molecules"].find({index: {"$in": molecule_ids}}, projection=proj)
+        query = {index+'__in': molecule_ids}
+        data = Molecule.objects(**query).exclude(*proj).as_pymongo()
```
dgasmith: Shouldn't this be `only` instead of `exclude`?

qcfractal/storage_sockets/models.py
```diff
@@ -53,8 +53,10 @@ def __str__(self):
     meta = {
         'collection': 'molecules',
         'indexes': [
-            {'fields': ('molecule_hash', 'molecular_formula'),
-             'unique': False
+            {'fields': ('molecule_hash',), 'unique': True},
```
dgasmith: This should *likely* be unique, but is not guaranteed. We use it effectively as a very fast lookup for molecules.

qcfractal/storage_sockets/mongoengine_socket.py
doaa-altarawy: Yes, thanks.

qcfractal/storage_sockets/models.py
doaa-altarawy: In normal cases, can we assume that molecule_formula is almost unique like molecule_hash?
If not, by which factor?

dgasmith: `molecular_formula` might relate to 1-1e6 molecules, average maybe 1e2-3.

doaa-altarawy: If we will have lots of duplicates in molecule_formula then we will need another index just for molecule_formula.
Another question, I don't see that we do a combined query on ('molecule_hash', 'molecular_formula'), we always do it on either one. If this is the case, then we should have single index of each field not a combined one.

dgasmith: Right, that one is pretty much an either or combination.

qcfractal/storage_sockets/mongoengine_socket.py
doaa-altarawy: I checked back and the original method was removing the projection, so it's `exclude` not `only`.

qcfractal/storage_sockets/models.py
```diff
@@ -53,8 +53,10 @@ def __str__(self):
...
+            {
+                'fields': ('molecule_hash', 'molecular_formula'),
```
dgasmith: This should just be `molecular_formula` here.

qcfractal/storage_sockets/mongoengine_socket.py
dgasmith: Ah, I see. Can we change `proj` to a list then so we are not unpacking keys in a dict?

doaa-altarawy:
This is ready to go, please merge.

qcfractal/storage_sockets/models.py
```diff
@@ -257,19 +256,20 @@ class TaskQueue(db.DynamicDocument):
...
 
     meta = {
         'indexes': [
-            '-created_on',
+            'created_on',
             'status',
             'manager',
             # {'fields': ("status", "tag", "hash_index"), 'unique': False}
-            {'fields': ("base_result",), 'unique': True}  # new
+            {'fields': ('base_result',), 'unique': True},  # new
```
dgasmith: Is this strictly true as this might reference several tables?

doaa-altarawy: Yes, because it's a DBref, i.e., both the table name and the ID.

PR #106, opened 2019-01-06
Heartbeat Manager

### LINKED ISSUE #90: Safe `kill` when run as background thread
Currently a `kill pid` when run as background thread will cause a hard stop as KeyBoardInterupt only catches SIGINT and kill causes throws a SIG. Between the Python `signal` modules and the [signal_callback](https://www.tornadoweb.org/en/stable/ioloop.html#tornado.ioloop.IOLoop.add_callback_from_signal) function this can be handled correctly.

### LINKED ISSUE #95: Manager Heartbeats
When managers shut down they tell their QCFractal instance to revert owned RUNNING jobs to WAITING so that they can be returned to the queue. However managers that have a hard shutdown (SIGINT/SIGKILL) cannot call back to the sever to free jobs. To prevent this, heartbeats should be implemented so that jobs related to a manager are freed if they are not heard from in a certain amount of time. Related to #90.

dgasmith:
## Description
This PR adds manager heartbeats to the project. This allows jobs that have been allocated to managers that do not shut down properly to be recycled and given to other managers.

The heartbeat interval is set by the FractalServer instance and passed to the managers to ensure consistency.

## Todos
 - [x] Adds heartbeats, closes #95.
 - [x] Allows shutdown of CLI invocations via SIGINT and SIGTERM to have a better chance of a clean shutdown, closes #90.
 - [x] A new FractalClient requests some data from the server to ensure the connection is live. Error messages should be slightly more explicit now.
 - [x] When a FractalServer owns a Manager, all manager REST queries are now run in a background thread so that communication is not blocked. This is not exactly an ideal solution as we need to continuously build threads. There are plenty of ways to improve in the future, but the current version works and is relatively robust.

## Status
- [x] Ready to go

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -84,7 +84,7 @@ before_script:
   - conda list
 
 script:
-  - py.test -v --runslow --cov=qcfractal qcfractal/
+  - py.test -v -rsx --runslow --cov=qcfractal qcfractal/
   - py.test -v --runexamples examples/
 
 notifications:
--- a/qcfractal/cli/cli_utils.py
+++ b/qcfractal/cli/cli_utils.py
@@ -5,6 +5,7 @@ Utilities for CLI programs
 import copy
 import importlib
 import json
+import signal
 
 import yaml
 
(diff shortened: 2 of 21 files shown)


dgasmith:
PyTest and cov will fail due to:
https://github.com/pytest-dev/pytest-cov/pull/253
https://github.com/conda-forge/pytest-cov-feedstock/pull/14

Waiting another few hours should see it resolved.

dgasmith:
If anyone has a moment to review this let me know, otherwise merging sometime tonight.

PR #110, opened 2019-01-15
CI cleanup and QCElemental addition

dgasmith:
## Description
This wound up being a bit winding of a PR due to due to issues with local ARC systems. 

## Todos
 - [x] Adds QCElemental dependency.
 - [ ] Replaces duplicate QCElemental code.
 - [x] Generates environments automatically to handle core dependency changes (overkill, but works well).
 - [x] Switch to pulling from conda-forge where we can obtain many additional depends.
 - [x] Allows a development env CI to test against several different upstreams.
 - [x] Adds python typing to additional files.

## Status
- [ ] Ready to go


devtools/conda-envs/dev_head.yaml
```diff
@@ -0,0 +1,34 @@
...
+# Environment specific includes
+  - rdkit
+
+# QCArchive includes
+
+# Pip includes
+  - pip:
+    - pydantic
+    - git+git://github.com/MolSSI/QCEngine#egg=qcengine
+    - git+git://github.com/MolSSI/QCElemental#egg=qcengine
```
loriab: I don't know this syntax, but should it be `egg=qcelemental`?

dgasmith: That could help yes :)

dgasmith:
Best to swap out QCElemental in the next PR rather than overload this one too much.

PR #116, opened 2019-01-20
Manager cleanup

dgasmith:
## Description
Refactoring in preparation for a Parsl Manager script and removes duplication in the CLI's.
- [x] Cascade repr from managers to adapters.
- [x] Cascade shutdown from managers to adapters.
- [x] Simplifies server and manager CLI.
- [x] Allows Parsl to be created via Config rather than Dataflow
- [x] Parsl manager script example

## Status
- [x] Ready to go


dgasmith:
Going to merge this unless anyone has comments.

PR #120, opened 2019-01-22
Refactors Adapter Testing and Server + Manager arrangement 

### LINKED ISSUE #79: QueueAdapter Base Class
A base class for the QueueAdapter tech should be created to lower the current duplication and ensure a number of functions are correctly implemented.

dgasmith:
## Description
This PR accomplishes the following:
 - [x] Refactors adapters with a base class, closes #79.
 - [x] Adds a general Python Executor adapter. Should work with any `concurrent.futures.Executor` like object.
 - [x] Refactors tests to almost always use Executors for Server + Compute
 - [x] Adds a set of tests that specifically tests the Adapter polymorphism.
 - [x] Refactors CLI to reflect the above. 

## Status
- [x] Ready to go


dgasmith:
Ok, this is ready to go I think.

Issue #108, opened 2019-01-14
Add QCElemental


dgasmith:
QCElemental is now on conda-forge. First pass integration includes:
 - Molecule parsing
 - Physical constants

Future passes will include:
 - Molecule/Results etc object base models

dgasmith:
Closed by #125.

PR #125, opened 2019-01-29
Start QCEngine 0.5 conversion in qcfractal [WIP]

Lnaden:
## Description
This PR will upgrade QCFractal to use QCEngine 0.5+ and start adding in the models from QCElemental 0.2+. This is a very much in progress PR, so expect lots of changes.

cc @dgasmith, you should be good to push to this branch.

## Status
- [ ] Ready to go


qcfractal/interface/collections/openffworkflow.py
```diff
@@ -202,7 +202,7 @@ def add_fragment(self, fragment_id, data, provenance=None):
 
             # add back to fragment data
             packet["hash_index"] = ret
-            packet["provenance"] = provenance
+            # packet["provenance"] = provenance
```
Lnaden: Will we want to add this back in when we have the `Provenance` model?

qcfractal/testing.py
```diff
@@ -380,7 +380,8 @@ def build_managed_compute_server(mtype):
         manager.close_adapter()
 
 
-@pytest.fixture(scope="module", params=["pool", "dask", "fireworks", "parsl"])
+# @pytest.fixture(scope="module", params=["pool", "dask", "fireworks", "parsl"])
+@pytest.fixture(scope="module", params=["pool"])
```
Lnaden: Note to revert this line

qcfractal/interface/collections/openffworkflow.py
dgasmith: No, thinking on it I don't think so here. The provenance information will be overridden by TorsionDrive itself here.

dgasmith: We will need to reorganize somethings here so that this is `fragment` provenance data not TD provenance data. I will spin out an issue.

Lnaden: Thats fine, do we want to fix it as part of this PR or a future one?

dgasmith: Future, writing an issue now.

dgasmith:
Note that this will close #108.

dgasmith:
So conda-forge QCEngine 0.5.1 had an issue with versioneer where reported version was `0+unknown`. I updated conda-forge to have a `py_1` QCEngine 0.5.1 with a fixed versioneer. However Travis keeps pulling `py_0`. Go figure.

dgasmith:
Whoo!

PR #130, opened 2019-01-30
Pydantic Body/Requests

dgasmith:
## Description
This PR adds REST models to the web web handler for non-queue interfaces. Provides both client and server side level of validation. Most of it is bare bones right now, but can be improved upon later.

## Status
- [x] Ready to go

edited by lnaden

### DIFF
--- a/qcfractal/interface/models/rest_models.py
+++ b/qcfractal/interface/models/rest_models.py
@@ -1,0 +1,40 @@
+from pydantic import BaseModel
+from typing import Dict, List, Tuple
+from enum import Enum
+from qcelemental.models import Molecule
+from qcelemental.models.common_models import ndarray_encoder
+
+__all__ = ["MoleculeGETBody", "MoleculeGETResponse"]
+
+
+class ResponseMeta(BaseModel):
+    errors: List[Tuple[str, str]]
+    n_found: int
+    success: bool
+    error_description: str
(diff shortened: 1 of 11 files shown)


Lnaden:
Sorry about the sloppy git history there, i rebased from master to get the current master's changes in to get tests to trigger, we can clean up again after the fact

Lnaden:
This is ready for formal review now, I think all the changes are in

qcfractal/interface/models/rest_models.py
```diff
@@ -0,0 +1,221 @@
...
+    class Meta(BaseModel):
+        projection: Dict[str, Any] = None
+
+    meta: Meta = Meta()
+    data: Dict[str, Any]
+
+    @validator("data", whole=True)
+    def only_data_keys(cls, v):
+        keys = ["program", "molecule", "driver", "method", "basis", "options", "hash_index", "task_id", "id", "status"]
+        final_data = {key: v[key] for key in keys if key in v}
```
dgasmith: Can we do something like:

```
valid_keys = {"program", ...}
final_data = {key: v[key] for key in (v.keys() & valid_keys)}
```

I feel like we miss out on set manipulation too much!

dgasmith:
I will merge this, and handle the one issue later.

PR #142, opened 2019-02-04
Add REST PyDantic Objects for Task- and ServiceQueue Objects

Lnaden:
## Description
Similar to #130, this PR adds the same GET and POST REST API to the TaskQueue and ServiceQueue.

There are a few oddities here, namely with a few of the methods accepting a variety of inputs, but otherwise this adds client and server side type checking to objects.

The only concern I have is on the `Molecule.dict()` and `.json()` calls in that we cannot reliably recover the `id` field without supplying all other fields, or doing some post-dict addition of the `id` field back in for the `TorsionDrive` Service. I have a quick 2-line workaround for now, so its fine, just something to note (i've added a line-comment below)

I did not touch the QueueManagerHandler object since that is all server side, and may not need validation (or may be more complex) than we need right now. 

## Status
- [x] Ready to go


qcfractal/queue/handlers.py
```diff
@@ -70,12 +93,19 @@ def post(self):
...
+        mol_response = MultiMoleculeGETResponse(**mol_query)
 
         # Build out services
         submitted_services = []
-        for idx, mol in mol_query["data"].items():
-            tmp = services.initializer(self.json["meta"]["service"], storage, self.json["meta"], mol)
+        for idx, mol in mol_response.data.items():
+            mol_dict = mol.json(as_dict=True)
+            # Ensure ID is present
+            if hasattr(mol, 'id') and mol.id is not None:
```
Lnaden: Workaround the `id` field being dropped on dict conversion

qcfractal/interface/client.py
```diff
@@ -384,13 +388,13 @@ def add_compute(self,
                     method: str,
                     basis: str,
                     driver: str,
-                    options: str,
-                    molecule_id: str,
+                    options: Union[str, None],
+                    molecule_id: Union[str, Dict[str, Any], List[Union[str, Dict[str, Any]]]],
```
dgasmith: We need to sort this out, the `Dict[str, Any]` should be a Molecule I think?

dgasmith:
Looks like there are failures in the examples CLI runs.

Lnaden:
I think i fixed the examples now (made the output more helpful)

I cleaned up some of the `import` statements involving `Molecule`, which was imported in 3 places before now

I'm not quite happy with that coverage decrease, I'll look into it

Lnaden:
No idea what's going on with CodeCov there, I think it'll be fine though

dgasmith:
CodeCov is going down due to incomplete testing I believe. It should patch itself it up if all tests pass.

Lnaden:
> CodeCov is going down due to incomplete testing I believe. It should patch itself it up if all tests pass.

So it did. +0.28 overall

Issue #16, opened 2018-08-17
REST and Schema documentation style


dgasmith:
Current schema documentation style should be improved as the [current version](https://molssi-qc-schema.readthedocs.io/en/latest/auto_topology.html) is perhaps not the most optimal. Several options could be:
 - [Foldable options](http://getyank.org/latest/yamlpages/options.html)
 - [Bootprint](https://github.com/bootprint/bootprint-json-schema)
 - [Swagger.io](https://petstore.swagger.io)
 - [Stripe](https://stripe.com/docs/api?utm_source=zapier.com&utm_medium=referral&utm_campaign=zapier)

Most of these options are build for REST endpoints (which we have). However, we also have pure "schema" such as the QCSchema that would be nice to present in a precise syntax.

Anyone have an opinion or additional options to examine?

Lnaden:
I would avoid using the foldable/collapsable styles for doing auto-gen things. What you linked from the YANK project are all the YAML options and documentation which I had write and link manually. There is an auto-gen API docs we use for the API, but I have never found the Sphinx autodoc modules to be that good for complex projects.

dgasmith:
I have been toying around with the idea of having pydantic classes as REST requests/responses templates. The benefit here is several fold:
- We can automatically validate requests 
- We can return consistent validation errors across all models
- We can automatically export pydantic to jsonschema for Swagger.io-style docs (which I think we should lean on)
- We keep a consistent object model across the project.

Thoughts?

loriab:
I've had minimal experience with pydantic, but what I've had I've liked. Prevents me from adding undeclared things to `self`. Prevents me from appending to wrong level of dict. If pydantic lets me write less json schema docs directly, and nicely extends to generating good stuff I'm even less acquainted with (e.g., REST), I'm for it.

dgasmith:
Related to #330.

dgasmith:
Closed in https://github.com/MolSSI/QCFractal/pull/349.

PR #145, opened 2019-02-05
Add final REST interface for QueueManager

Lnaden:
## Description
Similar to #130 and #142, this PR adds the final REST interfaces for the API

I believe this is the last REST validation we need to add for now to have them all in, even if rigorous validation is not fully preset beyond things like `Dict[str, Any]`

Advances #16 

## Status
- [x] Ready to go

### DIFF
--- a/qcfractal/interface/models/rest_models.py
+++ b/qcfractal/interface/models/rest_models.py
@@ -16,6 +16,8 @@ __all__ = [
     "ProcedureGETBody", "ProcedureGETReponse",
     "TaskQueueGETBody", "TaskQueueGETResponse", "TaskQueuePOSTBody", "TaskQueuePOSTResponse",
     "ServiceQueueGETBody", "ServiceQueueGETResponse", "ServiceQueuePOSTBody", "ServiceQueuePOSTResponse",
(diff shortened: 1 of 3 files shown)


Lnaden:
I'm adding some comments here for an issue which came up in case we run into it again and forget why.

tl;dr: Always use `Dict[Any, Any]` at minimum instead of `Dict` when typing with Pydantic

If you make a Pydantic object with the `typing.Dict` type and don't include the `[..., ...]` args, it will throw the following pair of errors:

```
TypeError: issubclass() arg 1 must be a class

The above exception was the direct cause of the following exception:
...
RuntimeError: error checking inheritance of ~KT (type: KT)
```

This is NOT the Boltzmann Temperature (kT), but instead the `typing` module's generic KeyType and ValueType objects which it uses as placeholders. Pydantic has no concept of these as doing `Dict` without `Dict[]` only throws this error in Python 3.7. (as of Pydantic 0.18)

For example, assuming Pydantic 0.18, the below will work in its entirety Python 3.6, but in Python 3.7, class `X` and class `Y` will.

```python
from pydantic import BaseModel
from typing import Dict, Any

class X(BaseModel):
    x: Dict

class Y(BaseModel):
    y: Dict[str, Any]
```

Lnaden:
The py37 tests do pass now? Unless you meant "must pass?"

PR #152, opened 2019-02-10
Gridopt: Relative starting grid

dgasmith:
## Description
Allow a GridOptimization to start from a relative position of the input molecule.

## Status
- [x] Ready to go

### DIFF
--- a/qcfractal/interface/models/gridoptimization.py
+++ b/qcfractal/interface/models/gridoptimization.py
@@ -39,8 +39,8 @@ class ScanDimension(BaseModel):
 
     @validator('steps', whole=True)
     def check_steps_monotonic(cls, v):
-        if not all(x < y for x, y in zip(v, v[1:])):
-            raise ValueError("Steps are not monotonically increasing.")
+        if not (all(x < y for x, y in zip(v, v[1:])) or all(x > y for x, y in zip(v, v[1:]))):
+            raise ValueError("Steps are not monotonically increasing or decreasing.")
 
         return v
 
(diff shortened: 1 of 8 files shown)


qcfractal/interface/models/gridoptimization.py
```diff
@@ -39,8 +39,8 @@ def check_indices(cls, v, values, **kwargs):
 
     @validator('steps', whole=True)
     def check_steps_monotonic(cls, v):
-        if not all(x < y for x, y in zip(v, v[1:])):
-            raise ValueError("Steps are not monotonically increasing.")
+        if not (all(x < y for x, y in zip(v, v[1:])) or all(x > y for x, y in zip(v, v[1:]))):
```
Lnaden: Does this check need the steps to be *strictly* in-/decreasing, i.e. can `v[i] == v[i+1]`? If not, then I would change the error message to say "Steps are not strictly monotonically...".

Do you expect this to be a large list? If so, we could reduce the iteration calls with a `np.diff` and do an `all` on the result of that. (really small optimization that is likely not needed)

dgasmith: Number of elements larger than 20 is not expected and NumPy could potentially make this more difficult using `np.diff`. Since we form a cartesian grid `v[i] == v[i+1]` doesn't make much sense to allow.

I think we can elide the minute details of order theory for strict monotonicity here.

dgasmith:
Thanks for the review, merged!

Issue #81, opened 2018-10-26
Client add_compute should throw if molecule not found


Lnaden:
**Describe the bug**
The current `add_compute` as part of the client will work if you give it a molecule ID already in the server or a fully defined molecule which will be inserted into the server.

The corner case is when you give it a molecule ID which is NOT in the server already, the current function will not error and will simply return an empty return object. We should change this to instead throw an error indicating which IDs were missing.

dgasmith:
Not quite as simple as just throwing an error. Imagine if a user submits `client.add_compute(... moluleces=["known_id", "unknown_molecule!"])`. We can:
 - Not accept any new compute and return an error
 - Only accept the valid compute and return a part success, part failure.
   - Currently we return a `{"submitted": , "completed"}` dictionary of task id's for future use. We can add a "error" field here if we with the problem molecules? This would require users to actively check the returned dictionary to see if anything went wrong.
   - This also applied to things like if the "options_name" field is not found. Everything will need to return with an error.

@doaa-altarawy do you have thoughts on how we should handle this as well?

PR #155, opened 2019-02-11
Molecule/Result add order

dgasmith:
## Description
I started working through the molecule/result add as it can fail if bad ID's are sent in the submission. We previous stated that order in = order out for all add/POST operations so this was the start of the fix.

Continuing this PR, the following changes are made:
 - `add_results/add_procedures` now have order in == order out.
 - `add_results/add_procedures` now returns (ids, submitted, existing). Where `ids` is the result or procedure id in the order provided and submitted/existing indicate whether the computation is new or not.
 - Options have been renamed to keywords to correctly reflect the schema.
 - The Options object has been renamed to KeywordSet to properly reflect its goal.
 - Fixes, #81.

## Status
- [ ] Ready to go


doaa-altarawy:
+1 for the rename to keywords.
Looks good to me. Is it ready to go? It's better for me to rebase now.

dgasmith:
I think its ready to go, if you approve we can get it in.

dgasmith:
Looks like the CI hook didn't go through. Restarted and stepping away for a bit. Feel free to merge once everything passes.

Issue #112, opened 2019-01-15
GridOptimization


dgasmith:
Currently we have the TorsionDrive service which causes geometry optimizations in a wavefront propagation, as this can spawn multiple geometry optimization without dependance this makes for a good QCFractal service. Currently a single optimization looks something like:

```python
{
 'keywords': # geometry optimizer keywords
 'input_specification' # A QCSchema input specification minus a molecule
 'initial_molecule' # The initial molecule

# Above is input, below is the output bit
 'trajectory' # A ordered list of single results
 'energies' # A ordered list of energies
 'final_molecule' # The final molecule specification (sometime not in the trajectory itself)
}
```

If we have a GridOptimization we will likely want to reuse the single optimization framework. Would we want something similar like:

```python
{
 'keywords': # geometry optimizer keywords
 'input_specification' # A QCSchema input specification minus a molecule
 'initial_molecule' # The initial molecule

 `optimizations` # A list (?) of optimizations performed
}
```

What other top level fields do we want?

@leeping @yudongqiu

leeping:
Here is what I'm imagining:
1) One of the optimizations on the grid will be run first, starting from the user-provided structure, and all the others would depend on a previous one.  There is a dependence graph structure to the calculations (see image).
2) The user should be able to provide an option called `absolute` or `relative`. In the case of `absolute`, the first optimization is constrained to the nearest grid point, and the constraint values are absolute. In the case of `relative`, the first optimization is fully relaxed, and the constraint values are relative to the first optimized structure.
3) The user should provide a list of lists of constraint values, starting with the leading dimension.

![wechatimg2](https://user-images.githubusercontent.com/1441560/51215035-7cf3da00-18d4-11e9-820b-972cb9f9151c.jpeg)

dgasmith:
It might be good to start with the 1D case and move to 2D.

In general I think this should likely support n-dimensions and arbitrary combinations of bond/angle/torsion scans?

leeping:
Sure, I agree starting from 1D is a good idea.  Jessica Maat in the Mobley group is already starting to generate and use 2D data so it would be optimal to support that as well.

- Lee-Ping

dgasmith:
Apologies, I meant 1D when describing the spec.

[...]

PR #161, opened 2019-02-14
GridOptimization: Pre-optimization and relative scans

dgasmith:
## Description
Adds a pre-optimization ability and relative scan values to grid optimization as indicated in #112.

Depends on https://github.com/MolSSI/QCElemental/pull/30 and https://github.com/leeping/geomeTRIC/pull/53.

## Status
- [x] Ready to go


dgasmith:
Both upstream changes have been merged and new versions deployed. This should be ready to go.

PR #162, opened 2019-02-14
Save and get services with ME for validation

doaa-altarawy:
Save and get services with mongoengine for validation

## Status
- [ x ] Ready to go

### DIFF
--- a/qcfractal/storage_sockets/mongoengine_socket.py
+++ b/qcfractal/storage_sockets/mongoengine_socket.py
@@ -29,7 +29,7 @@ from bson.objectid import ObjectId
 from mongoengine.connection import disconnect, get_db
 
 from qcfractal.storage_sockets.models import Keywords, Collection, Result, \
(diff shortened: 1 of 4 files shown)


dgasmith:
Codecov hit is due to the fact that `_get_generic` is now deprecated. This is a good thing and we can remove it in another PR.

qcfractal/storage_sockets/mongoengine_socket.py
```diff
@@ -1698,7 +1789,7 @@ def queue_mark_complete(self, task_ids: List[str]) -> int:
         #         # automatically calls ClientSession.commit_transaction().
         #         # If the block exits with an exception, the transaction
         #         # automatically calls ClientSession.abort_transaction().
-        #           found = TaskQueue._collection.update_many(
+        #           found = TaskQueue._get_collection().update_many(
```
dgasmith: What is the difference between `_collection` and `_get_collection()`?

PR #165, opened 2019-02-14
Dataset Rework

dgasmith:
## Description
Reworks datasets and begins to spill them out. It became apparent that our main dataset types:
 - Dataset - A single molecule-compute DataFrame
 - ReactionDataset - A reaction-compute DataFrame
 - OptimizationDataset - A single-molecule optimization DataFrame

could not quite live in the same object, but could share several functions. This PR is meant to separate out the initial two datasets so that this behavior is more expected as we add new dataset types. This also frees us up to do something interesting things in the different collections. Datasets still have some work to be done, but the general idea here is to set us on the correct structure for the future.

## Status
- [ ] Ready to go

### DIFF
--- a/qcfractal/interface/collections/dataset.py
+++ b/qcfractal/interface/collections/dataset.py
@@ -3,7 +3,7 @@ QCPortal Database ODM
 """
 import itertools as it
 from enum import Enum
-from typing import Dict, List, Tuple, Union, Optional
+from typing import Any, Dict, List, Tuple, Union, Optional
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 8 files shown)


qcfractal/interface/collections/dataset.py
```diff
@@ -1,40 +1,29 @@
...
-    stoichiometry: Dict[str, Dict[str, float]]
+    molecule_id: str
+    comment: Optional[str] = None
+    local_results: Dict[str, Any] = {}
 
 
 class Dataset(Collection):
     """
-    This QCA Dataset class.
+    The Dataset class for compu.
```
Lnaden: "for compu."?

qcfractal/interface/collections/dataset.py
```diff
@@ -301,64 +234,48 @@ def query(self,
...
 
-        if not reaction_results and (self.client is None):
+        if not local_results and (self.client is None):
             raise AttributeError("DataBase: FractalClient was not set.")
 
         query_keys = {
-            "method": method.lower(),
-            "basis": basis.lower(),
-            "driver": driver.lower(),
+            "method": method,
```
Lnaden: Do you not need to cast these to lower anymore?

dgasmith: Slowly pushing this into the server itself so that we only have to worry about it in one spot.

qcfractal/interface/collections/dataset.py
dgasmith: Whoops, fixed.

dgasmith:
Future.

Issue #157, opened 2019-02-12
Allow queries with keys that are intentionally empty


doaa-altarawy:
Currently, for any query with the `keys=None` or `key=[ ]` then this key is not used in the query.
Instead, we would like to allow explicit setting of this key to None and look for this value in the DB.

Related to #151, we should be consistent in saving 'empty' fields. I suggest to make all empty `None`, and then you search for `options=None`.

dgasmith:
+1 for all empty is `None`.

doaa-altarawy:
I have two questions here. Firstly, which fields are most likely to use this?
Secondly, I had an issue setting default values in the functions since it's common to let then None, but now I set them to "".

It's not pretty either ways.

dgasmith:
For example if I do:
```
r = get_molecules(id="bad_id"])["data"]
>>> r
[]

>>> get_results(molecule_id=r)
>>> [every result in the database]
```

When in reality this chain of events should yield me no results. Effectively using:

```
query = {}
if query_field:
    query["query_field"] = query_field
```

is problematic. I agree that setting everything to `""` is not optimal, but we do need a signature like:
```
get_results(molecule_id=NOTSET, option_id=NOTSET, ...)
query = {}
if query_field is not NOTSET: # explicit is check
    query["query_field"] = query_field
```

so that we do not change the query. We can set this to a singleton of some sort, but using `None` is problematic as we use that as a key.

doaa-altarawy:
I implemented this is molecules, results, and keywords, but in collections, there was an issue with Pydantic web handlers because it automatically sets missing keys to None,
So get_collections()  is sent to socket as get_collection(name=None, collection=None)
and thus it returns an empty list instead of all collections.

dgasmith:
Hmm, good point. Should we instead have NOTSET be none (exactly as you have it now which is awesome!) and then handle options in a special way which is the only field where `None` is valid?

doaa-altarawy:
Do you mean rwplace NOSET with lowercase none?

Currently, fields that can be None:
Results.keywords
Results.basis
Procedures.keywords

Anything else?

dgasmith:
Those are all of the fields that can be None in the database. There *should* be no other cases that can be None.

[...]

PR #167, opened 2019-02-14
Allow queries with keys that are intentionally empty

doaa-altarawy:
## Description
Fixes issue #157 

## Questions
1. In results, only basis and keywords are optional, thus allowed to be None, other required fields can't be empty so there is no meaning to search for, say, program=None; it doesn't exist.
1. In molecule, it doesn't support returing the full list (i.e., get_molecule() -> all molecules up to the max_limit).
1. Collection has both fields (collection, name) as required, so no meaning to search for name=None. Also, the Web hander gets any empty query to None in Pydantic validation. Therefore, it assumes storage.get_collection(name=None, collection=None) to return ALL collections.

## Status
- [x] Ready to go

### DIFF
--- a/qcfractal/storage_sockets/mongoengine_socket.py
+++ b/qcfractal/storage_sockets/mongoengine_socket.py
@@ -470,7 +470,7 @@ class MongoengineSocket:
 
         return ret
 
-    def get_molecules(self, molecule_ids, index="id"):
+    def get_molecules(self, molecule_ids=None, index="id"):
 
         ret = {"meta": storage_utils.get_metadata(), "data": []}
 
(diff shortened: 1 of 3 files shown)


doaa-altarawy:
This is ready

qcfractal/storage_sockets/mongoengine_socket.py
```diff
@@ -1041,18 +1042,18 @@ def get_results(self,
...
+            query['basis'] = basis if basis != 'null' else None
+        if molecule is not None:
             query['molecule'], _ = _str_to_indices_with_errors(molecule)
-        if driver:
+        if driver is not None:
             query['driver'] = driver
-        if keywords:
-            query['keywords'] = keywords
+        if keywords is not None:
+            query['keywords'] = keywords if keywords != 'null' else None
```
dgasmith: I think this should be `keywords if keywords != '' else None` now?

doaa-altarawy: Yes of course

doaa-altarawy: I think we should remove this now since we already have the default: `""`

dgasmith: +1 for that.

dgasmith:
It seems `None` is fairly special to Mongo so we should keep these fields as None in the database. At the moment we are fixing this in the REST query. Which is not ideal. 

However, this fixes the core issue and keeps DB entires as None which I think is the primary thing today. I vote merging this in and opening an issue to think about this a bit more.

PR #168, opened 2019-02-15
[WIP] Save stub procedure when creating a service

doaa-altarawy:
## Status
- [ ] Ready to go

### DIFF
--- a/qcfractal/interface/models/torsiondrive.py
+++ b/qcfractal/interface/models/torsiondrive.py
@@ -156,7 +156,7 @@ class TorsionDrive(TorsionDriveInput):
             # Grab procedures
             needed_ids = [x for v in self.optimization_history.values() for x in v]
             objects = self.client.get_procedures({"id": needed_ids})
-            procedures = {v._id: v for v in objects}
+            procedures = {v.id: v for v in objects}
 
             # Move procedures into the correct order
             ret = {}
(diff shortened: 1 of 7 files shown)


dgasmith:
The service documents have a `output` attribute which can be used as the Procedure stub. This chunk of data is validated, but I think we need to add a status section to it.

doaa-altarawy:
Whenever the service is updated, does this output dict change?
Do we need to keep updating the corresponding procedure whenever we update a service?

dgasmith:
Currently the output dict does not change. Only at the very end. I do think it would be neat however to update the dict every iteration, but if complex we can add this as a TODO.

dgasmith:
So this passes locally, but something is going wrong on Travis. Error:
```
qcfractal/interface/models/torsiondrive.py:159: in <dictcomp>
    procedures = {v._id: v for v in objects}
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
self = <TorsionDrive(id='5c6b6df1f97e560d9f6156a8', success='False', initial_molecule='['5c6b6df1f97e560d9f6156a7']')>
name = '_id'
    @no_type_check
    def __getattr__(self, name):
        try:
            return self.__values__[name]
        except KeyError:
>           raise AttributeError(f"'{self.__class__.__name__}' object has no attribute '{name}'")
E           AttributeError: 'TorsionDrive' object has no attribute '_id'
```

This is wrong those procedure, id's should _only_ be optimizations and never torsiondrive objects. The logging shows we are not progressing in the torsion drive itself. So compute is going quite wrong:
```
Done True
1 initial tasks pushed to opt_queue
 - - +
Done True
1 initial tasks pushed to opt_queue
 - - +
```

doaa-altarawy:
I got this `_id` error locally with the openff environment.

dgasmith:
Ok, I updated my TorsionDrive and I see the same problem now.

PR #172, opened 2019-02-18
Pydantic bumps

dgasmith:
## Description
Bumps pydantic version requirements to 0.20. We were using lists instead of sets which is was not correct, but worked until the recent bump.

This will require QCElemental v0.2.6 which is not yet released.

## Status
- [ ] Ready to go


dgasmith:
Going to merge this to get other PRs working. @Lnaden can you give this a review even if it is merged?

PR #173, opened 2019-02-19
Services: API overhaul for using only ids

dgasmith:
## Description
Finishes the conversion so that Services look similar to all other API's. That is, submit a list of services, get the id of the result back (in this case a procedures id).

## Status
- [x] Ready to go

### DIFF
--- a/qcfractal/interface/collections/collection.py
+++ b/qcfractal/interface/collections/collection.py
@@ -57,12 +57,6 @@ class Collection(abc.ABC):
         provenance: Dict[str, str] = {}
         id: str = 'local'
 
-        class Config:
-            # Allows extra args to prevent subclass errors, but just ignore them
-            # These might be the defaults, but make it explicit to be sure
-            allow_extra = False
-            ignore_extra = True
-
     def __str__(self) -> str:
         """
         A simple string representation of the Collection.
--- a/qcfractal/interface/collections/openffworkflow.py
+++ b/qcfractal/interface/collections/openffworkflow.py
@@ -2,12 +2,12 @@
 """
 
 import copy
-from typing import Dict, Any
+from typing import Any, Dict
 
-from . import collection_utils
+from ..models.torsiondrive import TorsionDrive, TorsionDriveInput
+from ..orm import OptimizationORM
 from .collection import Collection
-from .. import orm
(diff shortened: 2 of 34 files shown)


dgasmith:
@Lnaden @loriab Can I get another review on the last few changes?

qcfractal/interface/collections/collection.py
```diff
@@ -57,12 +57,6 @@ class DataModel(BaseModel):
         provenance: Dict[str, str] = {}
         id: str = 'local'
 
-        class Config:
```
Lnaden: Did you want datamodel to be mutable?

qcfractal/interface/collections/openffworkflow.py
```diff
@@ -85,14 +85,14 @@ class DataModel(Collection.DataModel):
             }
         }
         torsiondrive_static_options: Dict[str, Any] = {
-            "torsiondrive_meta": {},
-            "optimization_meta": {
+            "keywords": {},
```
Lnaden: This is fine, more from a user perspective, do we have anyone we need to tell about this change?

qcfractal/interface/models/gridoptimization.py
```diff
@@ -53,26 +53,20 @@ def check_step_type(cls, v):
...
-
     class Config:
+        extra = "forbid"
         allow_mutation = False
 
 
-class OptOptions(BaseModel):
+class GOKeywords(BaseModel):
     """
     GridOptimization options
```
Lnaden: This docstring now mimics the docstring in OptimizationSpecification, one of them should be changed to better match the true description

qcfractal/interface/collections/openffworkflow.py
dgasmith: Not yet, no.

qcfractal/interface/models/gridoptimization.py
dgasmith: Not quite `OptimizationSpecification` includes Keywords and other things. The `GOKeywords` are explicitly just keywords.

qcfractal/interface/collections/collection.py
dgasmith: This is never called so its attributes do not matter as Config is no inherited. But in general, yes the datamodel should be mutable.

PR #175, opened 2019-02-21
Storage Keyword and Molecule reworks

### LINKED ISSUE #160: Keywords as Foriegn key
Use `Keywords` are `ReferenceField` instead of a `StringField` (better validation). Also, why not also use this ReferenceField in Procedures.

Finally, I think the `Keywords` table doesn't need the program field as it can simply be included when creating the `hash_index` itself.

dgasmith:
## Description
This PR accomplishes the following:
 - Molecule querying now uses the standard `get_molecule(id=[])` syntax.
 - Reworks keywords to be a LazyReferenceField where possible, fixes #160.
 - Removes `program` from Keyword hash as it is not required for searching.
 - Services now have the standard status messages as the Task Queue.
 - Folks ME Storage parsing into a single function.

## Status
- [ ] Changelog updated
- [x] Ready to go


dgasmith:
@doaa-altarawy This is ready for merge, please review and merge at your leisure.

dgasmith:
Thanks, in we go!

PR #177, opened 2019-02-21
[WIP] Attempt to implement the Manager Helpers

Lnaden:
## Description

This attempts to implement Dask-JobQueue manager helper script.

Not yet tested. but I wanted to get this up here so we can talk about it and refine as needed.

## Todos
  - [ ] Test this on the ARC cluster

## Questions
- [x] Do these flags look reasonable.
- [x] I've set this up to handle a single job submission, will Dask setup multiple submissions as needed?
- [ ] What is the correct way to setup environments (e.g. Conda) for the workers once the Dask Job starts them?

## Status
- [ ] Changelog updated
- [ ] Ready to go


manager_helpers/dask_manager_template.py
```diff
@@ -0,0 +1,106 @@
...
+- Manager running on the head node
+- SLURM manager
+
+For additional information about the Dask Job Queue, please visit this site:
+https://jobqueue.dask.org/en/latest/
+"""
+
+# Fractal Settings
+# Location of the Fractal Server you are connecting to
+FRACTAL_ADDRESS = "localhost:7777"
```
dgasmith: We need username/password/update_frequency/verify here as well. Storing the password as plaintext makes me a bit nervous...

manager_helpers/dask_manager_template.py
```diff
@@ -0,0 +1,106 @@
...
+
+# How many cores per node you want your jobs to have access to
+CORES_PER_NODE = 0
+# How much memory per node (in GB) you want your jobs to have access to
+MEMORY_PER_NODE = 0
+# How many tasks per node you want to execute on
+MAX_TASKS_PER_NODE = 0
+# Maximum number of nodes to try and consume
```
dgasmith: Try to reword slightly, `consume` doesn't really resonate too well I think.  use/utilize maybe?

manager_helpers/dask_manager_template.py
```diff
@@ -0,0 +1,106 @@
...
+
+For additional information about the Dask Job Queue, please visit this site:
+https://jobqueue.dask.org/en/latest/
+"""
+
+# Fractal Settings
+# Location of the Fractal Server you are connecting to
+FRACTAL_ADDRESS = "localhost:7777"
+
+# How many cores per node you want your jobs to have access to
```
dgasmith: Maybe something with the connotations: How many cores per node (at most the number of cores on a node). Same for memory. "access" seems a bit different.

manager_helpers/dask_manager_template.py
```diff
@@ -0,0 +1,106 @@
...
+manager = qcfractal.queue.QueueManager(client, dask_client, update_frequency=0.5,
+                                       cores_per_task=CORES_PER_NODE // MAX_TASKS_PER_NODE,
+                                       memory_per_task=MEMORY_PER_NODE // MAX_TASKS_PER_NODE)
+
+# Important for a calm shutdown
+from qcfractal.cli.cli_utils import install_signal_handlers
+install_signal_handlers(manager.loop, manager.stop)
+
+# Start the loop
+manager.start()
```
dgasmith: Can we have a `TEST` option which will run `manager.test()` over `manager.start()` if true, or is this too much?

qcfractal/cli/qcfractal_template_generator.py
```diff
@@ -0,0 +1,438 @@
...
+manager = qcfractal.queue.QueueManager({FRACTAL_TEST_NONE}, {MANAGER_CLIENT}, update_frequency=0.5,
+                                       cores_per_task=CORES_PER_NODE // MAX_TASKS_PER_NODE,
+                                       memory_per_task=MEMORY_PER_NODE // MAX_TASKS_PER_NODE)
+
+# Important for a calm shutdown
+from qcfractal.cli.cli_utils import install_signal_handlers
+install_signal_handlers(manager.loop, manager.stop)
+
+# Start or test the loop. Swap with the .test() and .start() method respectively
+manager.{TEST_OR_START}()
```
dgasmith: Can we change this so that we have a variable like `RUNTEST = True` or similar so that a user can setup an entire manager script, test it, then spin up a real manager without having to re-render?

Lnaden: Sure! I think that's a great idea.

Lnaden:
I think I fixed all the the obvious syntax issues testing on our local cluster (SLURM-based), this includes having a unified set of inputs and parsing them in the Manager backends. However, I can't get any of the tests to actually execute to completion.

Dask:
Starts up jobs which live for about 2 seconds and then receive the "cancel" command per this line in the slurm output file (to stdout, this is the complete log file): 
```
slurmstepd: error: *** JOB {slurm job id} ON {node} CANCELLED AT 2019-02-25T12:28:18 ***
```
Two of the slurm jobs appear to start nannys, but they are not the first jobs, nor the last ones, just some intermediate jobs (this is also to stdout, and the complete log file):
```
distributed.nanny - INFO -         Start Nanny at: 'tcp://{ip}:{port}'
```
so the behavior is very strange. Dask also does not appear to be reporting a heartbeat back.

Parsl:
Spins up a block with workers, all the workers, the manager, the interchange, and the parsl log report starting normally, then nothing happens. They just wait. The heartbeat keeps going and eventually another block is started, but im worried multiple blocks would start and go nuts if I just let it run, and no work is ever returned.

Any ideas here @dgasmith ?

dgasmith:
If you look at the logs the workers are unable to connect back to the scheduler and kill themselves off after 10-15 seconds. This is due to the fact that ARC decided `eth0` was not a cool name so they used `bond0.4049`/`bond0.4002` for their interface ID. It doesn't appear Parsl can change the interface name (which also mean no ib0/infiniband) unfortunately.

dgasmith:
@Lnaden Looks like one combo is generating incorrect code, I can patch this tomorrow if needed.

Lnaden:
I can fix it

Lnaden:
That should fix it. `dedent()` doesn't like `\n` chars as part of the sting, so I had to escape them.

Yay tests!

PR #181, opened 2019-02-26
Add pagination to queries

doaa-altarawy:
This is done by adding `limit` and `skip`. Here `skip` is a number of results to skip not number of pages. Alternatively, this can be done by specifying page numbers inserted of skipping results.

An important note here is that the total number of results (`result`, `procedure`, `molecule`, etc) is returned in `meta['n_found']` and the size of the data itself is, obviously, not the size of all possible results.

## Todos
  - [x] Add tests

## Status
- [x] Ready to go


dgasmith:
Going to pull this in, failures due to QCA ecosystem disruption.

PR #189, opened 2019-03-02
Record objects as base document classes

### LINKED ISSUE #178: Database KVStore
We store several large pieces of data in the results/procedures table that are accessed very infrequently such as the execution log. We should create a id-value store so that these can be lazily referenced and not within the primary results/procedures document. This has the benefit of:
1) Not bloating an individual document or row allowing for better packing both on-disk and in-memory.
2) Opens the door to keep the results/procedures documents on fast disk (SSD) while shipping the logging information out to slow disk (spinning disk) as we will operate under the assumption that this data is accessed much less frequently.

dgasmith:
## Description
This PR will accomplish the following:
 - [x] Migrates the database high level schemas to Pydantic Record objects.
   - Can produce standard QCSchema and merge their data with output schema.
   - Handle all hash index formation.
   - Controls all status, stdout, task_id, and error information.  
- [x] The Record object is now the base schema for Result, Optimization, TorsionDrive, GridOptimization, and all future procedure documents.
- [x] Partial document updates will be removed in favor of full document saves.
- [x] Task documents are combined with results from managers before adding to the database.
- [x] Deprecates old hooks code, in preparation for future observer/signals.
- [x] Moves database interface to only accept models.
- [x] Adds a key/value store and move stdout/stderr to this store, fixes #178.

## Status
- [x] Ready to go


dgasmith:
Ready for review!

qcfractal/interface/client.py
```diff
@@ -240,18 +242,46 @@ def add_molecules(self, mol_list: List[Molecule], full_return: bool = False) ->
...
+        -------
+        List[KeywordSet]
+            The requested KeywordSet objects.
+        """
+        body = KeywordGETBody(meta={}, data=id)
         r = self._request("get", "keyword", data=body.json())
         r = KeywordGETResponse.parse_raw(r.text)
 
-        return r.data
+        if full_return:
```
loriab: I rather prefer the `return_full` pattern established by https://github.com/MolSSI/QCElemental/blob/master/qcelemental/molparse/from_string.py#L17 and `return_wfn` but little matters.

qcfractal/interface/models/common_models.py
```diff
@@ -35,6 +34,7 @@ class DriverEnum(str, Enum):
     energy = 'energy'
     gradient = 'gradient'
     hessian = 'hessian'
+    properties = 'properties'
```
loriab: always wondered about the necessity of an extra "properties" option. is it mostly for psi, where run through a different fn, or do you generally regard it as a separate calc type?

qcfractal/procedures/procedures.py
```diff
@@ -53,15 +56,16 @@ def parse_input(self, data):
         new_tasks = []
         results_ids = []
         existing_ids = []
-        for inp in inputs:
-            if inp is None:
+        for mol in moleucle_list:
```
loriab: probable typo

qcfractal/storage_sockets/mongoengine_socket.py
```diff
@@ -179,34 +173,98 @@ def _clear_db(self, db_name: str):
...
-            Result.drop_collection()
-            Molecule.drop_collection()
-            Keywords.drop_collection()
-            Collection.drop_collection()
-            TaskQueue.drop_collection()
-            QueueManager.drop_collection()
-            Procedure.drop_collection()
-            User.drop_collection()
+            ResultORM.drop_collection()
+            MoleculeORM.drop_collection()
```
loriab: good, there's getting to be `Molecule`s everywhere.

qcfractal/interface/models/common_models.py
dgasmith: Not entirely sure, I view that as more experimental and we will see how it evolves.

dgasmith:
@doaa-altarawy Can you give this a once over when you have a moment?

dgasmith:
Ok, final changes in. @doaa-altarawy please review when you have a moment.

PR #197, opened 2019-03-07
General tweaks from live evaluation

dgasmith:
## Description
This PR provides a host of tweaks and bug fixes from using the code at a production level. These are:
 - FractalClient now automatically connects to `api.qcarchive.molssi.org`
 - Fixes a bug where config-file data was not respected.
 - Removed override from the `collection.save()` feature which is now automatically used correctly on updates.
 - Several hash changes due to fixes in QCElemental v0.3.1
 - `_fractal_tags` were leaking into results and are now properly removed.
 - Removes stoic from the base dataset where this is not used.
 - QueueManager CLI is now reduced to just a PoolExecutor for now.
 - Bumps ecosystem required versions.

## Status
- [ ] Changelog updated
- [x] Ready to go

### DIFF
--- a/qcfractal/cli/qcfractal_manager.py
+++ b/qcfractal/cli/qcfractal_manager.py
@@ -6,6 +6,7 @@ import argparse
 
 import qcfractal
 import tornado.log
+import qcengine as qcng
 
 from . import cli_utils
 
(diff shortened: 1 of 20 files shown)


qcfractal/cli/qcfractal_manager.py
```diff
@@ -6,64 +6,56 @@
...
+        type=int,
+        help="The number of simultaneous tasks for the executor to run, resources will be divided evenly.")
+    executor.add_argument("--cores", type=int, help="The number of process for the executor")
+    executor.add_argument("--memory", type=int, help="The total amount of memory on the system in GB")
 
     # FractalClient options
-    parser.add_argument(
+    server = parser.add_argument_group('FractalServer connection settings')
+    server.add_argument(
         "--fractal-uri", type=str, default="localhost:7777", help="FractalServer location to pull from")
```
Lnaden: Do you want to change the default from localhost here as well?

qcfractal/cli/qcfractal_manager.py
```diff
@@ -74,47 +66,54 @@ def main(args=None):
...
         from concurrent.futures import ProcessPoolExecutor
 
-        queue_client = ProcessPoolExecutor(max_workers=args["nprocs"])
+        queue_client = ProcessPoolExecutor(max_workers=args["ntasks"])
 
     else:
-        raise KeyError("Unknown adapter type '{}', available options: 'fireworks', 'dask', .".format(
-            args["adapter_type"]))
+        raise KeyError(
+            "Unknown adapter type '{}', available options: 'fireworks', 'dask', .".format(args["adapter_type"]))
```
Lnaden: Probably should change this since you disabled fireworks and dask for now

qcfractal/cli/qcfractal_manager.py
dgasmith: No, QCPortal is user facing. Anything here will be developer facing and compute cannot be arbitrarily hooked up to our central server.

qcfractal/cli/qcfractal_manager.py
dgasmith: Good point. Lets ear mark to delete later.

PR #203, opened 2019-03-08
Client Documentation

dgasmith:
## Description
Attempt to move around the Client documentation into something reasonable. Still a ways to go, but good structure I think where we break it down into three categories:

 - Add/Query Objects - For all whole object transfers such as Molecule, KeywordSet, etc
 - Records Querying - Querying of compute records, including partial queries and projections.
 - New Compute - New compute tasks and inspection of the current queue.

Also renames `FractalClient.get_*` to `FractalClient.query_*` to better reflect their function. The exception is `get_collection` as that isn't really a query, but a get of a single known collection.

## Status
- [x] Ready to go


docs/source/client.rst
```diff
@@ -0,0 +1,38 @@
...
+.. code-block:: python
+
+    >>> import qcportal as ptl
+    >>> client = ptl.FractalClient("localhost:8888")
+    >>> client
+    FractalClient(server='http://localhost:8888/', username='None')
+
+The ``FractalClient`` handles all communication to the ``FractalServer`` from
+the Python API layer. This includes adding new molecules, computations,
+collections, and querying for records.
```
Lnaden: I think it would be worth adding here something to the effect of

"Fractal client without any arguments will automatically connect to the MolSSI QCArchive server if it can in read-only mode"

docs/source/client.rst
```diff
@@ -0,0 +1,38 @@
...
+that addresses and username do not have to be retyped for every line and
+reduces the chance that a username and password could accidentally be added to
+a version control system. Creation from file uses the classmethod
+``FractalClient.from_file()``, by default the client searches for a
+``qcportal_config.yaml`` file in either the current working directory or from
+the canonical ``~/.qca`` folder.
+
+.. toctree::
+   :maxdepth: 1
+   :hidden:
```
Lnaden: Do we really want this toc tree hidden, or is this a "for now" thing until we flesh out more of the options?

docs/source/client-add-query.rst
```diff
@@ -0,0 +1,84 @@
...
+               H         -0.972396644160    -0.787126321701     0.424194864034
+
+
+    >>> data = client.add_molecules([hooh])
+    >>> data
+    ['5c82c51895d5923b946989c1']
+
+
+Molecules can either be queried from their Molecule ID or Molecule
+hash:
```
Lnaden: Do you think a new `:term:` for hash might be helpful, just for explaining how hash and id are different, especially since users are not really going to know nor care how the data on the back-end is accessed and stored beyond "its indexed and queriable."

qcfractal/tests/test_storage.py
```diff
@@ -416,7 +416,7 @@ def test_empty_get(storage_results):
     assert 0 == len(storage_results.get_results(program='null')['data'])
 
 
-def test_results_query_total(storage_results):
+def test_results_get_total(storage_results):
 
     assert 6 == len(storage_results.get_results()["data"])
```
Lnaden: These seem more like queries to me, I know that we are explicitly getting data, but its conditioned on broad terms, not specific fetches. No need to change them, but just a terminology confusion.

dgasmith: They are, will look into storage rename after we switch out SQL.

docs/source/client.rst
dgasmith: Nested TOC trees are messy unless hidden and at the top of the base file. They still correctly show up in the side bar.

PR #211, opened 2019-03-12
Queue Manager now reads Config File

### LINKED ISSUE #205: Queue Manager Template
Currently missing the following fields:
 - username
 - password
 - queue_tag

Lnaden:
## Description
Added manager config files to the CLI, ~looks in `./qcf_manager_config.yaml` and `~/.qca/qcf_manager_config.yaml`~ falling back to the defaults in the pydantic models, but individual options can be overwritten through the CLI.

Individual options are also searched through environment variables looking for `QCA_XXXXX` where `XXXXXX` is the variable name.

Some of CLI variables have been changed, as have their defaults. Some names have been changed in kwargs to be more clear what they do, namely I have renamed the `QueueManager`'s `cluster` kwarg to `manager_name`.

Dask JobQueue can now be set up through the config file + cli. If no config file is found (or an insufficient number of environment variables are set), then only the `ProcessPoolExecutor` can be run.

Ready for feature and function/design review, but still needs more work on new unit tests and testing in a live environment.

This also fixes #205 while I was in there

## Todos
  - [x] Fix all tests
  - [x] Figure out how to test the Dask distributed (edit: will do this in post, Pool works)

## Status
- [x] Changelog updated (edit: due to #212, changelog edits in PRs before that are unsafe, will do before next release)
- [x] Ready to go

### Edits:
* See individual edit notes above
* No automatic queue manager template can be found now, it MUST be specified or only use CLI (in which case its Pool only)

### DIFF
--- a/qcfractal/cli/cli_utils.py
+++ b/qcfractal/cli/cli_utils.py
@@ -10,12 +10,14 @@ import signal
 import yaml
 
 
-def import_module(module):
+def import_module(module, package=None):
     """Protected import of a module
     """
     try:
-        ret = importlib.import_module(module)
+        ret = importlib.import_module(module, package=package)
     except ImportError:
+        if package is not None:
+            raise ImportError("Requested module/package '{}/{}' not found.".format(module, package))
         raise ImportError("Requested module '{}' not found.".format(module))
 
     return ret
@@ -44,13 +46,14 @@ def argparse_config_merge(parser, parsed_options, config_options, parser_default
     ----------
     parser : ArgumentParser
     config_options : dict
-    parser_default : None, optional
+    parser_default : List, Optional
     """
     config_options = copy.deepcopy(config_options)
 
     default_options = vars(parser.parse_args(args=parser_default))
     diff = config_options.keys() - default_options.keys()
     if diff:
+        import argparse
         raise argparse.ArgumentError(None,
                                      "Unknown arguments found in configuration file: {}.".format(", ".join(diff)))
 
--- a/qcfractal/cli/cli_utils.py
+++ b/qcfractal/cli/cli_utils.py
@@ -33,13 +33,16 @@ def read_config_file(fname):
     else:
         raise TypeError("Did not understand file type {}.".format(fname))
 
-    with open(fname, "r") as handle:
-        ret = rfunc(handle)
+    try:
+        with open(fname, "r") as handle:
+            ret = rfunc(handle)
+    except FileNotFoundError:
+        raise FileNotFoundError("No config file found at {}.".format(config_file))
 
     return ret
 
 
-def argparse_config_merge(parser, parsed_options, config_options, parser_default=None):
+def argparse_config_merge(parser, parsed_options, config_options, parser_default=None, check=True):
     """Merges options between a configuration file and a parser
 
     Parameters
@@ -50,11 +53,12 @@ def argparse_config_merge(parser, parsed_options, config_options, parser_default
     """
     config_options = copy.deepcopy(config_options)
 
-    default_options = vars(parser.parse_args(args=parser_default))
-    diff = config_options.keys() - default_options.keys()
-    if diff:
-        raise argparse.ArgumentError(None,
-                                     "Unknown arguments found in configuration file: {}.".format(", ".join(diff)))
+    if check:
+        default_options = vars(parser.parse_args(args=parser_default))
+        diff = config_options.keys() - default_options.keys()
+        if diff:
+            raise argparse.ArgumentError(None,
+                                         "Unknown arguments found in configuration file: {}.".format(", ".join(diff)))
 
     # Add in parsed options
     for k, v in parsed_options.items():
--- a/qcfractal/cli/cli_utils.py
+++ b/qcfractal/cli/cli_utils.py
@@ -2,6 +2,7 @@
 Utilities for CLI programs
 """
 
+import argparse
 import copy
 import importlib
 import json
@@ -15,11 +16,10 @@ def import_module(module, package=None):
     """
     try:
         ret = importlib.import_module(module, package=package)
-    except ImportError:
+    except ModuleNotFoundError:
         if package is not None:
-            raise ImportError("Requested module/package '{}/{}' not found.".format(module, package))
-        raise ImportError("Requested module '{}' not found.".format(module))
-
+            raise ModuleNotFoundError("Requested module/package '{}/{}' not found.".format(module, package))
+        raise ModuleNotFoundError("Requested module '{}' not found.".format(module))
     return ret
 
 
@@ -53,7 +53,6 @@ def argparse_config_merge(parser, parsed_options, config_options, parser_default
     default_options = vars(parser.parse_args(args=parser_default))
     diff = config_options.keys() - default_options.keys()
     if diff:
-        import argparse
         raise argparse.ArgumentError(None,
                                      "Unknown arguments found in configuration file: {}.".format(", ".join(diff)))
 
--- a/qcfractal/cli/qcfractal_manager.py
+++ b/qcfractal/cli/qcfractal_manager.py
@@ -64,6 +64,8 @@ class SchedulerEnum(str, Enum):
     slurm = "slurm"
     torque = "torque"
     pbs = "pbs"
+    sge = "sge"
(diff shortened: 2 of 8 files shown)


qcfractal/cli/qcfractal_manager.py
```diff
@@ -64,83 +186,148 @@ def main(args=None):
...
     exit_callbacks = []
-    args["adapter_type"] = "executor"
 
-    # Handle Dask adapters
-    # if args["adapter_type"] == "dask":
-    #     dd = cli_utils.import_module("distributed")
+    # Try to read a config file first
+    config_file = args["config_file"]
+    if type(config_file) is str:  # Cast to unified code for same list
+        config_file = [config_file]
```
dgasmith: This is a case where a config file should be passed in explicitly I think.

qcfractal/cli/qcfractal_manager.py
```diff
@@ -64,83 +186,148 @@ def main(args=None):
...
-
     exit_callbacks = []
-    args["adapter_type"] = "executor"
 
-    # Handle Dask adapters
-    # if args["adapter_type"] == "dask":
-    #     dd = cli_utils.import_module("distributed")
+    # Try to read a config file first
+    config_file = args["config_file"]
+    if type(config_file) is str:  # Cast to unified code for same list
```
dgasmith: Always `isinstance`.

qcfractal/cli/qcfractal_manager.py
```diff
@@ -64,83 +186,148 @@ def main(args=None):
...
-    #         queue_client = dd.Client(args["dask_uri"])
+    if settings.common.test:
+        # Test, nothing needed
+        client = None
+    elif settings.server.file is not None:
+        # Gave a file? Neat!
+        client = qcfractal.interface.FractalClient.from_file(settings.server.file)
+    elif settings.server.specified:
+        # Specified something? Okay!
+        client = qcfractal.interface.FractalClient(settings.server.dict(exclude={"file"}, skip_defaults=True))
```
dgasmith: `**settings.server...`

qcfractal/cli/tests/test_cli.py
```diff
@@ -34,7 +34,7 @@ def active_server(request):
     with testing.popen(args, **_options) as server:
         time.sleep(2)
 
-        server.test_uri_cli = "--fractal-uri=localhost:" + port
+        server.test_uri_cli = "--fractal-address=localhost:" + port
```
dgasmith: This is a URI not an address, please revert.

qcfractal/cli/qcfractal_manager.py
```diff
@@ -12,16 +16,136 @@
...
+    adapter: AdapterEnum = AdapterEnum.pool
+    log_file_prefix: str = None
+    queue_tag: str = None
+
+    class Config(SettingsCommonConfig):
+        pass
+
+
+class FractalServerSettings(BaseSettings):
+    file: str = None
```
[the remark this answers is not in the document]
Lnaden: Or we just assume we try to read and overwrite from cli, its what most programs which use a `~/.{folder}/config` file do anyways, no need to make it more complex, especially since that means we would have to engineer a `False` condition

qcfractal/cli/qcfractal_manager.py
```diff
@@ -12,16 +16,136 @@
...
+
+
+class AdapterEnum(str, Enum):
+    dask = "dask"
+    pool = "pool"
+
+
+class CommonManagerSettings(BaseSettings):
+    ntasks: int = 1
+    cores: int = qcng.config.get_global("ncores")
```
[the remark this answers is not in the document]
Lnaden: But this is what we do right now in the existing code. Unless you are suggesting setting this to `None` and let Engine by default handle it?

qcfractal/cli/cli_utils.py
```diff
@@ -10,12 +10,14 @@
 import yaml
 
 
-def import_module(module):
+def import_module(module, package=None):
     """Protected import of a module
     """
     try:
-        ret = importlib.import_module(module)
+        ret = importlib.import_module(module, package=package)
```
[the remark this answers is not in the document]
Lnaden: Do we even need this anymore since all we are doing is trapping an import error, and raising an import error with effectively the same message?

qcfractal/cli/qcfractal_manager.py
Lnaden: It is, I just have this engineered as such because `config_file` (the Manager Settings) defaults to search in the 2 default locations through 
`add_argument("config_file", nargs="?", type=str,
                        default=[os.path.join(os.path.realpath("."), MANAGER_CONFIG_NAME),
                                 os.path.join(os.path.expanduser("~"), ".qca", MANAGER_CONFIG_NAME)])`
which returns a list, but if the user feeds in one arg, it returns a string and casting to list by `list(x)` with a list/tuple is fine, but it will turn a string into a list of 1-char long elements. By doing this, the code below is then identically handled

qcfractal/cli/tests/test_cli.py
Lnaden: Would it be better to call the `FractalClient`' kwarg `uri` instead of `address`? This would help make the term more uniform if we want to be strict about calling it `uri`.

If thats the case, would you accept the alias of `address`?

dgasmith: `FractalClient` can be a URI or address, this *must* be a URI.

qcfractal/cli/qcfractal_manager.py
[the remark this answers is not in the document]
dgasmith: Right, this is fine to set for Executor, but not for any other distributed backend.

qcfractal/cli/qcfractal_manager.py
dgasmith: For a manager, like a server we do not want to have default configs. They must be passed in explicitly.

Lnaden:
Some notable changes since inception:

* No automatic queue manager template can be found now, it MUST be specified or only use CLI (in which case its Pool only)
* We have agreed on using the URI here in the scripts, but leave the `address` kwarg in Client (for now), most use cases use it as a positional arg auto assigning to the kwarg, but considered that too drastic of a change for this pr
* You cannot target a `from_file` for the Fractal Client anymore, must be specified
* I added a `--no-verify` and a `--verify` flags (mutually exclusively) to the CLI because the three-way logic of `FractalClient(verify={True, False, None -> whatever the default is}) with only one option of just  a `--no-verify` flag was beyond me.

** Important **:

This PR will be *untested* for Dask JobQueue, but still works for the pool executor. Any bugs introduced with the jobqueue will have to be tested manually, which I will be doing after this is merged in. So there may be a future pr fixing those things. 

Otherwise, this should be ready to go if tests pass (and approval)

Lnaden:
Hang on on this, some other comments I just saw due to GH collapsing the conversation. More edits coming

dgasmith:
Reworked the pydantic and config file parsing a bit. Let me know what you think.

Lnaden:
So I allow the `DaskQueueSettings` to accept additional kwargs, so we can actually have pass throughs to the constructor. 

I also had 2 outstanding questions:

* Do you want me to add the other `*Cluster` classes in here for users?
* Do you want me to remove the PBS<->Torque equivalency mappings and say "pbs or bust?"

Otherwise, made a couple small cleanups.

Lnaden:
I did both questions (added SGE and Moab),  added warning about the fact that torque is mapped to PBS

Lnaden:
I think your changes are good, @dgasmith, and if you think my tweaks are good, we can merge this

qcfractal/cli/qcfractal_manager.py
```diff
@@ -243,7 +242,7 @@ def main(args=None):
         _cluster_loaders = {"slurm": "SLURMCluster", "pbs": "PBSCluster", "torque": "PBSCluster"}
 
         # Create one construct to quickly merge dicts with a final check
-        dask_construct = _DaskQueueSettingsNoCheck(
+        dask_construct = dict(
```
dgasmith: We try to use literals where possible in the code please rewrite.

Lnaden:
Rebased from master for the doc structure change

PR #215, opened 2019-03-14
Dask fixes

### LINKED ISSUE #214: Ensure Dask JobQueue works for qcfactal-manager
We introduced #211 without being able to fully test dask jobqueue on a real cluster, although we know the ProcessPoolExecutor works

This issue is for us (mainly me) to track the progress of this and mark when it is complete.

Lnaden:
## Description

After doing some real testing and with the help of @dgasmith, this PR should get the Dask JobQueue working (tested with SLURM). We also found a non-failing infinite loop condition which was causing problems (See dask/dask-jobqueue\#249)

This also fixed a deprecation warrning from PyYaml (https://github.com/yaml/pyyaml/wiki/PyYAML-yaml.load(input)-Deprecation) and made it possible to create a large number of test cases for the queue manager `test()` method.

Fixes #214 and resolves the outstanding issue in #211

## Status
- [x] Changelog updated
- [x] Ready to go

### DIFF
--- a/qcfractal/cli/qcfractal_manager.py
+++ b/qcfractal/cli/qcfractal_manager.py
@@ -58,6 +58,7 @@ class QueueManagerSettings(BaseSettings):
     log_file_prefix: str = None
     update_frequency: float = 30
     test: bool = False
+    ntests: int = 1
 
 
 class SchedulerEnum(str, Enum):
@@ -86,9 +87,10 @@ class DaskQueueSettings(BaseSettings):
 
     def __init__(self, **kwargs):
         """Enforce that the keys we are going to set remain untouched"""
-        bad_set = set(kwargs.keys()) - {
-            "name", "cores", "memory", "queue", "processes", "walltime", "env_extra", "qca_resource_string"
+        forbidden_set = {
+            "name", "cores", "memory", "processes", "walltime", "env_extra", "qca_resource_string"
         }
+        bad_set = set(kwargs.keys()) & forbidden_set
         if bad_set:
             raise KeyError("The following items were set as part of dask_jobqueue, however, "
                            "there are other config items which control these in more generic "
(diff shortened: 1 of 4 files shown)


qcfractal/cli/qcfractal_manager.py
```diff
@@ -148,6 +150,7 @@ def parse_args():
     # Additional args
     optional = parser.add_argument_group('Optional Settings')
     optional.add_argument("--test", action="store_true", help="Boot and run a short test suite to validate setup")
+    optional.add_argument("--ntests", type=int, help="How many tests per found program to run, does nothing without --test set")
```
dgasmith: Default to one here?

qcfractal/cli/qcfractal_manager.py
```diff
@@ -205,7 +207,13 @@ def main(args=None):
         client = None
     else:
         # Connect to a specified fractal server
-        print(settings.server.fractal_uri, settings.server.dict(skip_defaults=True, exclude={"fractal_uri"}))
+        print_string = "Attempting to connect to FractalClient at: {}".format(settings.server.fractal_uri)
+        other_data = settings.server.dict(skip_defaults=True, exclude={"fractal_uri"})
+        if "password" in other_data:
+            other_data["password"] = "{provided}"
+        if other_data:
+            print_string += "\n   With credentials: {}".format(other_data)
```
dgasmith: This is printed internally within the Manager. I don't think we want to duplicate this here.

qcfractal/cli/qcfractal_manager.py
Lnaden: It does

https://github.com/MolSSI/QCFractal/pull/215/files#diff-d72c13bbbcebb1df47f9d2d3f874c558R61

qcfractal/cli/qcfractal_manager.py
Lnaden: Okay, there was an outstanding print statement, I think from debugging now that I think about it, but it was showing the password too. I'll remove it

dgasmith: Just we shouldn't use print. Just logging and all of that is initialize in the manager itself.

PR #216, opened 2019-03-14
Smart Manager job pulls and priority queue

### LINKED ISSUE #13: Handle import errors on compute
If a worker logs an import error (fails to import RDKit) the result is currently marked as failed. ImportErrors (ModuleNotFound) should be detected and resubmitted to the queue as a "soft" error up to a maximum of `n` times. Matching of tasks to corresponding workers with arbitrary queue backends will continue to be an issue under the current structure.

### LINKED ISSUE #124: Task/Worker Matching
Currently we depend on a user to correctly spin up workers that can cover all tasks. As we expand the number of supported programs that probability of this will steadily decrease. Currently the queue knows which programs is needed by a task and the manager understands which programs it has. If only one program was required this would be a pretty simple in operation, but occasionally multiple programs are required.

 - Task1: ["psi4", "geometric"]
 - Task2 : ["psi4"]

- Manager1: ["psi4"]
- Manager2: ["psi4", "geometric"]

The system is quasi load balancing as is so that isn't too much of a concern at the moment. The real question is if we can do this quickly as most `$in` operations will need to scan the entire DB.

dgasmith:
## Description
This PR overhauls the primary queue to take into account several new items:
 - [x] There is now HIGH/MEDIUM/LOW priority jobs so that HIGH priority jobs can "skip" ahead.
 - [x] Managers now tell the Server what programs and procedures they have as well as version information.
 - [x] Managers now only pull tasks that they have the known available compute for, fixes #13, fixes #124.
 - [x] Tasks are now `TaskRecord` pydantic objects and fixed associated errors with wrong schema there.

The full query is now:
 - Check for programs/procedures so tasks do not go to managers that cannot support them.
 - Narrow remaining possible tasks by tag (if present)
 - Sort by (priority, created_date) so that all HIGH are pulled in order before all MEDIUM and then finally LOW.

Ready for review and merge. Still some edge case to go through and cleanup. Would rather do this in another PR to prevent this from becoming titanic.

CI may fail and may need: https://github.com/MolSSI/QCEngine/pull/50. Not sure why this is just showing up now and only for Psi4 gradient computations.

## Status
- [ ] Changelog updated
- [x] Ready to go


qcfractal/services/service_util.py
```diff
@@ -112,6 +116,8 @@ class BaseService(BaseModel, abc.ABC):
     procedure_id: Optional[ObjectId] = None
 
     # Task manager
+    task_tag: Optional[str] = None
+    task_priority: PriorityEnum
     task_manager: TaskManager = TaskManager()
 
     status: str = "WAITING"
```
Lnaden: Should this call the `TaskStatusEnum` model?

PR #219, opened 2019-03-17
Project Snowflake

### LINKED ISSUE #148: Project Snowflake
One request that I often see is if we can make a new Fractal instance very easily that is *explicitly* temporary. As MongoDB startup is lightning fast (0.06 seconds) and we can use temporary directories this is fairly easy to do:

```
tmpdir = tempfile.TemporaryDirectory()
mongo_args = ["mongod", "--dbpath", tmpdir.name, "--port", str(port)]

with qcfractal.testing.popen(mongo_args):
    time.sleep(1)
    client = pymongo.MongoClient("localhost:" + str(port))
```

I think we would have a signature like the following:
```
client = qcfractal.Snowflake(dbpath=None)
```

which would have the `interface.FractalClient` attributes plus a few things like `await_results` so that a user can let a result complete. A typical use case would to use this in a Jupyter notebook to explore a bit before running a real instance.

Thoughts?

dgasmith:
## Description
Adds in project snowflake, or a server instance that is temporary and does not required a database instance to be running. This mostly involved lots of little changes making sure all of the Asyncio calls cascaded correctly. To this end several other changes were made:
- [x] Snowflake is now implemented and allows users to spin up instances within python scripts. Closes #148.
- [x] `qcfractal-manager` no longer relies on Tornado/Asyncio and instead uses the `scheduler` module. This greatly simplifies the code where Asyncio was not needed.
- [x] Testing now relies on Snowflake where applicable, we may want to consider allowing `pytest` to use two threads to reduce overall time.
- [x] Managers now push back tasks on shutdown, a chance to return a very small number of jobs.

A quick example:
```py
server = FractalSnowflake()

client = ptl.FractalClient(server)
mol = ptl.Molecule(symbols=["H", "H"], geometry=[0, 0, 0, 0, 5, 0], connectivity=[(0, 1, 1)])
ret = client.add_compute("psi4", "HF", "sto-3g", "energy", None, [mol])

server.await_results()
print(client.query_results(id=ret.ids))
```

As a note MongoDB is particularly nice as the process shuts down if it cannot access the data folder which is cleaned up by Python. The current implementation seems very robust against accidental dangling processes during odd shutdowns.

## Status
- [ ] Changelog updated
- [ ] Ready to go

### DIFF
--- a/examples/test_examples.py
+++ b/examples/test_examples.py
@@ -8,6 +8,7 @@ from qcfractal import testing
 
 _pwd = os.path.abspath(os.path.dirname(__file__))
 
+
 def wait_true(wait_time, func, *args, **kwargs):
 
     periods = wait_time // 4
--- a/qcfractal/__init__.py
+++ b/qcfractal/__init__.py
@@ -9,6 +9,7 @@ from .storage_sockets import storage_socket_factory
 
 # Handle top level object imports
 from .server import FractalServer
+from .snowflake import FractalSnowflake
 from .queue import QueueManager
 
 # Handle versioneer
--- a/qcfractal/cli/qcfractal_manager.py
+++ b/qcfractal/cli/qcfractal_manager.py
@@ -3,13 +3,16 @@ A command line interface to the qcfractal server.
 """
 
 import argparse
+import signal
 from enum import Enum
+from functools import partial
 from typing import List, Optional
 
-from pydantic import BaseSettings, BaseModel, conint, confloat
-import qcfractal
 import tornado.log
+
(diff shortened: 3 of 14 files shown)


Lnaden:
> As a note MongoDB is particularly nice as the process shuts down if it cannot access the data folder

Will SQL do something similar when we switch over to it? Thinking from a future proofing way.

I'll be able to look over this tomorrow morning

dgasmith:
Yes, of course.  We can likely have a fully in memory version as well or even a python native through SQLite if enough features are supported.

PR #220, opened 2019-03-18
Add scratch directory to Manager settings

Lnaden:
CLI manager had an old `cores` instead of `ncores`

## Status
- [ ] Changelog updated
- [x] Ready to go

### DIFF
--- a/qcfractal/cli/qcfractal_manager.py
+++ b/qcfractal/cli/qcfractal_manager.py
@@ -286,7 +286,7 @@ def main(args=None):
         update_frequency=settings.manager.update_frequency,
         cores_per_task=cores_per_task,
         memory_per_task=memory_per_task,
-        scratch_directory=settings.manager.scratch_directory
+        scratch_directory=settings.common.scratch_directory
     )
 
     # Add exit callbacks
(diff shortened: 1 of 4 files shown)


dgasmith:
Did you want to fix scratch_directory in another PR?

Lnaden:
I'll do it in this one. I had just opened this one before we talked about that.

Lnaden:
Updated the PR to include the scratch directory management. Also fixed a small bug in the forbidden dask keywords

qcfractal/cli/qcfractal_manager.py
```diff
@@ -90,7 +91,7 @@ class DaskQueueSettings(BaseSettings):
 
     def __init__(self, **kwargs):
         """Enforce that the keys we are going to set remain untouched"""
-        forbidden_set = {"name", "ncores", "memory", "processes", "walltime", "env_extra", "qca_resource_string"}
+        forbidden_set = {"name", "cores", "memory", "processes", "walltime", "env_extra", "qca_resource_string"}
```
dgasmith: Why this switch? It is `ncores` everywhere else.

Lnaden: Its `cores` here because the forbidden set is for *dask specific keywords*, so the user cannot overload the `SLURMCluster(cores=...)` kwarg through extra keys in the model.

PR #222, opened 2019-03-19
REST Interface Validation

dgasmith:
## Description
Fully documents and validates the REST requests at both the server and client endpoints using BaseModels with `extra = 'forbid'`. In addition: 
- [x] Creates a wrapper around responses to serve the correct classes in a concise way.
- [x] Removes duplicate FractalClient code into a `_automodel_request` wrapper.
- [x] Adds full python docstrings and typing to the FractalClient.
- [x] Homogenizes all queries to take in 
- [x] Additional typing and docstrings for Datasets.

## Status
- [ ] Changelog updated
- [x] Ready to go

### DIFF
--- a/qcfractal/interface/client.py
+++ b/qcfractal/interface/client.py
@@ -5,14 +5,13 @@ import os
 from collections import defaultdict
 from typing import Any, Dict, List, Optional, Union
 
+from pydantic import ValidationError
+
 import requests
 
 from .collections import collection_factory, collections_name_map
 from .models import GridOptimizationInput, Molecule, ObjectId, TorsionDriveInput, build_procedure
-from .models.rest_models import (rest_model, TaskQueueGETBody, TaskQueueGETResponse, TaskQueuePOSTBody,
-                                 TaskQueuePOSTResponse, ComputeResponse)
-
-from .models.rest_models import QueryStr, QueryObjectId, QueryProjection
+from .models.rest_models import ComputeResponse, rest_model, QueryStr, QueryObjectId, QueryProjection
 
 ### Common docs
 
@@ -153,7 +152,12 @@ class FractalClient(object):
         """
         body_model, response_model = rest_model(name, rest)
 
-        payload = body_model(**payload)
+        # Provide a reasonable traceback
+        try:
+            payload = body_model(**payload)
+        except ValidationError as exc:
+            raise TypeError(str(exc))
+
         r = self._request(rest, name, data=payload.json())
         response = response_model.parse_raw(r.text)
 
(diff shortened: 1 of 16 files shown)


qcfractal/interface/client.py
```diff
@@ -130,6 +131,59 @@ def _request(self,
...
+            Optionally do not raise error if
+            Description
+        """
+        body_model, response_model = rest_model(name, rest)
+
+        # Provide a reasonable traceback
+        try:
+            payload = body_model(**payload)
+        except ValidationError as exc:
+            raise TypeError(str(exc))
```
Lnaden: Why not just raise a `ValidationError` here?

qcfractal/interface/client.py
```diff
@@ -200,64 +254,60 @@ def server_information(self) -> Dict[str, str]:
...
+        molecular_formula : QueryStr, optional
             Queries the Molecule ``molecular_formula`` field.
         full_return : bool, optional
-            Optionally returns the full request result including the ``meta`` field.
+            Returns the full server response if True that contains additional metadata.
 
         Returns
         -------
-        Dict[str, 'Molecule']
+        List[Molecule]
```
Lnaden: General thought, do we want to make the returns a Union between `List[Molecule]` and `Dict[str, Any]` because of the `full_return` option? We don't even before this PR, but its a thought. Maybe change the documentation?

qcfractal/interface/collections/dataset.py
```diff
@@ -262,7 +289,21 @@ def add_keywords(self, alias: str, program: str, keyword: 'KeywordSet', default:
         return True
 
     def get_keywords(self, alias: str, program: str) -> 'KeywordSet':
+        """Pulls the keywords alias from the server for inspection.s
```
Lnaden: small typo

qcfractal/interface/models/rest_models.py
```diff
@@ -8,25 +8,80 @@
 from .common_models import KeywordSet, Molecule, ObjectId
 from .gridoptimization import GridOptimizationInput
 from .model_utils import json_encoders
+from .records import ResultRecord
 from .task_models import TaskRecord
 from .torsiondrive import TorsionDriveInput
-from .records import ResultRecord
 
 __all__ = [
```
Lnaden: Since you've refactored this to reduce the number of things needed to be imported, it would be good to also change the `__all__` to be lest complicated now. From the `client` file, the `__all__` should now be 
`[ComputeResponse, rest_model, QueryStr, QueryObjectId, QueryProjection]`, unless you want to also expose all of the `GET`, `PUT`, and `POST` methods still?

qcfractal/interface/client.py
dgasmith: I would like to just give return types for the non full option. Complex otherwise.

qcfractal/interface/client.py
dgasmith: Print out complexity, we can reconsider later.

PR #223, opened 2019-03-20
TorsionDriveDataset

dgasmith:
## Description
Adds the TorsionDriveDataset object's primary compute functions. Delaying creation of analysis functions til we have a body constructed.

Also muddles around a bit with docs, reps, snowflake.

## Status
- [ ] Changelog updated
- [x] Ready to go

### DIFF
--- a/examples/test_examples.py
+++ b/examples/test_examples.py
@@ -20,6 +20,7 @@ def wait_true(wait_time, func, *args, **kwargs):
 
     return False
 
+
 @testing.using_psi4
 @testing.using_unix
 @testing.mark_example
@@ -40,7 +41,6 @@ def test_local_server_example():
             assert testing.run_process(["python", "compute_database.py"], **kwargs)
 
             assert wait_true(40, testing.run_process, ["python", "query_database.py"], **kwargs)
-
 
 
 @testing.using_parsl
--- a/qcfractal/interface/client.py
+++ b/qcfractal/interface/client.py
@@ -254,7 +254,7 @@ class FractalClient(object):
 
 ### KVStore section
 
-    def query_kvstore(self, id: QueryObjectId, full_return: bool=False) -> Dict[str, Any]:
+    def query_kvstore(self, id: 'QueryObjectId', full_return: bool=False) -> Dict[str, Any]:
         """Queries items from the database's KVStore
 
         Parameters
@@ -275,9 +275,9 @@ class FractalClient(object):
 ### Molecule section
 
     def query_molecules(self,
-                        id: QueryObjectId=None,
-                        molecule_hash: QueryStr=None,
-                        molecular_formula: QueryStr=None,
+                        id: 'QueryObjectId'=None,
+                        molecule_hash: 'QueryStr'=None,
+                        molecular_formula: 'QueryStr'=None,
                         full_return: bool=False) -> List[Molecule]:
         """Queries molecules from the database.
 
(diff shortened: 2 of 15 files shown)


examples/test_examples.py
```diff
@@ -20,6 +20,7 @@ def wait_true(wait_time, func, *args, **kwargs):
 
     return False
 
+
```
Lnaden: \<sarcasm\> How *dare* you bring empty lines into this PR. Don't you care about the diff drift!? \</sarcasm\>

qcfractal/interface/client.py
```diff
@@ -466,16 +466,16 @@ def add_collection(self, collection: Dict[str, Any], overwrite: bool=False,
...
-                      task_id: QueryObjectId=None,
-                      program: QueryStr=None,
-                      molecule: QueryObjectId=None,
-                      driver: QueryStr=None,
-                      method: QueryStr=None,
-                      basis: QueryStr=None,
-                      keywords: QueryObjectId=None,
-                      status: QueryStr="COMPLETE",
-                      projection: QueryProjection=None,
+                      id: 'QueryObjectId'=None,
```
Lnaden: Curiosity question, how much time do all these delayed type loads save?

qcfractal/server.py
```diff
@@ -100,9 +100,45 @@ def __init__(
...
+            Provide an IOLoop to use for the server
+        compress_response : bool, optional
+            Automatic compression of responses, turn on unless behind a proxy that
+            provides this capability.
+        security : Optional[str], optional
+            The security options for the server {None, "local"}. The local security
+            option uses the database to cache users.
+        allow_read : bool, optional
+            Allow unregistered to perform GET operations on Molecule/KeywordSets/KVStore/Results/Procedures
+        ssl_options : Optional[Dict[str, str]], optional
```
Lnaden: The `ssl_options` are checked on `is False` down below and in the `test_managers` file, but that is not an allowed type. 

There is the block [https://github.com/MolSSI/QCFractal/blob/f3bd1ae9853362c56b3ef444c3314ad7e69a43c6/qcfractal/server.py#L146-L149](here (L146)) which appears to do nothing now, but checks for the `is False` (and can probably be safely removed), and then the `elif` [https://github.com/MolSSI/QCFractal/blob/f3bd1ae9853362c56b3ef444c3314ad7e69a43c6/qcfractal/server.py#L200-L201](here (L200)) a bit below that which does do something with `ssl_options=False`. It would be good to figure these out now since we are in here tinkering.

qcfractal/server.py
```diff
@@ -100,9 +100,45 @@ def __init__(
...
+        query_limit : int, optional
+            The maximum number of entries a query will return.
+        logfile_prefix : str, optional
+            The logfile to use for logging.
+        queue_socket : BaseAdapter, optional
+            An optional Adapter to provide for server to have limited local compute.
+            Should only be used for testing and interactive sessions.
+        max_active_services : int, optional
+            The maximum number of active Services that can be running at any given time.
+        heartbeat_frequency : int, optional
```
Lnaden: Does this strictly have to be `int`? There are some tests in `test_manager` which pass a float

qcfractal/server.py
```diff
@@ -372,7 +432,7 @@ def get_address(self, endpoint=None):
 
 ## Updates
 
-    def update_services(self):
+    def update_services(self) -> None:
```
Lnaden: This returns an int

qcfractal/server.py
dgasmith: No, but close enough.

qcfractal/server.py
dgasmith: Changed to `Union[bool, Dict[str, str]]`, all of the below logic is correct.

qcfractal/interface/client.py
dgasmith: It is to improve the documentation. The objects were coming up as `QCFractal.interface.model.QueryObjectId` in the documentation which is bad.

PR #226, opened 2019-03-22
Added LSF to the queue-manager

### LINKED ISSUE #225: Dask-jobqueue add LSFCluster
The `qcfractal-manager` interface is missing the LSFCluster option.

Lnaden:
Fixes #225

I thought I had fixed this previously, was in an experimental branch I had while figuring out dask, whoops.

## Status
- [ ] Changelog updated
- [x] Ready to go

### DIFF
--- a/qcfractal/cli/qcfractal_manager.py
+++ b/qcfractal/cli/qcfractal_manager.py
@@ -11,7 +11,7 @@ import tornado.log
 
 import qcengine as qcng
 import qcfractal
-from pydantic import BaseModel, BaseSettings, confloat, conint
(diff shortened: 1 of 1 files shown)


dgasmith:
Let's also changed the wall time to 6 hours and check the case sensitivity of the cluster keyword?

Lnaden:
> the case sensitivity of the cluster keyword

Which keyword exactly? The `ClusterSettings.scheduler` option? Are you wanting to ensure cast to lower case?

dgasmith:
Exactly.

Lnaden:
Got it, 6 hour default wall clock, scheduler name is now cast to lower case so it should work so long as the user passes in a string

PR #228, opened 2019-03-24
Visualization Pass

### LINKED ISSUE #126: OpenFFWorkflow Fragment Provenance Information
Currently the OpenFFWorkflow has a Provenance area for each fragment. However the data structure is not capable of supporting this as per fragment we simply have torsiondrive or optimizations which are looped over without a category.

We should change this so that each fragment is a Pydantic object that several fields including provenance information.

### LINKED ISSUE #208: Dataset DataFrame units
As @lothian pointed out the Dataset Dataframes do not keep units on them. The plan was to use the Pandas array extension as show [here](https://pint.readthedocs.io/en/latest/pint-pandas.html). Which looks great, but it appears they are splitting this out to [here](https://github.com/hgrecco/pint-pandas). Implementing our own [Extension Types](https://pandas.pydata.org/pandas-docs/stable/development/extending.html) straightforward, but is not quite stable yet. A straightforward solution of `dataset.df_units` might be sufficient for now.

dgasmith:
## Description
This is a first pass at Dataset visualization which to plot statistics in Juypter notebooks. We picked the Plotly graph library because it both produce nice interactive graphs and also has the Dash library built on top which will easily allow us to build exploration web apps with the same code.

Bar plot example:
image

Violin plot example:
image (1)

Other items:
 - [x] `Dataset.get_history()` will not automatically perform a variety of queries to pull down all matching data.
- [x] Canonicalizes Dataset internal data frames names with a canonical function.

Random bits:
 - [x] `qcfractal-manager` config top level now long accepts additional arguments.
 - [x] Minor fixes to the server and includes a `start-periodic` flag as an optional arg.
 - [x] Fixes a TorsionDriveDataset query index issues (@ChayaSt).
 - [x] Provides the ability to filter submission in TorsionDriveDataset.
 - [x] Datasets now have a canonical unit, can be show by `Dataset.units`. `Dataset.units="eV"` would change all internal units to electron volts for example, fixes #208.
 - [x] Removes OpenFF provenance information until we have a clear use case, will bring it back then. Fixes #126.

## Status
- [ ] Changelog updated
- [x] Ready to go

### DIFF
--- a/qcfractal/interface/collections/dataset.py
+++ b/qcfractal/interface/collections/dataset.py
@@ -12,7 +12,7 @@ from qcelemental import constants
 from .collection import Collection
 from .collection_utils import register_collection
 from ..statistics import wrap_statistics
-from ..models import ObjectId
+from ..models import ObjectId, Molecule
 from ..visualization import bar_plot
 
 
@@ -64,8 +64,6 @@ class Dataset(Collection):
         """
         super().__init__(name, client=client, **kwargs)
 
-        # Initialize internal data frames
-        self.df = pd.DataFrame(index=self.get_index())
         self._units = self.data.default_units
 
         # If we making a new database we may need new hashes and json objects
@@ -73,6 +71,14 @@ class Dataset(Collection):
         self._new_keywords = {}
         self._new_records = []
         self._updated_state = False
+
+        # Initialize internal data frames and load in contrib
+        self.df = pd.DataFrame(index=self.get_index())
+
+        # Inherited classes need to call this themselves
+        if self.__class__.__name__ == "Dataset":
+            for cv in self.data.contributed_values.keys():
+                self.query(cv, contrib=True)
 
     class DataModel(Collection.DataModel):
 
@@ -172,7 +178,7 @@ class Dataset(Collection):
         df.sort_index(inplace=True)
         return df
 
-    def get_history(self, **search: Dict[str, Optional[str]]) -> 'DataFrame':
+    def _get_history(self, **search: Dict[str, Optional[str]]) -> 'DataFrame':
         """Queries for all history that matches the search
 
         Parameters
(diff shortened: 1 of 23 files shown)


Lnaden:
Nice! I'll be able to get a proper review first thing in the morning. What outstanding things do you have for it?

dgasmith:
Lots of things really, but good enough to get in.

qcfractal/interface/collections/dataset.py
```diff
@@ -381,98 +661,102 @@ def get_contributed_values_column(self, key: str, scale='hartree') -> 'Series':
...
         field : str, optional
             The result field to query on
+        as_array : bool, optional
+            Converts the returned values to NumPy arrays
+        force : bool, optional
+            Forces a requery if data is already present
 
         Returns
         -------
         success : bool
```
Lnaden: This is now a `str`.

qcfractal/interface/collections/torsiondrive_dataset.py
```diff
@@ -80,6 +80,7 @@ def add_specification(self,
         spec = TorsionDriveSpecification(
             name=lname, optimization_spec=optimization_spec, qc_spec=qc_spec, description=description)
         self.data.td_specs[lname] = spec
+        self.save()
```
Lnaden: Why does the TDDaset's `add_` method here call a `save` back to the Fractal server? This seems to be different behavior from the other `Collection` classes which don't invoke `save` until its manually called or `compute` is called?

qcfractal/interface/collections/torsiondrive_dataset.py
```diff
@@ -141,14 +142,18 @@ def add_entry(self,
             raise KeyError(f"Record {name} already in the dataset.")
 
         self.data.records[lname] = record
+        self.save()
```
Lnaden: Same question about `save` here in a `get` method

qcfractal/interface/collections/torsiondrive_dataset.py
dgasmith: Yes, Procedure Datasets track their data by ObjectId as our querying is less exact for procedures. Result Datasets only need the molecule ObjectId and it can find the exact result it needs.

qcfractal/interface/collections/torsiondrive_dataset.py
dgasmith: This is an `add_entry` we must call save here to update the record. Slower, but safer version of Dataset `add_entry`, we do not expect tens of thousands of entries here.

dgasmith:
Everything is fine here except for the minor return type issue. Please approve and merge, I will fix the string issue in another PR.

PR #242, opened 2019-04-18
Add parsl compute manager to CLI

Lnaden:
After a bunch of double checking all the variables and init calls, I think this should be the first passable implementation of the Parsl queue manager to the CLI command. I think its just the SLURM 

No testing done yet, but at least I think I got it down to minimal amounts of if/elif in the main block.

## Description
Provide a brief description of the PR's purpose here.

## Todos
  - [ ] Test on our cluster
  - [ ] Handle non-standard ethernet port communication between worker and executor, see if this is even an issue

## Questions
- [x] HighThroughputExecutor or ExtremeScalingExecutor?
- [x] How to handle memory assignment, as I don't think there is a way to do so, leave it to QCEngine?
- [x] What to do about missing LSF and MOAB managers as Parsl does not have them right now.

## Status
- [ ] Changelog updated
- [x] Ready to go

### DIFF
--- a/qcfractal/cli/qcfractal_manager.py
+++ b/qcfractal/cli/qcfractal_manager.py
@@ -21,6 +21,8 @@ __all__ = ["main"]
 
 QCA_RESOURCE_STRING = '--resources process=1'
 
+logger = logging.getLogger("qcfractal.cli")
+
 
 class SettingsCommonConfig:
     env_prefix = "QCA_"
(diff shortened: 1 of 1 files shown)


qcfractal/cli/qcfractal_manager.py
```diff
@@ -119,12 +120,61 @@ class Config(SettingsCommonConfig):
...
+        """Enforce that the keys we are going to set remain untouched"""
+        # This set blocks the Parsl Executor and Provider keywords which we set, so the names of the keywords align
+        # to those classes' kwargs, not whatever Fractal chooses to use as keywords.
+        forbidden_set = {"nodes_per_block", "max_blocks", "worker_init", "scheduler_options", "wall_time"}
+        bad_set = set(kwargs.keys()) & forbidden_set
+        if bad_set:
+            raise KeyError("The following items were set as part of parsl's provider, however, "
+                           "there are other config items which control these in more generic "
+                           "settings locations: {}".format(bad_set))
+        super().__init__(**kwargs)
```
dgasmith: We may want to have a class at some point that does this for us if we keep repeating these operations.

Lnaden:
> Overall this LGTM. Good to merge when you are.

Thanks, its not ready yet since I have not hashed out all the debugging and robustness I want to do yet. I'll update the title and post here when I think its ready for merge

Lnaden:
This PR should now be ready to go if tests pass. There are a number of Parsl changes which we are working with the devs to get in, but for now this should be stable enough to merge and make changes on later.

PR #243, opened 2019-04-18
Support new features from torsiondrive API update

yudongqiu:
## Description
Some updates are done to support the new features recently included in `torionsdrive.td_api`

## Todos
  - [x] Updated TDKeywords model in `interface/models/torsiondrive.py`
  - [x] Updated `torsiondrive_service.py` to parse the new arguments, and support "extra constraints"
  - [x] Updated `test_services.py` to test each of the newly added options
  - [x] Updated conda env "openff.yaml" to require `torsiondrive >= 0.9.6`

## Questions
- [ ] Do we need to double check if the "extra constraints" conflict with the torsion scan?

## Status
- [ ] Changelog updated
- [x] Ready to go

### DIFF
--- a/qcfractal/tests/test_services.py
+++ b/qcfractal/tests/test_services.py
@@ -120,8 +120,9 @@ def test_service_torsiondrive_duplicates(torsiondrive_fixture):
     base_run, duplicate_run = procedures
     assert base_run.optimization_history == duplicate_run.optimization_history
 
-def test_service_torsiondrive_optional(torsiondrive_fixture):
-    """"Tests torsiondrive with optional keywords"""
+
+def test_service_torsiondrive_option_dihedral_ranges(torsiondrive_fixture):
+    """"Tests torsiondrive with dihedral_ranges optional keyword """
 
     spin_up_test, client = torsiondrive_fixture
 
@@ -143,6 +144,11 @@ def test_service_torsiondrive_optional(torsiondrive_fixture):
     assert set(final_molecules.keys()) == {(-150,), (-120,), (-90,), (-60,)}
     assert all(hasattr(m, "symbols") for m in final_molecules.values())
 
+
+def test_service_torsiondrive_option_energy_decrease_thresh(torsiondrive_fixture):
+    """"Tests torsiondrive with energy_decrease_thresh optional keyword"""
+
+    spin_up_test, client = torsiondrive_fixture
     ## test optional energy_decrease_thresh feature
     ret = spin_up_test(keywords={"grid_spacing": [90], "energy_decrease_thresh": 3e-5})
     # Get a TorsionDriveORM result and check data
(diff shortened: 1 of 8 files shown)


qcfractal/tests/test_services.py
```diff
@@ -120,6 +120,94 @@ def test_service_torsiondrive_duplicates(torsiondrive_fixture):
...
+    assert pytest.approx(0.0005683235570009067, abs=1e-6) == final_energies[(-150,)]
+    assert pytest.approx(0.0002170694130912583, abs=1e-6) == final_energies[(-120,)]
+    assert pytest.approx(0.0001565537585121726, abs=1e-6) == final_energies[(-90,)]
+    assert pytest.approx(0.0007991274441437338, abs=1e-6) == final_energies[(-60,)]
+    # check final molecules
+    final_molecules = result.get_final_molecules()
+    assert set(final_molecules.keys()) == {(-150,), (-120,), (-90,), (-60,)}
+    assert all(hasattr(m, "symbols") for m in final_molecules.values())
+
+    ## test optional energy_decrease_thresh feature
```
dgasmith: Can you split each of these up into separate tests rather than a single lump?

qcfractal/tests/test_services.py
```diff
@@ -120,6 +120,94 @@ def test_service_torsiondrive_duplicates(torsiondrive_fixture):
...
+    assert set(final_energies.keys()) == {(-150,), (-120,), (-90,), (-60,)}
+    assert pytest.approx(0.0005683235570009067, abs=1e-6) == final_energies[(-150,)]
+    assert pytest.approx(0.0002170694130912583, abs=1e-6) == final_energies[(-120,)]
+    assert pytest.approx(0.0001565537585121726, abs=1e-6) == final_energies[(-90,)]
+    assert pytest.approx(0.0007991274441437338, abs=1e-6) == final_energies[(-60,)]
+    # check final molecules
+    final_molecules = result.get_final_molecules()
+    assert all(hasattr(m, "symbols") for m in final_molecules.values())
+
+    ## test optional "extra_constraints" feature
```
dgasmith: Could we try consolidating these tests? It should be sufficient to ensure that simply every option was touched rather than having a test that tries each optimization individually. As TD tests are some of our most expensive, keeping these down to a reasonable number is much preferable.

qcfractal/tests/test_services.py
yudongqiu: Sure I have separated this part into 3 test functions.

qcfractal/tests/test_services.py
yudongqiu: I agree it's better to have short tests.  The additional tests added here takes about 6 seconds on my workstation. Another way to test here is to only parse these options without running actual jobs, however this will allow bugs in the implementation of these optional features to pass the test. Therefore I went the first and more robust way.

dgasmith: Ok, sounds good. I suspect we hit lots of duplicate computations and the results are simply returned through here.

PR #244, opened 2019-04-22
Make task errors more verbose in the logs

Lnaden:
## Description

This PR has the loggers report more details about tasks which fail,
both on the manager and the server side.

## Status
- [x] Changelog updated
- [x] Ready to go

### DIFF
--- a/qcfractal/queue/handlers.py
+++ b/qcfractal/queue/handlers.py
@@ -253,6 +253,14 @@ class QueueManagerHandler(APIHandler):
         name = self._get_name_from_metadata(body.meta)
         self.storage.manager_update(name, completed=completed, failures=error)
 
+        # Report failures after updating database for speed and data stability
+        if error:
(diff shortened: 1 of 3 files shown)


qcfractal/queue/handlers.py
```diff
@@ -253,6 +253,14 @@ def post(self):
         name = self._get_name_from_metadata(body.meta)
         self.storage.manager_update(name, completed=completed, failures=error)
 
+        # Report failures after updating database for speed and data stability
+        if error:
+            self.logger.info("The following received tasks failed with the errors:")
+            for key, result in body.data.items():
+                if not result['success']:
+                    self.logger.info(f"Job {key} failed: {result['error']['error_type']} - "
+                                     f"Msg: {result['error']['error_message']}")
```
dgasmith: Errors are already printed above.

PR #249, opened 2019-04-30
Add many to many relations and Model services in SQL

doaa-altarawy:
## Description
Add many to many relations and Model services in SQL

## Questions
- [ ] Question1

## Status
- [ ] Changelog updated
- [x] Ready to go

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -82,6 +82,7 @@ before_script:
   - python -V
   - mongod --version
   - conda list
+  - psql --version
   - psql -c "create database qcarchivedb;" -U postgres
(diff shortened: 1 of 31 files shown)


dgasmith:
I miss QCArchiveBot... but the LGTM integration is better.

dgasmith:
I patched up the `name` vs `lname` issues in the collections object. It is a bit odd here as the `lname` column is a database-only column used purely for quick lookup.

The remaining failure is that pulling a result under project returns an int index rather than string.

Getting close!

dgasmith:
CI should pass, and then we should be ready to merge this!

Issue #191, opened 2019-03-04
Check minimum Portal version


dgasmith:
A `FractalServer` should also pin a minimal version of `FractalClient` that is supported. If the `FractalClient` is too old there should be a nice warning telling them to update and try again.

dgasmith:
Manager should do this as well. Keep everything consistent and up to date.

dgasmith:
Closed by #251.

PR #251, opened 2019-05-02
Add sanity server/client version check

Lnaden:
Simple check to ensure the client is at least equal to the server
from a minor revision level.

Open to debate on if this is the correct check.

## Status
- [ ] Changelog updated
- [ ] Ready to go

### DIFF
--- a/qcfractal/interface/client.py
+++ b/qcfractal/interface/client.py
@@ -84,16 +84,16 @@ class FractalClient(object):
         self.server_name = self.server_info["name"]
 
         def version_list(version):
-            re.sub(r"\+.*$", "", version.strip("v"))
+            version = re.sub(r"\+.*$", "", version.strip("v"))
             return [int(x) for x in version.split(".")]
 
         from . import __version__  # Import here to avoid circular import from __init__
         server_version = version_list(self.server_info["version"])
         client_version = version_list(__version__)
-        assert client_version[:-1] >= server_version[:-1], (f"Sever's' Fractal version of {server_version} appears to "
-                                                            f"be ahead of this Client's Fractal version of "
-                                                            f"{client_version}! Please up your client! "
-                                                            f"Only MAJOR.MINOR version checked.")
+        assert client_version[:2] >= server_version[:2], (f"Server Fractal version of {server_version} appears to "
+                                                          f"be ahead of this Client's Fractal version of "
+                                                          f"{client_version}! Please up your client! "
+                                                          f"Only MAJOR.MINOR version checked.")
 
     def __str__(self) -> str:
         """A short representation of the current FractalClient.
--- a/qcfractal/interface/client.py
+++ b/qcfractal/interface/client.py
@@ -2,6 +2,7 @@
 
 import json
 import os
+import re
 from collections import defaultdict
 from typing import Any, Dict, List, Optional, Union
 
@@ -81,6 +82,18 @@ class FractalClient(object):
         self.server_info = self._request("get", "information", {}).json()
 
         self.server_name = self.server_info["name"]
+
+        def version_list(version):
+            re.sub(r"\+.*$", "", version.strip("v"))
+            return [int(x) for x in version.split(".")]
+
+        from . import __version__  # Import here to avoid circular import from __init__
+        server_version = version_list(self.server_info["version"])
+        client_version = version_list(__version__)
+        assert client_version[:-1] >= server_version[:-1], (f"Sever's' Fractal version of {server_version} appears to "
+                                                            f"be ahead of this Client's Fractal version of "
+                                                            f"{client_version}! Please up your client! "
+                                                            f"Only MAJOR.MINOR version checked.")
 
     def __str__(self) -> str:
         """A short representation of the current FractalClient.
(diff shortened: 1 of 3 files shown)


qcfractal/interface/client.py
```diff
@@ -79,6 +80,18 @@ def __init__(self,
...
         self.server_name = self.server_info["name"]
 
+        def version_list(version):
+            re.sub(r"\+.*$", "", version.strip("v"))
+            return [int(x) for x in version.split(".")]
+
+        from . import __version__  # Import here to avoid circular import from __init__
+        server_version = version_list(self.server_info["version"])
+        client_version = version_list(__version__)
+        assert client_version[:-1] >= server_version[:-1], (f"Sever's' Fractal version of {server_version} appears to "
```
dgasmith: To be save and explicit I would do `[:2]`. Typos is "Sever".

dgasmith: I think for now major/minor must match. Backwards compute is not guaranteed either.

qcfractal/interface/client.py
```diff
@@ -79,6 +80,18 @@ def __init__(self,
 
         self.server_name = self.server_info["name"]
 
+        def version_list(version):
+            re.sub(r"\+.*$", "", version.strip("v"))
```
dgasmith: Does `re` update in place?

Lnaden: No, typo on my part

qcfractal/interface/client.py
Lnaden: So right now the server is on 0.5.4 but the developmental stuff we are doing is on 0.6.0, if we make them match, we will stop being able to develop or run production at this moment unless we downgrade. 

I also don't think this PR should be merged in until we have a set upgrade path/procedure in place for the server.

dgasmith:
Link to #191.

dgasmith:
For now, let's have a simple solution where the FractalServer has a minimum and maximum Client version it will connect to. When the client opens the connection to a server it will pull this data from `/information` (which will likely never change) and will make sure the client version is within the bracket. 

This allows uncoupling of the client and rightly puts all onus on the server which is much more controllable. The only issue is transitionary periods and testing. We currently need to pin to `0.7.x`, but that will break testing as the version is not out yet. I will think about that solution, might be possible to do fill git revision values as well.

qcfractal/interface/client.py
loriab: better to run `safe_version` on everything?

dgasmith: We control both sides... I think we better convert this to a tuple and ship it that way from the server and be done with it.

Lnaden:
This should be good to go. I'd like to add a test to remind us to update the min/max versions, but Im not sure how to do that since the release is what sets the version numbers.

Lnaden:
I'm not quite sure how to work around the failing tests here. They fail because they are trying to connect to the central, current MolSSI server which doesn't present this information which is now checked for.

dgasmith:
Please rebase the PR, everything should work again.

qcfractal/interface/client.py
```diff
@@ -82,6 +83,27 @@ def __init__(self,
...
+        from . import __version__  # Import here to avoid circular import from __init__
+        try:
+            server_version_min_client = version_list(self.server_info["client_lower_version_limit"])
+            server_version_max_client = version_list(self.server_info["client_upper_version_limit"])
+        except KeyError:
+            raise IOError(f"The Server at {self.address}, version {self.server_info['version']} does not report what "
+                          f"Client versions it accepts! It can be almost asserted your Client is too new for the Server "
+                          f"you are connecting to.")
+        client_version = version_list(__version__)
+        assert server_version_min_client <= client_version <= server_version_max_client, (
```
dgasmith: We should raise an actual error here rather than an assert.

qcfractal/interface/client.py
```diff
@@ -82,6 +83,27 @@ def __init__(self,
...
+                f"This Client of version {client_version} does not fall within the Server's allowed Client versions "
+                f"of [{server_version_min_client}, {server_version_max_client}]. Server address: {self.address}"
```
dgasmith: Perhaps note that it is likely that they need to upgrade their portal version with `pip install qcportal` or `conda install qcportal -c conda-forge.`


These kinds of error messages both here and above should tell a user what to do to fix the issue not just tell them there is an issue.

qcfractal/interface/client.py
```diff
@@ -82,6 +83,27 @@ def __init__(self,
 
         self.server_name = self.server_info["name"]
 
+        def version_list(version):
```
dgasmith: Could we move this above the class with a `_version_list` name? We might want to move this to a util eventually.

qcfractal/interface/client.py
Lnaden: It will raise an actual error if the assert fails, and gives the error string which is the 2nd arg to `assert`

dgasmith: Yes, but why use an assert here over a canonical raise exception? If you truly want an assertion error you can `raise AssertionError`.

PR #253, opened 2019-05-06
GridOpt: Fixes qc_spec hashing

dgasmith:
## Description
Miss spelled a keyword that is not caught outside of hash testing. Added an additional test for this.

## Status
- [ ] Changelog updated
- [x] Ready to go


dgasmith:
No, I am not aware of any production runs at the moment.

PR #255, opened 2019-05-14
Add missing dask and dask-jobqueue dependencies for openff.yaml

jchodera:
## Description
This PR adds two critical dependencies (`dask` and `dask-jobqueue`) to the OpenFF conda devtools environment.

## Status
- [x] Changelog updated
- [x] Ready to go

### DIFF
--- a/devtools/conda-envs/openff.yaml
+++ b/devtools/conda-envs/openff.yaml
@@ -33,3 +33,7 @@ dependencies:
 # QCArchive includes
   - qcengine>=0.6.2
   - qcelemental>=0.3.1
+
+# Distributed computing
+  - dask
+  - dask-jobqueue
--- a/devtools/conda-envs/openff.yaml
+++ b/devtools/conda-envs/openff.yaml
@@ -37,3 +37,6 @@ dependencies:
 # Distributed computing
   - dask
   - dask-jobqueue
+
+# geometric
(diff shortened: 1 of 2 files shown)


dgasmith:
The `devtools/env` is for development packages. I think what we want to do for deployment packages is manage Conda environment channel, see [here](http://docs.anaconda.com/anaconda-cloud/user-guide/tasks/work-with-environments/). The end result will be:
```
conda env create qcarchive/openff-manager
conda activate openff-manager
```

so that downstream folks do not need the repo itself and have a single point which we can update periodically as well. We can get this up and running today if you agree.

I don't think `geometric` needs OpenMM, it's just a small warning there for some reason. Added a geomeTRIC issue [here](https://github.com/leeping/geomeTRIC/issues/92).

jchodera:
That's fine with me, but is at odds with the installation briefing @chayast gave me.

Provided there is a well-documented installation route, I'm happy!

loriab:
> so that downstream folks do not need the repo itself and have a single point which we can update periodically as well. We can get this up and running today if you agree.

if you're doing this, what anaconda.org channel are you pulling from? a molssi or qca channel with no other pkgs?

dgasmith:
A qca channel with explicitly *no* packages.

Issue #245, opened 2019-04-23
Database changes


dgasmith:
Several database changes are going to propagate:
 - Molecule hash changes due to connectivity changes (recompute all molecules hashes, QCElemental)
 - MP2 ResultRecord attribute name changes (rename field, mp2_total_correlation_energy -> mp2_correlation_energy, @sjrl).
 - TorsionDrive hash changes due to extra kwargs (might be fixable with defaults).
 - Torsion/Gird Opt procedure id changes.

Collection breaking changes here so we can systematically fix.

dgasmith:
Fixed in the primary DB.

PR #256, opened 2019-05-14
QCEl: Bumps depends to v0.4

dgasmith:
## Description
Updates API to reflect QCEl changes including the molecular hash differences. This is noted in #245.

## Status
- [ ] Changelog updated
- [x] Ready to go


dgasmith:
Beyond the scope of this PR. Still need to tackle the upgrade plan before 0.7.

PR #259, opened 2019-05-15
Implement get_final_results() for GridOptimizationRecord and TorsionDriveRecord

yudongqiu:
## Description
Current `get_final_energies()` and `get_final_molecules()` does not expose gradients of the last frame, which is useful in force field fitting, so as suggested by @dgasmith, I implemented the `get_final_results()` method, which returns a dictionary with the same key as above methods, and the values are the ResultRecord for the last gradient evaluation of the final geometry optimization.

## Todos
- [x] Implement get_final_results() for GridOptimizationRecord
- [x] Implement get_final_results() for TorsionDriveRecord

## Questions
- [ ] Where is the best place to write tests

## Status
- [ ] Changelog updated
- [x] Ready to go

### DIFF
--- a/qcfractal/interface/models/torsiondrive.py
+++ b/qcfractal/interface/models/torsiondrive.py
@@ -170,7 +170,7 @@ class TorsionDriveRecord(RecordBase):
 
         Returns
         -------
-        energy : Dict[str, Any]
+        energy : Dict[str, float]
             Returns energies at each grid point in a dictionary or at a
             single point if a key is specified.
 
@@ -193,7 +193,7 @@ class TorsionDriveRecord(RecordBase):
 
         Returns
         -------
-        final_molecules : Dict[str, Any]
+        final_molecules : Dict[str, qcelemental.models.molecule.Molecule]
             Returns molecule at each grid point in a dictionary or at a
             single point if a key is specified.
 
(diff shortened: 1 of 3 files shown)


qcfractal/interface/models/torsiondrive.py
```diff
@@ -223,6 +225,61 @@ def get_final_molecules(self, key: Union[int, Tuple[int, ...], str]=None) -> Dic
...
+                final_opt_task = tasks[minpos]
+                if len(final_opt_task.trajectory) > 0:
+                    final_grad_record_id = final_opt_task.trajectory[-1]
+                    # store the id -> grid id mapping
+                    map_id_key[final_grad_record_id] = k
+            # combine the ids into one query
+            query_result_ids = list(map_id_key.keys())
+            # put the ids into batch of 10k limit
+            batch_size_limit = 10000
+            for i_batch in range((len(query_result_ids)+batch_size_limit-1) // batch_size_limit):
```
dgasmith: These will never hit the batch size and batch sizes will be handled differently in the future. My previous batch size comments was mostly as an FYI that this may happen in other circumstances. Can you simplify the code based on the assumption that you will not hit the batch size?

yudongqiu: OK sure.

PR #260, opened 2019-05-16
Envs: Adds production environments to the devtools folder

dgasmith:
## Description
Adds production environment for QCFractal managers and snowflakes. These have also been uploaded to conda so that one can run:
```
conda env create qcarchive/qcf-snowflake
```

and have a fully featured conda environment ready to go. @Lnaden will add these to the managers docs as well. `conda env create qcarchive/qcf-manager-openff` will get a fully featured OpenFF environment ready for example. @jchodera

## Status
- [ ] Changelog updated
- [x] Ready to go

### DIFF
--- a/devtools/prod-envs/README.md
+++ b/devtools/prod-envs/README.md
@@ -1,0 +1,23 @@
+Production Environments
+=======================
+
+This file contains production environments for the QCArchive ecosystem. These environments are for a mix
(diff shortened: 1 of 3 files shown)


dgasmith:
We want base conda to be highest priority due to BLAS issues, rank is top first.

PR #263, opened 2019-05-21
Add SQL DB indexes

doaa-altarawy:
## Description
Add DB indexes to SQLAlchemy.
Results index to be optimized later with a different kind of index if needed
## Status
- [ ] Changelog updated
- [x] Ready to go

### DIFF
--- a/qcfractal/storage_sockets/sql_models.py
+++ b/qcfractal/storage_sockets/sql_models.py
@@ -786,6 +786,7 @@ class TaskQueueORM(Base):
         Index('ix_task_queue_created_on', "created_on"),
         Index('ix_task_queue_keys', "status", "program", "procedure", "tag"),
         Index('ix_task_queue_manager', "manager"),
+        Index('ix_task_queue_base_result_id', "base_result_id")
     )
 
     # meta = {
--- a/qcfractal/storage_sockets/sql_models.py
+++ b/qcfractal/storage_sockets/sql_models.py
@@ -784,7 +784,7 @@ class TaskQueueORM(Base):
 
     __table_args__ = (
         Index('ix_task_queue_created_on', "created_on"),
-        Index('ix_task_queue_status', "status"),
+        Index('ix_task_queue_keys', "status", "program", "procedure", "tag"),
         Index('ix_task_queue_manager', "manager"),
     )
 
@@ -824,7 +824,7 @@ class ServiceQueueORM(Base):
 
     __table_args__ = (
         Index('ix_service_queue_status', "status"),
-        Index('ix_task_queue_status_tag_hash', "status", "tag"),
+        Index('ix_service_queue_status_tag_hash', "status", "tag"),
         Index('ix_service_queue_hash_index', "hash_index")
     )
 
--- a/qcfractal/storage_sockets/sql_models.py
+++ b/qcfractal/storage_sockets/sql_models.py
@@ -2,7 +2,7 @@ import datetime
 # from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy import (Column, Integer, String, DateTime, Boolean,
                         ForeignKey, JSON, Enum, Float, Binary, Table,
-                        inspect, Index)
+                        inspect, Index, UniqueConstraint)
 from sqlalchemy.orm import relationship, object_session, column_property
 from qcfractal.interface.models.records import RecordStatusEnum, DriverEnum
 from qcfractal.interface.models.task_models import TaskStatusEnum, ManagerStatusEnum, PriorityEnum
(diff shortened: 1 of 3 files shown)


qcfractal/storage_sockets/sqlalchemy_socket.py
```diff
@@ -1169,12 +1169,19 @@ def get_procedures(self,
...
         elif procedure == 'torsiondrive':
             className = TorsionDriveProcedureORM
+        elif procedure == 'gridoptimization':
+            className = GridOptimizationProcedureORM
         else:
             # raise TypeError('Unsupported procedure type {}. Id: {}, task_id: {}'
             #                 .format(procedure, id, task_id))
-            # self.logger.warning('Procedure type not specified({}). Query include: Id= {}, task_id= {}'
-            #                 .format(procedure, id, task_id))
+            self.logger.warning('Procedure type not specified({}), so only ID is allowed. '
```
dgasmith: Can we move this warning to below? If the `id` is specified this is a perfectly valid query. In addition, we might not want to print the id/task_id as this list could be very long.

qcfractal/storage_sockets/sql_models.py
```diff
@@ -665,6 +700,10 @@ def optimization_history(self, dict_values):
 
         return dict_values
 
+    __table_args__ = (
+        Index('ix_torsion_drive_program', 'program'),  # todo: needed for procedures?
```
dgasmith: Yes, best to move to procedures.

qcfractal/storage_sockets/sql_models.py
```diff
@@ -367,6 +384,19 @@ class ResultORM(BaseResultORM):
...
+    __table_args__ = (
+        # TODO: optimize indexes
+        # A multicolumn GIN index can be used with query conditions that involve any subset of
+        # the index's columns. Unlike B-tree or GiST, index search effectiveness is the same
+        # regardless of which index column(s) the query conditions use.
+        # Index('ix_result_combined', "program", "driver", "method", "basis",
+        #       "keywords", postgresql_using='gin'),  # gin index
+        # Index('ix_results_molecule', 'molecule'),  # b-tree index
+
+        UniqueConstraint("program", "driver", "method", "basis",
```
dgasmith: We find that we often have lookups that look ups where the full 6 indices are not provided. Does a single multi-index provide good lookup speeds for just `program=rdkit`?

qcfractal/storage_sockets/sql_models.py
```diff
@@ -779,6 +822,12 @@ class ServiceQueueORM(Base):
     # created_on = Column(DateTime, nullable=False)
     # modified_on = Column(DateTime, nullable=False)
 
+    __table_args__ = (
+        Index('ix_service_queue_status', "status"),
+        Index('ix_task_queue_status_tag_hash', "status", "tag"),
```
dgasmith: Should this be `task_queue` here?

qcfractal/storage_sockets/sql_models.py
```diff
@@ -737,6 +776,18 @@ def base_result(self, val):
...
+
+    # An important special case is ORDER BY in combination with LIMIT n: an
+    # explicit sort will have to process all the data to identify the first n
+    # rows, but if there is an index matching the ORDER BY, the first n rows
+    # can be retrieved directly, without scanning the remainder at all.
+
+    __table_args__ = (
+        Index('ix_task_queue_created_on', "created_on"),
+        Index('ix_task_queue_status', "status"),
+        Index('ix_task_queue_manager', "manager"),
```
dgasmith: We look up based on tag, program, procedure, status, priority, and base_result_id. Do we need additional indices here?

qcfractal/storage_sockets/sqlalchemy_socket.py
doaa-altarawy: Ok, done

qcfractal/storage_sockets/sql_models.py
doaa-altarawy: The B-tree index will still be used but not efficiently if the key is on the far right. Therefore,  I marked this as todo, so that we order the 6 keys where the ones on the left are the most used query keys.

qcfractal/storage_sockets/sql_models.py
doaa-altarawy: No, it's a typo, thanks

qcfractal/storage_sockets/sql_models.py
dgasmith: Is this a multi index? We look up by `status` and `base_result_id` separately.

The main multi index is [here](https://github.com/MolSSI/QCFractal/blob/master/qcfractal/storage_sockets/sqlalchemy_socket.py#L1516)

doaa-altarawy: `Status` will work from the multi-column index because it's the left most one. I forgot to add the `base_result_id`, I'll add it.

PR #266, opened 2019-05-22
Changelog for 0.7.0 and Doc fixes

Lnaden:
## Description
Creates the changelog for the 0.7.0 release (still needs final date).
Also updates some docs for setup and installing

## Status
- [x] Changelog updated
- [x] Ready to go

### DIFF
--- a/docs/qcfractal/source/changelog.rst
+++ b/docs/qcfractal/source/changelog.rst
@@ -18,6 +18,13 @@ Changelog
 0.7.0 / 2019-05-??
 ------------------
 
+.. warning:: Final MongoDB Supported Release
+
+    **This is the last major release which support MongoDB.** Fractal is moving towards a PostgreSQL for database to
+    make upgrades more stable and because it is more suited to the nature of QCArchive Data. The upgrade path for
+    MongoDB -> PostgreSQL will be provided by the Fractal developers in the next release. This first one will likely
+    be a bit manual in th form of scripts, but afterwords will be engineered to be more automatic and documented.
+
 New Features
 ++++++++++++
 
@@ -58,7 +65,7 @@ Enhancements
 - (:pr:`248`) Jobs which fail, or cannot be returned correctly, from Queue Managers are now better handled in the
   Manager and don't sit in the Manager's internal buffer. They will attempt to be returned to the Server on later
   updates. If too many jobs become stale, the Manager will shut itself down for safety.
-- (:pr:`258`) Fractal Managers are now fully documented, both from the CLI and through the doc pages themselves.
+- (:pr:`258`) Fractal Queue Managers are now fully documented, both from the CLI and through the doc pages themselves.
 - (:pr:`251`) The Fractal Server now reports valid minimum/maximum allowed client versions. The Portal Client will try
   check these numbers against itself and fail to connect if it is not within the Server's allowed ranges. Clients
   started from Fractal's ``interface`` do not make this check.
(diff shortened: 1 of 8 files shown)


Lnaden:
This should be ready now. I made some last changes:

* Bump the min/max version check of server<-> client to 0.7
* Check only MAJOR.MINOR versions
* Update the release procedure to be specific to Fractal and Portal

I did also check to ensure comparing unequal lengths of `List[int]` on both sides of the equation works fine.

docs/qcfractal/source/install.rst
```diff
@@ -10,12 +10,29 @@ You can update qcfractal using `conda <https://www.anaconda.com/download/>`_::
 
     conda install qcfractal -c conda-forge
 
-This installs qcfractal and the NumPy dependancy.
+This installs qcfractal and the NumPy dependency, but *not* any of the quantum
```
dgasmith: `The above command install qcfractal and its required dependancies, but...`

docs/qcfractal/source/changelog.rst
```diff
@@ -15,15 +15,61 @@ Changelog
 .. Bug Fixes
 .. +++++++++
 
-0.6.1 / 2019-??-??
+0.7.0 / 2019-05-??
 ------------------
```
dgasmith: Can we add a big warning here that this will be the last Mongo compatible release?

docs/qcfractal/source/install.rst
```diff
@@ -10,12 +10,29 @@ You can update qcfractal using `conda <https://www.anaconda.com/download/>`_::
...
+Conda Pre-Created Environments
+++++++++++++++++++++++++++++++
+
+Fractal can also be installed through pre-configured environments you can pull through our Conda Channel::
+
+    conda env create qcarchive/{environment name}
+    conda activate {environment name}
+
+The environments are created from the YAML files hosted on the Anaconda Cloud, which then need to be activated
+to use. You can find all of the environments `here <https://anaconda.org/qcarchive>`_.
```
dgasmith: https://anaconda.org/QCArchive/environments

The above link gives us room to write descriptions for each env.

docs/qcfractal/source/setup_compute.rst
```diff
@@ -27,9 +27,14 @@ libraries while Dask requires ``dask``, ``dask.distributed``, and
 Using the Command Line
 ----------------------
 
+.. note::
+
+    The CLI + YAML config file is the current recommended way to start and run
+    Fractal Queue Managers
```
dgasmith: We sometimes call this by different names. Can we standardize to `FractalManager` or similar? I think somewhere we had tried to do this.

qcfractal/interface/client.py
```diff
@@ -107,15 +107,17 @@ def __init__(self,
                               f"the one of following commands (pip or conda):"
                               f"\n\t- pip install qcportal=={self.server_info['version']}"
                               f"\n\t- conda install -c conda-forge qcportal=={self.server_info['version']}"
+                              f"\n(Only MAJOR.MINOR versions are checked)"
                               )
             client_version = _version_list(__version__)
-            if not server_version_min_client <= client_version <= server_version_max_client:
+            if not server_version_min_client[:2] <= client_version[:2] <= server_version_max_client[:2]:
                 raise IOError(f"This Client of version {client_version} does not fall within the Server's allowed "
                               f"Client versions of [{server_version_min_client}, {server_version_max_client}] at "
```
dgasmith: We need to change these lines to only show major/minor versions.

dgasmith: I guess this is somewhat fixed by setting the parameters in the server. Still maybe do it for the client?

Lnaden:
I was going to make the name changes a separate PR as I'm guessing there will be some back and forth on that.

Lnaden:
I've updated from the comments. This is the banner which is displayed in the change log about Mongo

image

dgasmith:
Attempting to retrigger travis.

PR #267, opened 2019-05-22
Implement pagination in sql with testing

doaa-altarawy:
## Description
Implement pagination in sql with testing

## Status
- [ ] Changelog updated
- [x] Ready to go

### DIFF
--- a/qcfractal/tests/test_storage_sql.py
+++ b/qcfractal/tests/test_storage_sql.py
@@ -12,6 +12,7 @@ from qcfractal.services.services import TorsionDriveService
 from qcfractal.testing import sqlalchemy_socket_fixture as storage_socket
 from qcfractal.interface.models.task_models import TaskStatusEnum
 from datetime import datetime
+from time import time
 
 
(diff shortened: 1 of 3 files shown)


qcfractal/tests/test_storage_sql.py
```diff
@@ -934,16 +934,16 @@ def test_results_pagination(storage_socket):
...
+    #
+    # query (~ 0.03 msec/doc)
+    t1 = time()
 
     ret = storage_socket.get_results(method='M2', status=None, limit=limit, skip=skip)
 
-    # total_time = (time() - t1) * 1000 / first_half
-    # `('Query {} results in {:.2f} msec /doc'.format(first_half, total_time))
+    total_time = (time() - t1) * 1000 / first_half
+    print('Query {} results in {:.2f} msec /doc'.format(first_half, total_time))
```
dgasmith: It looks like the comment was rebased out, but the printing needs to be removed. I would likely be better to shift all of this to the benchmarking folder.

dgasmith:
Thanks!

PR #268, opened 2019-05-22
Manager settings QOL Update: Names

Lnaden:
## Description

Introduces some name changes to some of the Manager options to be
more exact on what they do. The following changes are proposed:

* `ntasks` -> `tasks_per_worker`
* `ncores` -> `cores_per_worker`
* `memory` -> `memory_per_worker`
* `cluster.max_nodes` -> `common.max_workers`
    ~* This one is already in but open for discussion since no release tied to it yet~

This also changes `max_tasks` to be functionally set by default as:

`ceil(cores_per_worker * max_workers * 1.2) + 1`
~

Users can still set this, but are discouraged from doing so.

cc @ChayaSt @vtlim @j-wags as my references for you who have tried to set this up on your own.

## Questions
- Opinions on this (and any setting I didn't touch too!)

## Status
- [ ] Changelog updated
- [x] Ready to go

## Edits:

Updated names based on recent changes

2nd: Final names set, moved the max_nodes in `cluster` to `max_workers` in `common`.

### DIFF
--- a/docs/qcfractal/source/managers_faq.rst
+++ b/docs/qcfractal/source/managers_faq.rst
@@ -16,7 +16,7 @@ Turn on ``verbose`` mode, either add the ``-v`` flag to the CLI, or set the
 ``common.verbose`` to ``True`` in the YAML file. Setting this flag will produce
 much more detailed information. This sets the loggers to ``DEBUG`` level.
 
-In the future, we may allow for different levels of verbosity, but for now there is
+In the future, we may allow for different levels of increased verbosity, but for now there is
 only the one level.
 
 Can I start more than one Manager at a time?
@@ -26,6 +26,8 @@ Yes. This is often done if you would like to create multiple :term:`task<Task>` 
 have different resource requirements or spin up managers that can access
 different resources. Check with your cluster administrators though to find out
 their policy on multiple processes running on the clusters head node.
+
+You can reuse the same config file, just invoke the CLI again.
 
 Can I connect to a Fractal Server besides MolSSI's?
 +++++++++++++++++++++++++++++++++++++++++++++++++++
@@ -139,6 +141,12 @@ This problem is obfuscated by the fact that
 incomplete Conda configuration. **We strongly recommend you do not try setting the absolute Python path** in your
 scripts to get around this, and instead try to ``source`` the Conda ``profile.d`` first.
 
+Other variants:
+
+- "Tests from one program pass, but others don't"
+- "I get errors about unable to find program, but its installed"
+- "I get path and/or import errors when testing"
+
 
 My jobs appear to be running, but only one (or few) workers are starting
 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
@@ -148,5 +156,7 @@ a few things may be happening.
 
 - If jobs are completing very fast, the :term:`Adapter` may not feel like it needs to start more
   :term:`workers<Worker>`, which is fine.
-- Check your ``manger.max_tasks`` arg to pull more :term:`tasks<Task>` from the :term:`Server` to fill
-  the jobs you have started.
+- (Not recommended, use for debug only) Check your ``manger.max_tasks`` arg to pull more :term:`tasks<Task>`
+  from the :term:`Server` to fill the jobs you have started. This option is usually automatically calculated based on
+  your ``common.tasks_per_worker`` and ``cluster.max_cluster_jobs`` to keep all :term:`workers<Worker>` busy and
+  still have a buffer.
--- a/docs/qcfractal/source/managers_faq.rst
+++ b/docs/qcfractal/source/managers_faq.rst
@@ -139,7 +139,16 @@ This problem is obfuscated by the fact that
 :term:`workers<Worker>` such as Dask Workers can still start initially despite being a Python code themselves. Many
 :term:`adapters<Adapter>` will start their programs using the absolute Python binary path which gets around the
 incomplete Conda configuration. **We strongly recommend you do not try setting the absolute Python path** in your
-scripts to get around this, and instead try to ``source`` the Conda ``profile.d`` first.
+scripts to get around this, and instead try to ``source`` the Conda ``profile.d`` first. For example, you might
+need to add something like this to your YAML file (change paths/environment names as needed):
+
+.. code-block:: yaml
+
+    cluster:
+        task_startup_commands:
+            - source ~/miniconda3/etc/profile.d/conda.sh
+            - conda activate qcfractal
+
 
 Other variants:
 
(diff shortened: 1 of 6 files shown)


dgasmith:
I would propose:
```
ntasks -> tasks_per_worker
ncores -> cores_per_worker
memory -> memory_per_worker
max_nodes -> max_concurrent_workers
```

I think job and worker sounds too close, and the official name that most distributed systems pick up is workers.

I would also propose changing `max_tasks` to `max_queued_tasks` or similar.

The default is correctly noted in the code as: `ceil(settings.common.tasks_per_job * cluster_jobs * 1.20) + 1`

Lnaden:
But none of those settings are per worker. The per worker numbers are all `X / tasks_per_job` where X is memory or cores.

dgasmith:
I keep waffling on this, but I would say that is more like worker processes. The scheduler submits workers which are SLURM/PBS jobs, workers spawn on nodes as dictated by their SLURM/PBS description, workers spawn processes which execute the compute.

https://radicalpilot.readthedocs.io/en/latest/overview.html#what-is-a-pilot
https://parsl.readthedocs.io/en/stable/stubs/parsl.executors.HighThroughputExecutor.html

I believe the image in the docs is consistent with the above.

Lnaden:
I agree with the model you laid out, but they lay out multiple worker**s**, where each worker started can handle 1 task at a time, the `memory_` variable is the block of memory the pool of workers draws from, and so the number is what is divided by the number of tasks, meaning the number in the YAML file is *not* on a per-worker basis as its currently implemented.

There is the problem where terms are not consistent over the different adapters which is why I locked them down in the docs and ensured, at least in our writings, they always mean a fixed set of things. There is no release tied to these yet, but I want to make sure we are very set on what each thing is called with as little confusion as possible and a few conditional words as possible

In my opinion: the concept of a "Worker" being the thing which spins up multiple processes that can compute multiple tasks is confusing since that implies 1 worker handles multiple tasks concurrently, and even programs which do this often then refer to the individual process to handle task as a "worker" in shorthand when they mean "worker process". 

I much prefer the term "worker" being a 1-to-1 with process and task and just abstracting away the notion of the higher-level "Worker which spawns multiple processes to do concurrent work" within or language. I think its easier to wrap your head around, even if its not exactly true at each adapter's implementation level.

I'd love to get other people's opinion on this too!

Lnaden:
After speaking with @dgasmith a bit, we're going to present the slack with some definitions and ask for feedback on what people would call them that would make intuitive sense and be the most clear.

Lnaden:
I have updated all the names to the `_per_worker` name and double checked all the docs and settings to ensure our terminology is consistent everywhere. A `job` always refers to the submitted Cluster Job now

Lnaden:
I left `max_cluster_jobs` instead of `max_concurrent_workers` because we did discuss some cases where they are not 1-to-1, although we don't implement any of those cases. Still open for discussion though.

This aside, I think this is ready for review

docs/qcfractal/source/managers_faq.rst
```diff
@@ -139,6 +141,12 @@ This problem is obfuscated by the fact that
 incomplete Conda configuration. **We strongly recommend you do not try setting the absolute Python path** in your
 scripts to get around this, and instead try to ``source`` the Conda ``profile.d`` first.
```
dgasmith: Maybe give an explicit example of this?

```
source ~/anaconda3/etc/profile.d/conda.sh
```

docs/qcfractal/source/managers_samples.rst
```diff
@@ -117,17 +118,20 @@ Sun Grid Engine (SGE) cluster, start a conda environment, and load a module.
 An important note about this one, we have now set ``max_cluster_jobs`` to something larger than 1.
 Each :term:`Job` will still request 4 cores and 256 GB of memory to be evenly distributed between the
 4 :term:`tasks<Task>`, however, the :term:`Adapter` will **attempt to start 5 independent** :term:`jobs<Job>`, for a
-total of 80 cores, 1.280 TB of memory, distributed over 20 :term:`workers<Worker>`. If the :term:`Scheduler` does not
+total of 80 cores, 1.280 TB of memory, distributed over 5 :term:`workers<Worker>` collectively running 20
```
dgasmith: "running 20 concurrent tasks"

qcfractal/cli/qcfractal_manager.py
```diff
@@ -52,27 +52,27 @@ class CommonManagerSettings(BaseSettings):
...
+                    "pool adapter, and/or if cluster.max_cluster_jobs=1, this is the number of concurrent jobs."
     )
-    ncores: int = Schema(
+    cores_per_worker: int = Schema(
         qcng.config.get_global("ncores"),
-        description="Number of cores to be consumed and distributed over the ntasks. These tasks are divided evenly, "
-                    "so it is recommended that quotient of ncores/ntasks be a whole number else the core distribution "
-                    "is left up to the logic of the adapter. The default value is read from the number of detected "
-                    "cores on the system you are executing on.",
+        description="Number of cores to be consumed by the Worker and distributed over the tasks_per_worker. These "
```
dgasmith: We mix worker case (title/lower). Best to pick one. Maybe `*Worker*`?

qcfractal/cli/qcfractal_manager.py
```diff
@@ -132,28 +132,24 @@ class QueueManagerSettings(BaseSettings):
...
-        description="Number of tasks to pull from the Fractal Server to keep locally at all times. "
-                    "As tasks are completed, the local pool is filled back up to this value. These tasks will all "
-                    "attempt to be run at once, but parallel tasks are limited bu number of cluster jobs and ntasks.",
-        gt=0
-    )
     manager_name: str = Schema(
         "unlabeled",
         description="Name of this scheduler to present to the Fractal Server. Descriptive names help the server "
-                    "identify you and debugging purposes."
+                    "identify you and help with debugging."
```
dgasmith: "identify the manager resource and assists with debugging"

qcfractal/cli/qcfractal_manager.py
```diff
@@ -176,6 +172,18 @@ class QueueManagerSettings(BaseSettings):
                     "number times the number of found quantum chemistry programs. Does nothing if `test` is False",
         gt=0
     )
+    max_tasks: Optional[int] = Schema(
```
dgasmith: Did we want to move this to `max_queued_tasks` or similar?

qcfractal/cli/qcfractal_manager.py
```diff
@@ -176,6 +172,18 @@ class QueueManagerSettings(BaseSettings):
...
+        None,
+        description="Generally should not be set. Number of tasks to pull from the Fractal Server to keep locally at "
+                    "all times. If `None`, this is automatically computed as "
+                    "`ciel(common.tasks_per_worker*cluster.max_cluster_jobs*1.2) + 1`. As tasks are completed, the "
```
dgasmith: ciel -> ceil.

dgasmith:
`max_concurrent_workers` might always be a good target since it seems to be the 90% case and doesn't introduce new terminology in the parameters. We can adapt special launchers like Balsam to accommodate this.

Lnaden:
This should be RTG!

PR #269, opened 2019-05-22
Jupyter Server Interactive Sessions

dgasmith:
## Description
To make it easier to get started with QCFractal we have had many requests for the ability to provide interactive sessions so that we can build tutorials around this. Originally I believed that we could have built an async client that would use FractalSnowflake and build synchronous calls from there. After talking with Jupyter a bit with @Carreau (Jupyter core) I gave up this tack and simply switched over to a background process.

Instead of going through snowflake we go through a canonical `qcfractal-server` the reason that we do this is that we are able to control the mongo/postgres temporary directory and guarantee that these processes shut down (though this reminds me that this doesn't work on windows sadly). The subprocess approach does loose some interactivity such as `server.await_...()` kinds of calls, but this is something we could potentially write in by using a client to constant ping and check different task queues.

I think in Py3.8+, future changes to Jupyter (thanks @Carreau once more for working on this!), and/or figuring out exactly how `Dask.distributed` does this we might be able to revert. However, the current implementation is relatively robust and we can move on with things for now.

Accomplish items:

- [x] Attempts to write and destroys a async client.
- [x] Adds a FractalSnowflakeHandler for Jupyter (a mouthful, happy to take name suggestions)
- [x] Tweaks the docs slightly attempting to delineate between a full server, a snowflake, and a snowflake handler. I think this will need more work and unfortunately makes things more complex than desired.
- [x] Tweaks docs here and there.
- [x] Write a short quick start tutorial using the new tech and add under getting started docs.

## Discussion
Any thoughts on the following:
- [ ] Consider hiding `FractalServer` in the python API as this should only be called via Python.
- [ ] Better name than FractalSnowflakeHandler for the Jupyter version of this.
- [ ] Should we replace the current Snowflake with `FractalSnowflakeHandler` tech? The current snowflake is very useful for testing however.

## Status
- [ ] Changelog updated
- [ ] Ready to go

### DIFF
--- a/docs/qcfractal/source/setup_server.rst
+++ b/docs/qcfractal/source/setup_server.rst
@@ -10,12 +10,12 @@ and have access to permanent storage.  This location is often either research
 groups local computers, a supercomputer with  appropriately allocated
 resources for this task, or the cloud.
 
+
 Using the Command Line
 ----------------------
 
-
-A ``qcfractal-server`` instance can be created from the command line:
-
+A ``qcfractal-server`` instance for long-term computation should be created from
+the command line like so:
 
 .. code-block:: console
 
@@ -40,14 +40,6 @@ MongoDB locations can be passed in with the ``--database-uri`` flag. Multiple
         >>> mkdir -p $MONGOPATH
         >>> mongod --dbpath $MONGOPATH
 
-Using the Python API
---------------------
-
-TODO
-
-Server with Compute
--------------------
-
 For small or trial instances of ``qcfractal-server`` an instance can be spun
 up with compute with the following lines:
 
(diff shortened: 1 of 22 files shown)


dgasmith:
@Lnaden Ready for review!

docs/qcfractal/source/setup_server.rst
```diff
@@ -58,3 +50,49 @@ up with compute with the following lines:
...
+
+.. warning::
+
+    All data inside a ``FractalSnowflake`` is temporary and will be deleted when the
+    ``FractalSnowflake`` shuts down.
+
+.. code-block:: python
+
+    >>> from qcfractal import FractalSnowflake
+    >>> from qcfractal.int
```
Lnaden: Not sure what this line is supposed to do?

qcfractal/tests/test_server.py
```diff
@@ -132,3 +133,30 @@ def test_storage_socket(test_server):
...
+
+def test_snowflakehandler_log():
+
+    with FractalSnowflakeHandler() as server:
+        proc = server._qcfractal_proc
+
+        assert "No SSL files passed in" in server.show_log(show=False, nlines=100)
+        assert "0 task" not in server.show_log(show=False, nlines=100)
+
+    assert proc.poll() is not None
```
Lnaden: May as well add the empty line here

docs/qcfractal/source/setup_server.rst
dgasmith: Accidental leftovers.

Lnaden:
I'll merge this when this passes!

PR #270, opened 2019-05-23
LSF Dask Workarounds now tested and working properly

Lnaden:
## Description
The untested Dask + LSF unit fixes had a few bugs in them, this corrects that and makes them more robust to other patterns in the lsf.conf file (i.e. comments)

cc @ChayaSt as this will fix the LSF issues you were seeing.

## Status
- [x] Changelog updated
- [x] Ready to go


Lnaden:
I've also proposed an upstream fix so we can hopefully remove the monkey-patching at some point.

dask/dask-jobqueue#274

Lnaden:
I think this is ready, between @ChayaSt and my's debugging, I think this is working as intended now.

Issue #288, opened 2019-06-13
Adapter options in config files


jeffschriber:
When I run a manager using either Dask or Parsl adapters, I am required to specify some options, i.e.,
```
parsl:
  provider:
    cmd_timeout: 30
```
If no options are present, I see errors like,
`AttributeError: 'NoneType' object has no attribute 'provider'`. Relatedly, I'm not so sure which options I should be using in these blocks.
 
**To Reproduce**
Start any qcfractal-manager with no parsl/dask options specified in the config file

**Expected behavior**
I expect a default parsl/dask object to be created instead of the errors with `Nonetype` that are produced

Lnaden:
The choice of the `common.adapter` setting impacts the other top level blocks you may need for Parsl and/or Dask.

>  expect a default parsl/dask

Part of the problem I had with this block is that instancing this is that the Cluster, Dask, and Parsl blocks tend to be so specific to the exact hardware you are running on that setting defaults of any kind runs the very high risk of setting the wrong defaults which make no sense, and then cause additional confusion for the user who then has an adapter just blindly start consuming resources.

There are a couple of different approaches I can take here, I'd like your opinion on what you would prefer:

1. Default settings, with the known assumption that many of them will likely not apply to your hardware, but at least won't throw a python-level error
2. Much more clear error message (which I admit the one you have isn't helpful) but still require the user to actually set a parsl/dask block?

dgasmith:
I think 1) as #291 completed. There are some simple cases where the extra workflow specific config is not required. 2) would also help where possible :)

Lnaden:
I'll close this as #291 handled most of this issue, and should prevent that particular error from here. As we find more unhelpful errors, we can clean those up in separate issues and PRs

Issue #287, opened 2019-06-13
qcfractal-manager --help throws an error


jchodera:
**Describe the bug**
```
(snowflake) jchodera1@titan-ext2:/lustre/atlas/scratch/jchodera1/chm126/qcfractal> qcfractal-manager --help
Traceback (most recent call last):
  File "/lustre/atlas/scratch/jchodera1/chm126/miniconda/envs/snowflake/bin/qcfractal-manager", line 6, in <module>
    from qcfractal.cli.qcfractal_manager import main
  File "/lustre/atlas/scratch/jchodera1/chm126/miniconda/envs/snowflake/lib/python3.7/site-packages/qcfractal/cli/qcfractal_manager.py", line 419, in <module>
    cli_utils.doc_formatter(ParslQueueSettings)
  File "/lustre/atlas/scratch/jchodera1/chm126/miniconda/envs/snowflake/lib/python3.7/site-packages/qcfractal/cli/cli_utils.py", line 140, in doc_formatter
    prop_type = prop["type"]
KeyError: 'type'
```
**To Reproduce**
```
wget https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh
bash Miniconda3-latest-Linux-x86_64.sh -b -p miniconda
# install miniconda in shell
conda install --yes anaconda-client
conda env create qcarchive/qcf-snowflake -n snowflake
conda activate snowflake
qcfractal-manager --help
```

**Expected behavior**
The command help would be returned.

**Additional context**
<details>
  <summary>Click to expand conda environment info</summary>
<pre>
(snowflake) jchodera1@titan-ext2:/lustre/atlas/scratch/jchodera1/chm126/qcfractal> conda list
# packages in environment at /lustre/atlas/scratch/jchodera1/chm126/miniconda/envs/snowflake:
#
# Name                    Version                   Build  Channel
ambit                     0.3                  h137fa24_1    psi4
asn1crypto                0.24.0                   py37_0  
atomicwrites              1.3.0                    py37_1  
attrs                     19.1.0                   py37_1  
backcall                  0.1.0                    py37_0  
bcrypt                    3.1.6            py37h7b6447c_0  
blas                      1.0                         mkl  
bleach                    3.1.0                    py37_0  
boost                     1.70.0           py37h9de70de_1    conda-forge
boost-cpp                 1.70.0               ha2d47e9_0    conda-forge
bzip2                     1.0.6                h14c3975_5  
[...]

PR #291, opened 2019-06-13
Add real tests for Queue Managers and fix a number of issues

Lnaden:
## Description

In addressing #287 and #288, I figured out how to get some real tests in on the Queue Managers. They can only test that the `adapter` and `QueueManager` constructs are built correctly, so they cannot submit jobs and therefore cannot properly test running on Schedulers like SLURM, PBS, SGE, etc. That said, its still a start and adding the tests revealed a few bugs. Itemized list of changes below:

* Make auto-doc generator work for Pydantic 0.28. Will work with pre 0.28 as well
* Removed the Manager Template Generator 
* Added real tests for the Queue Managers
* Tests revealed a number of corner cases which needed fixing
* Fixed example doc for Manager
* Made a default Dask/Cluster/Parsl set for Queue Managers which don't
  set them (or worse, set them and leave them empty). Potential
  resolution to #288

## Questions
- [x] @jeffschriber this PR does make default constructs for the dask/parsl/cluster blocks and should also avoid unhelpful error from #288. I'm not sure this is quite everything you wanted from that issue (i.e. more clarity on what items need set when in blocks like `parsl`), but in case it is, let me know if we can close that issue. If not, lets leave it open and keep working through that.

## Status
- [x] Changelog updated
- [x] Ready to go


qcfractal/cli/qcfractal_manager.py
```diff
@@ -439,9 +441,9 @@ class ManagerSettings(BaseModel):
     common: CommonManagerSettings = CommonManagerSettings()
     server: FractalServerSettings = FractalServerSettings()
     manager: QueueManagerSettings = QueueManagerSettings()
-    cluster: Optional[ClusterSettings] = None
-    dask: Optional[DaskQueueSettings] = None
-    parsl: Optional[ParslQueueSettings] = None
+    cluster: Optional[ClusterSettings] = ClusterSettings()
+    dask: Optional[DaskQueueSettings] = DaskQueueSettings()
+    parsl: Optional[ParslQueueSettings] = ParslQueueSettings()
```
dgasmith: +1 for this. No reason not to auto init I think.

qcfractal/cli/tests/test_cli.py
```diff
@@ -70,28 +72,54 @@ def test_manager_executor_manager_boot_from_file(active_server, tmp_path):
...
-    if adapter == "parsl" and scheduler == "lsf":
-        pytest.xfail("Parsl has no LSF implementation")
-
-    tmpl_path = tmp_path / "tmp_template.py"
-    args = ["qcfractal-template", adapter, scheduler, "--test", "-o", str(tmpl_path)]
-    testing.run_process(args)
+def cli_manager_runs(config_data, tmp_path):
+    temp_config = tmp_path / "temp_config.yaml"
+    with open(temp_config, 'w') as config:
+        config.write(yaml.dump(config_data))
```
dgasmith: Hmm, might be better to use the `temp_config.write(yaml.dumps())`?

qcfractal/cli/tests/test_cli.py
```diff
@@ -70,28 +72,54 @@ def test_manager_executor_manager_boot_from_file(active_server, tmp_path):
...
+    ("dask", "lSf"),
+    ("parsl", "slurm"),
+    ("parsl", "PBS"),
+    ("parsl", "MoAb"),
+    ("parsl", "SGE"),
+    pytest.param("parsl", "lSf", marks=pytest.mark.xfail),
+    pytest.param("NotAParser", "slurm", marks=pytest.mark.xfail),
+    pytest.param("dask", "NotAScheduler", marks=pytest.mark.xfail),
+])
+def test_cli_managers(adapter, scheduler, tmp_path):
```
dgasmith: How much does this hit the CI?

qcfractal/cli/tests/test_cli.py
```diff
@@ -70,28 +72,54 @@ def test_manager_executor_manager_boot_from_file(active_server, tmp_path):
...
+    if config_present is not None:
+        config[adapter] = None
+
+
+def test_cli_managers_quick_exits():
+    """Test that --help and --schema correctly work"""
+    args = ["qcfractal-manager", "--help"]
+    testing.run_process(args)
+    args = ["qcfractal-manager", "--schema"]
+    testing.run_process(args)
```
dgasmith: Two separate tests perhaps and this should be:
```python
_options = {"coverage": True, "dump_stdout": True}
testing.run_process(args, interupt_after=10, **_options)
```

Same with the above, we should keep all CLI tests somewhat equal.

qcfractal/cli/tests/test_cli.py
Lnaden: Do you mean time wise how much does each of these take to run on the CI?

dgasmith: Yes.

Lnaden: Each one takes anywhere from 3-10 seconds. And much of that is a function of which adapter is being used

dgasmith: Gotcha, seems reasonable.

jeffschriber:
Adding the default inits definitely will solve the issue I was having, thanks! Enabling these minimal config files should also be helpful for general usability (at least by non dask/parsl experts like myself). The docs do touch on some useful options, though I was a bit thrown off by the dask options used in the example YAML files not playing well with the PBS queuing system (e.g. `queue: default`). I think a big part of my confusion was just a general unfamiliarity with dask/parsl which isn't the responsibility of the fractal docs.

Lnaden:
I finalized the last test I meant to do and then reworked them a bit to reduce the time, the whole set of 19 tests takes ~85 seconds now, It was ~155s before the re-work.

I think this is ready to go now

qcfractal/cli/tests/test_cli.py
```diff
@@ -70,28 +71,79 @@ def test_manager_executor_manager_boot_from_file(active_server, tmp_path):
...
+def load_manager_config(adapter, scheduler):
+    config = read_config_file(os.path.join(_pwd, "manager_boot_template.yaml"))
+    config["common"]["adapter"] = adapter
+    config["cluster"]["scheduler"] = scheduler
+    return config
+
+
+@testing.mark_slow
+@testing.using_dask_jobqueue
+@testing.using_parsl
```
dgasmith: One last thing here is that it requires both Parsl and Dask to be present. I think you can do:

```
testing.using_dask_jobqueue("dask", "slurm"),
```

If not can you remove the above decorators and within the test_cli_managers function use a 
```
if (adapter == "dask") and not testing.has_module("Dask"):
    pytest.mark.skip("reasons)
```

Lnaden:
It turns out it was not quite that, but close.

`pytest.param("dask", "slurm", marks=testing.using_dask_jobqueue)`

And to further my love of pytest, we can multi-mark as well, *because of course we can,* apparently!

`pytest.param("parsl", "lSf", marks=[testing.using_parsl, pytest.mark.xfail])`

Issue #232, opened 2019-03-24
Service priority


dgasmith:
Somehow we never added time data to the service queue so the current queue runs them in an unordered fashion. This queue needs the normal `created/modified_on`, `tags`, and `priority` and the service activator needs to respect this ordering just like the task queue.

PR #296, opened 2019-06-21
Services: Now ordered by created data and priority

dgasmith:
## Description
Services are now prioritized by date created to properly order the queue. In addition:
- Adds option to test update frequency and enhances service printing
- Small testing optimizations to reduce total default full test time
- Example tests are being deprecated in favor of snowflake

## Status
- [ ] Changelog updated
- [x] Ready to go


qcfractal/storage_sockets/sql_models.py
```diff
@@ -810,14 +810,22 @@ class ServiceQueueORM(Base):
     procedure_id = Column(Integer, ForeignKey("base_result.id"), unique=True)
     procedure_obj = relationship(BaseResultORM, lazy='joined')
 
+    priority = Column(Integer, default=int(PriorityEnum.NORMAL))
+    created_on = Column(DateTime, default=datetime.datetime.utcnow)
```
doaa-altarawy: We need to write a DB migration to be able to add this, and in indices to the existing production DB.
Please don't merge this pull request until we get the migration script as this will break the production if pushed by mistake.

dgasmith: Isn't this expected that between versions we will continuously make database changes? The primary fractal server will always be updated to a release version.

dgasmith:
Closes #232

dgasmith:
@doaa-altarawy I think I made the upgrade script correctly, please review!

dgasmith:
It looks like our env are not currently resolvable. Not sure if I can solve today, @Lnaden would you have a few minutes to take a peek at this?

Lnaden:
Sure, let me look at this a bit.

Lnaden:
Looks like it was a temporary bug with conda. I re-ran the builds and the dependencies were solvable. Pytest still isn't passing, but it is running tests now.

dgasmith:
@doaa-altarawy Ok, I think this is ready to go if you are happy with the alembic snapshot.

PR #297, opened 2019-06-21
[WIP, DNM] Balsm Adapter

Lnaden:
## Description
This PR adds the Balsam Adapter to the Fractal Queue Managers.

This is still a notable work in progress, but it does run and complete a QueueManager locally in test mode. I can run the Balsam package locally to do the tests, but I have not explored how to set that up yet for a CI environment since I would have to initialize the Database, Start the balsam activate, set the environment variable, start the database, start the service/launcher and then run tests.

I'm going to keep updating the lists in the Todos and Questions as they develop and are addressed.

## Todos
  - [ ] Write formal tests
  - [ ] Update documentation
  - [ ] Add more options for the Balsam Queue Settings

## Questions
- [ ] This tries its best to mimic the existing managers in behavior. 
- [ ] What should the resource request error/warnings be (e.g. Memory has no effect here)
- [ ] How should we handle ranks, threads, and serial/mpi settings?
- [ ] What are the correct defaults to use?

## Status
- [ ] Changelog updated
- [ ] Ready to go


qcfractal/cli/qcabalsam_interface.py
```diff
@@ -0,0 +1,44 @@
...
+"""
+
+import os
+import traceback
+
+from ..fractal_utils import get_function
+from qcelemental.models import FailedOperation
+
+
+def main():
```
dgasmith: Could we write this once to `tmp` during Balsam startup rather than having it as an actual CLI?

Lnaden: I've thought about that, but the path the absolute binary needs to be consistent in order to register it as a Balsam App. Unless we want to re-register a new App for every single job or manager instance.

So my fix for doing that was to have an absolute path from the get go which will be constant between most invocations (at least the same python environment). I thought maybe I could have the user or some configuration point to a file which would be written, but then there is the same App registration problem. As such, my fix was to tie it to the install of QCFractal itself, then the user never has to interact with `ApplicationDescription` object, `BalsamJob` object, or the CLI for balsam

Lnaden:
We are going to keep this as a branch for now, so until further notice, this shall not be merged, but left open for testing, development, and discussion

bennybp:
Certainly outdated now. We have moved to a parsl-only compute manager

Issue #293, opened 2019-06-14
Let Parsl manager control which Launcher it uses


Lnaden:
**Is your feature request related to a problem? Please describe.**
There are some conditions under which the Parsl manager launches with the Aprun launcher instead of the SingleNodeLauncher. The conditions under which this is happens and how to control it should be better explored and configurable. 

**Describe the solution you'd like**
Either correctly automatically choose the Launcher and/or allow the YAML config file to control which one is used.

**Additional context**
Reports are this happens under a PBS environment, but the doc say the default should be SingleNodeLauncher instead

PR #300, opened 2019-06-24
get_future error's return FailedOperation instead of dict

Lnaden:
## Description
Not sure what exactly is causing the errors to be thrown, but
I'm pretty sure this is the code being executed which is causing the
failures.

More debugging from live data may be needed as I have not
spent too long figuring out how to test this.

Several uses have reported this symptom where the `managers.py` file tries to do a `result.success` call, but `dict` does not have such a method. This should trap that at least so further debugging can happen. I suspect this happens when the future crashes (e.g. the Cluster's scheduler does a `kill -9` on the job the worker is in), but I have yet to prove that.

## Todos
  - [ ] Add a way to test this, open to ideas here. Even without testing, we can refine as this starts getting trapped in runs.

## Status
- [ ] Changelog updated
- [x] Ready to go

### DIFF
--- a/qcfractal/cli/qcfractal_manager.py
+++ b/qcfractal/cli/qcfractal_manager.py
@@ -11,7 +11,7 @@ from enum import Enum
 from functools import partial
 from math import ceil
 
-from typing import List, Optional
+from typing import List, Optional, Any, Dict, Union
 
 import tornado.log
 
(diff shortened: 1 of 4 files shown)


Lnaden:
As part of my testing for parsl scale up, I needed to add the fix for #293 on this as well. I think this is ready to go for now, we can further debug the issues from the original part of the PR as they come up.

qcfractal/cli/qcfractal_manager.py
```diff
@@ -370,6 +370,67 @@ class ParslExecutorSettings(SettingsBlocker):
...
+        launcher_map = {
+            "simplelauncher": "SimpleLauncher",
+            "singlenodelauncher": "SingleNodeLauncher",
+            "srunlauncher": "SrunLauncher",
+            "aprunlauncher": "AprunLauncher",
+            "srunmpiLauncher": "SrunMPILauncher",
+            "gnuparallellauncher": "GnuParallelLauncher",
+            "mpiexeclauncher": "MpiExecLauncher"
+        }
+        launcher_string = launcher_map[launcher_lower] if launcher_lower in launcher_map else launcher_base
```
dgasmith: `launcher_string = launcher_map.get(launcher_lower, launcher_base)`?

PR #303, opened 2019-06-25
TorsionDS Status

dgasmith:
## Description
Adds the ability to show TorsionDrive status information from a corresponding dataset.

image

@yudongqiu @ChayaSt 

## Status
- [ ] Changelog updated
- [x] Ready to go

### DIFF
--- a/qcfractal/interface/collections/torsiondrive_dataset.py
+++ b/qcfractal/interface/collections/torsiondrive_dataset.py
@@ -43,7 +43,7 @@ class TorsionDriveDataset(BaseProcedureDataset):
                           name: str,
                           optimization_spec: OptimizationSpecification,
                           qc_spec: QCSpecification,
-                          description: str=None,
+                          description: str = None,
                           overwrite=False) -> None:
         """
         Parameters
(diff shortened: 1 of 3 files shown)


qcfractal/interface/collections/torsiondrive_dataset.py
```diff
@@ -324,5 +326,76 @@ def visualize(self,
...
+            return super().status(specs, collapse=collapse, status=status)
+
+        if status not in [None, 'INCOMPLETE']:
+            raise KeyError("Detailed status is only available for incomplete procedures.")
+
+        if not (isinstance(specs, str) or len(specs) == 1):
+            raise KeyError("Detailed status is only available for a single specification at a time.")
+
+        mapper = self._get_procedure_ids(specs)
+        reverse_map = {v: k for k, v in mapper.items()}
```
Lnaden: Can this ever result in a double up? Or is it because each service has to be unique anyways (as it would call the already finished one from the database)?

dgasmith: Your right, this could potentially result in a doubling if someone accidentally had two of the same starting configurations.

dgasmith: I think we might go with it, the only thing it hits is the labels. You still get the current status of all running computations. Fixing it would be difficult.

PR #312, opened 2019-07-03
Add SQL migration script

doaa-altarawy:
## Description
Added the ability to run migration scripts to upgrade and downgrade the SQLAlchemy DB.
As an example, I added the first migration which adds `username` to the `queue_manager` table.

Examples of how to use it: [https://alembic.sqlalchemy.org/en/latest/](https://alembic.sqlalchemy.org/en/latest/)

A quick cheat sheet: `examples/SQLAlchemy_Setup.md`

## Status
- [ ] Changelog updated
- [x] Ready to go


alembic/versions/d28e52d86633_add_username_to_manager.py
```diff
@@ -0,0 +1,28 @@
...
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('queue_manager', sa.Column('username', sa.String(), nullable=True))
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_column('queue_manager', 'username')
+    # ### end Alembic commands ###
```
dgasmith: Can we move this inside the `qcfractal/*` folder so that users can upgrade their database after they have installed the project?

qcfractal/storage_sockets/sqlalchemy_socket.py
```diff
@@ -1817,8 +1864,10 @@ def _copy_managers(self, record_list: Dict):
 
                 if get_count_fast(doc) == 0:
                     doc = QueueManagerORM(**manager)
-                    doc.created_on = dt.fromtimestamp(doc.created_on / 1e3)
-                    doc.modified_on = dt.fromtimestamp(doc.modified_on / 1e3)
+                    if isinstance(doc.created_on, float):
+                        doc.created_on = datetime.datetime.fromtimestamp(doc.created_on / 1e3)
```
dgasmith: It looks like this is a incorrect merge as well. Can you please change the signature to `dt.fromtimestamp` and remove the `import datetime` at the top?

doaa-altarawy: I added the check `isinstance`, but maybe I was using the `datetime` as it was in the file.

devtools/conda-envs/base.yaml
```diff
@@ -18,6 +18,7 @@ dependencies:
   - sqlalchemy>=1.3
   - psycopg2
   - postgresql
+  - alembic
```
dgasmith: This needs to be added to the `generate_envs.py` in this folder and then run so that this change propagates to all envs.

qcfractal/storage_sockets/sqlalchemy_socket.py
dgasmith: Once this and the env issue is fixed, I will go ahead and merge.

Issue #262, opened 2019-05-20
Enhanced manager reporting


dgasmith:
**Is your feature request related to a problem? Please describe.**
Managers should log their current resource consumption and several metrics on their cumulative consumption. This will allow manager maintainers to obtain a quick overview of consumed resources and also investigate potential issues.

**Describe the solution you'd like**
Every several minutes in the `qcfractal-manager` logs print out the following:
 - Approximate number of nodes, cores, and memory that the manager is consuming. This data can be pulled from most distributed managers by querying the number of active workers.
 - Approximate total number of core-hours consumed since manager boot up. A good approximation is total number of current core * update interval. This does not have to be incredibly accurate, but more to give a general guideline.
 - The number of pulled, completed, batched returned tasks, and failed tasks periodically.

This information would also be very useful to push back to the server, but should be a future step.

**Additional context**
Related to #261.

Lnaden:
Should be fixed by #313

PR #313, opened 2019-07-03
Manager Quality of Life statistics

Lnaden:
Added quality of life stats to the Queue Manager reporting,
only really works in parsl for efficiency, but will track totals
as well.

Progress towards (and possible resolution) of #262

Still a work in progress, but I wanted to open this anyways. Getting stats on running jobs turns out to be kind of a large pain.

## Status
- [ ] Changelog updated
- [x] Ready to go

### DIFF
--- a/qcfractal/cli/qcfractal_manager.py
+++ b/qcfractal/cli/qcfractal_manager.py
@@ -814,9 +814,10 @@ def main(args=None):
 
     # Build out the manager itself
     # Compute max tasks
+    max_concurrent_tasks = settings.common.tasks_per_worker * settings.common.max_workers
     if settings.manager.max_queued_tasks is None:
         # Tasks * jobs * buffer + 1
-        max_queued_tasks = ceil(settings.common.tasks_per_worker * settings.common.max_workers * 1.20) + 1
+        max_queued_tasks = ceil(max_concurrent_tasks * 1.20) + 1
     else:
         max_queued_tasks = settings.manager.max_queued_tasks
 
@@ -832,6 +833,9 @@ def main(args=None):
         scratch_directory=settings.common.scratch_directory,
         verbose=settings.common.verbose
     )
+
+    # Set stats correctly since we buffer the max tasks a bit
+    manager.statistics.max_concurrent_tasks = max_concurrent_tasks
 
     # Add exit callbacks
     for cb in exit_callbacks:
--- a/qcfractal/cli/qcfractal_manager.py
+++ b/qcfractal/cli/qcfractal_manager.py
@@ -186,7 +186,7 @@ class QueueManagerSettings(BaseSettings):
         None,
         description="Generally should not be set. Number of tasks to pull from the Fractal Server to keep locally at "
                     "all times. If `None`, this is automatically computed as "
-                    "`ceil(common.tasks_per_worker*common.max_workers*1.2) + 1`. As tasks are completed, the "
+                    "`ceil(common.tasks_per_worker*common.max_workers*2.0) + 1`. As tasks are completed, the "
                     "local pool is filled back up to this value. These tasks will all attempt to be run concurrently, "
                     "but concurrent tasks are limited by number of cluster jobs and tasks per job. Pulling too many of "
                     "these can result in under-utilized managers from other sites and result in less FIFO returns. As "
@@ -817,7 +817,7 @@ def main(args=None):
     max_concurrent_tasks = settings.common.tasks_per_worker * settings.common.max_workers
     if settings.manager.max_queued_tasks is None:
         # Tasks * jobs * buffer + 1
-        max_queued_tasks = ceil(max_concurrent_tasks * 1.20) + 1
+        max_queued_tasks = ceil(max_concurrent_tasks * 2.00) + 1
     else:
         max_queued_tasks = settings.manager.max_queued_tasks
 
--- a/qcfractal/queue/base_adapter.py
+++ b/qcfractal/queue/base_adapter.py
@@ -199,6 +199,23 @@ class BaseAdapter(abc.ABC):
             True if the closing was successful.
         """
 
+    def count_running(self) -> int:
+        """
+        Similar to count tasks, but relies on adapter-specific implementation to count the currently
+        running tasks. May not be implemented or possible for each adapter, nor is it required for
+        operation. As such, this it is not required to be implemented as an abstract method.
+
+        Returns
+        -------
+        int
+            Number of running tasks
+
+        Raises
+        ------
+        NotImplementedError
+        """
+        raise NotImplementedError("This adapter has not implemented this method yet")
+
     @abc.abstractmethod
     def _submit_task(self, task_spec: Dict[str, Any]) -> Tuple[Hashable, Any]:
         """
--- a/qcfractal/queue/base_adapter.py
+++ b/qcfractal/queue/base_adapter.py
@@ -199,16 +199,16 @@ class BaseAdapter(abc.ABC):
             True if the closing was successful.
         """
 
-    def count_running(self) -> int:
+    def count_running_workers(self) -> int:
         """
-        Similar to count tasks, but relies on adapter-specific implementation to count the currently
-        running tasks. May not be implemented or possible for each adapter, nor is it required for
+        Adapter-specific implementation to count the currently running workers, helpful for resource consumption.
+        May not be implemented or possible for each adapter, nor is it required for
         operation. As such, this it is not required to be implemented as an abstract method.
 
         Returns
         -------
         int
-            Number of running tasks
+            Number of running workers
 
         Raises
         ------
--- a/qcfractal/queue/executor_adapter.py
+++ b/qcfractal/queue/executor_adapter.py
@@ -36,8 +36,8 @@ class ExecutorAdapter(BaseAdapter):
         return task_spec["id"], task
 
     def count_running_workers(self) -> int:
-        # If there are tasks, they are running
-        return bool(self.queue)
+        # This is always "running", even if there are no tasks since its running locally
+        return 1
 
     def acquire_complete(self) -> Dict[str, Any]:
         ret = {}
@@ -83,8 +83,8 @@ class DaskAdapter(ExecutorAdapter):
         return task_spec["id"], task
 
     def count_running_workers(self) -> int:
-        # Have not worked through this yet for Dask
-        raise NotImplementedError
+        # Note: This number may not quite be right if its treating "worker" as a "job" or Dask-Distributed "worker"
+        return self.client._count_active_workers()
 
     def await_results(self) -> bool:
         from dask.distributed import wait
--- a/qcfractal/queue/executor_adapter.py
+++ b/qcfractal/queue/executor_adapter.py
@@ -35,6 +35,10 @@ class ExecutorAdapter(BaseAdapter):
         task = self.client.submit(func, *task_spec["spec"]["args"], **task_spec["spec"]["kwargs"])
         return task_spec["id"], task
 
+    def count_running_workers(self) -> int:
+        # If there are tasks, they are running
+        return bool(self.queue)
+
     def acquire_complete(self) -> Dict[str, Any]:
         ret = {}
         del_keys = []
@@ -77,6 +81,10 @@ class DaskAdapter(ExecutorAdapter):
         task = self.client.submit(
             func, *task_spec["spec"]["args"], **task_spec["spec"]["kwargs"], resources={"process": 1})
         return task_spec["id"], task
+
+    def count_running_workers(self) -> int:
+        # Have not worked through this yet for Dask
+        raise NotImplementedError
 
     def await_results(self) -> bool:
         from dask.distributed import wait
--- a/qcfractal/queue/executor_adapter.py
+++ b/qcfractal/queue/executor_adapter.py
@@ -84,10 +84,11 @@ class DaskAdapter(ExecutorAdapter):
 
     def count_running_workers(self) -> int:
         if hasattr(self.client.cluster, '_count_active_workers'):
-            # Note: This number may not quite be right if its treating "worker" as a "job" or Dask-Distributed "worker"
+            # Note: This should be right since its counting Dask Workers, and each Dask Worker = 1 task, which we then
+            # Multiply by cores_per_task in the manager.
             return self.client.cluster._count_active_workers()
         else:
-            return len(self.client.scheduler_info()['workers'])
+            return len(self.client.cluster.scheduler.workers)
 
     def await_results(self) -> bool:
         from dask.distributed import wait
--- a/qcfractal/queue/executor_adapter.py
+++ b/qcfractal/queue/executor_adapter.py
@@ -83,8 +83,11 @@ class DaskAdapter(ExecutorAdapter):
         return task_spec["id"], task
 
     def count_running_workers(self) -> int:
-        # Note: This number may not quite be right if its treating "worker" as a "job" or Dask-Distributed "worker"
-        return self.client._count_active_workers()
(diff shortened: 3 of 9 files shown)


qcfractal/queue/managers.py
```diff
@@ -428,9 +460,26 @@ def update(self, new_tasks: bool = True, allow_shutdown=True) -> bool:
         self._update_stale_jobs(allow_shutdown=allow_shutdown)
 
         results = self.queue_adapter.acquire_complete()
+
+        # Stats fetching for running tasks, as close to the time we got the jobs as we can
+        last_time = self.statistics.last_update_time
+        now = self.statistics.last_update_time = time.time()
+        time_delta_seconds = now - last_time
+        try:
+            running_tasks = self.queue_adapter.count_running()
```
dgasmith: Instead of running tasks why not count the number of running workers? The number of running workers is often available see `connected_workers` [here](https://parsl.readthedocs.io/en/stable/stubs/parsl.executors.HighThroughputExecutor.html) and [Dask](https://github.com/dask/dask-jobqueue/blob/bff96a329439c23414bd0cb11c0ee8e67145d96b/dask_jobqueue/core.py#L485).

qcfractal/queue/managers.py
```diff
@@ -448,11 +497,24 @@ def update(self, new_tasks: bool = True, allow_shutdown=True) -> bool:
 
             self.active -= n_result
             for key, result in results.items():
+                wall_time_seconds = 0
                 if result.success:
                     n_success += 1
+                    try:
```
dgasmith: Would recommend `if hasattr(result.provenance 'wall_time')`.

qcfractal/queue/managers.py
```diff
@@ -448,11 +497,24 @@ def update(self, new_tasks: bool = True, allow_shutdown=True) -> bool:
...
+                    except:
+                        pass
                 else:
                     error_payload.append(f"Job {key} failed: {result.error.error_type} - "
                                          f"Msg: {result.error.error_message}")
+                    try:
+                        wall_time_seconds = result.input_data['provenance'].get('wall_time')
+                    except:
+                        pass
+                try:
```
dgasmith: Why a try here?

qcfractal/queue/managers.py
```diff
@@ -469,6 +531,23 @@ def update(self, new_tasks: bool = True, allow_shutdown=True) -> bool:
...
+        # Crunch Statistics
+        self.statistics.total_failed_tasks += n_fail
+        self.statistics.total_successful_tasks += n_success
+        self.statistics.total_cpu_hours += task_cpu_hours
+        self.logger.info(f"Stats: Number of processed jobs: {self.statistics.total_completed_tasks}")
+        self.logger.info(f"Stats: {self.statistics.total_successful_tasks} successful / "
+                         f"{self.statistics.total_failed_tasks} failed "
+                         f"({self.statistics.total_successful_tasks/self.statistics.total_completed_tasks*100}% "
+                         f"success rate)")
+        self.logger.info(f"Stats: {self.statistics.total_cpu_hours} CPU Hours logged successfully (estimate)")
```
dgasmith: Can we try to make this a single line?

`(processed=5, failed=1, success=80%, core_hours=53.3)`

qcfractal/queue/managers.py
```diff
@@ -469,6 +531,23 @@ def update(self, new_tasks: bool = True, allow_shutdown=True) -> bool:
...
+        # Handle efficiency calculations
+        if log_efficiency:
+            efficiency = min((max_cpu_hours_running + task_cpu_hours)/max_cpu_hours_possible * 100, 100)
```
dgasmith: Be very careful here, this isn't cpu hours but core hours.

qcfractal/queue/base_adapter.py
```diff
@@ -199,6 +199,23 @@ def close(self) -> bool:
             True if the closing was successful.
         """
 
+    def count_running(self) -> int:
```
dgasmith: If we do leave this, perhaps `count_running_tasks` to be explicit.

qcfractal/queue/managers.py
Lnaden: Because counting the running tasks is a better gauge of how many resources are being consumed. If we only count workers, then we don't take into account that each worker can have multiple tasks running on it. I did try to work it out such that cores cannot be oversubscribed, but I don't know how effective that is.

qcfractal/queue/managers.py
Lnaden: That's a cleaner construct, I'll change that

qcfractal/queue/managers.py
Lnaden: In case for whatever reason the `wall_time` variable does not get filled in with an actual float. I'm trying to preempt getting something like a `None` here, or having a job crash because the cluster job died and something mangled came back. This is just a safety catch which won't bring down the whole manager if something goes very wrong. 

Then again, we may want it to do exactly that to better catch actual bugs. Thoughts?

qcfractal/queue/managers.py
Lnaden: Yes. I like that better

qcfractal/cli/qcfractal_manager.py
```diff
@@ -814,9 +814,10 @@ def main(args=None):
 
     # Build out the manager itself
     # Compute max tasks
+    max_concurrent_tasks = settings.common.tasks_per_worker * settings.common.max_workers
     if settings.manager.max_queued_tasks is None:
         # Tasks * jobs * buffer + 1
-        max_queued_tasks = ceil(settings.common.tasks_per_worker * settings.common.max_workers * 1.20) + 1
+        max_queued_tasks = ceil(max_concurrent_tasks * 1.20) + 1
```
dgasmith: Let us go ahead and bump this to 2x.

qcfractal/queue/executor_adapter.py
```diff
@@ -78,6 +82,10 @@ def _submit_task(self, task_spec: Dict[str, Any]) -> Tuple[Hashable, Any]:
             func, *task_spec["spec"]["args"], **task_spec["spec"]["kwargs"], resources={"process": 1})
         return task_spec["id"], task
 
+    def count_running_workers(self) -> int:
```
dgasmith: We should be able to call [this](https://github.com/dask/dask-jobqueue/blob/master/dask_jobqueue/core.py#L485) function.

qcfractal/queue/managers.py
```diff
@@ -428,9 +469,28 @@ def update(self, new_tasks: bool = True, allow_shutdown=True) -> bool:
...
+        now = self.statistics.last_update_time = time.time()
+        time_delta_seconds = now - last_time
+        try:
+            running_workers = self.queue_adapter.count_running_workers()
+            log_efficiency = True
+        except NotImplementedError:
+            running_workers = 0
+            log_efficiency = False
+        max_core_hours_running = time_delta_seconds * running_workers * self.statistics.cores_per_task / 3600
+        max_core_hours_possible = (time_delta_seconds * self.statistics.max_concurrent_tasks
```
dgasmith: Im not sure `max_core_hours_possible` is too interesting here? This number focuses on your own queue and how active you were. I guess for low queue its possibly nice.

qcfractal/queue/managers.py
```diff
@@ -469,6 +547,33 @@ def update(self, new_tasks: bool = True, allow_shutdown=True) -> bool:
         # Get new tasks
         payload = self._payload_template()
         payload["data"]["limit"] = open_slots
+
+        # Crunch Statistics
+        self.statistics.total_failed_tasks += n_fail
+        self.statistics.total_successful_tasks += n_success
+        self.statistics.total_core_hours += task_cpu_hours
+        try:
```
dgasmith: Can we do this via an `if`?

qcfractal/queue/parsl_adapter.py
```diff
@@ -77,6 +78,29 @@ def _submit_task(self, task_spec: Dict[str, Any]) -> Tuple[Hashable, Any]:
         task = func(*task_spec["spec"]["args"], **task_spec["spec"]["kwargs"])
         return task_spec["id"], task
 
+    def count_running_workers(self) -> int:
+
+        running = 0
+        executor_running_task_map = {key: False for key in self.client.executors.keys()}
```
dgasmith: This sounds like something we could ask of Parsl.

qcfractal/queue/managers.py
Lnaden: I'd like to leave this one in because it gives an upfront estimate of how well we are utilizing the resources you *requested* over the resources you were *allocated*. If, say, someone fires this up and they only get half the nodes they requested, they can look into that. Or, for instance in case of Dask sometimes, users request a bunch of resources they know are available and could be allocated, but then the adapter only spins up a fraction of them, it would be indicative. 

I could also just mask this behind a verbosity level since it could get distracting, especially in the case of drained queues, or remove it outright, but I think there are valid reasons for tracking it at least.

qcfractal/queue/managers.py
Lnaden: Sure, I do this twice so I'll switch it, any particular reason or just preference of `if...else` over `try...except`?

qcfractal/queue/parsl_adapter.py
Lnaden: Possibly, it would really need to be a way to loop over every executor in the Dataflow client and count the workers in a way which makes sense to the executor. Since we only really allow ThreadPool and HighThroughput executors, I just dealt with those for now.

I could probably make this faster and simpler by simply assuming "Is there a ThreadPoolWorker present? Of so it might as well be working since its effectively running locally." Thoughts?

qcfractal/queue/managers.py
dgasmith: It just seems a bit cleaner and more inline with the rest of the code base. I have often found exception catching to be imprecise for these catches.

qcfractal/queue/managers.py
dgasmith: Fair enough, let’s leave it in for now. I think it might be too much standard info, but easier to remove or move to a debug logging than to add back later.

qcfractal/queue/parsl_adapter.py
dgasmith: Right, this is something the parsl team can easily integrate and test without us being in the loop. I think it’s fine for now, but this level of introspection is often easily breakable.

Lnaden:
The most recent commit made a few other changes which I needed to do while getting this sample of the logfile output from below, I'll want a sign off on some of these since they do touch a few things I was not working on at first (cc @dgasmith for the first 3, I can undo them and make a separate PR if you want for them.):

* Fixed a bug in Snowflake I found in an error catch
* Made LGTM ignore the @validator error
* Found a bug in the recursive directory creation of the `qcfractal-server init`
* Made the stats output numbers less long floats
     * On a related note: Nested `f`-string formatting is a thing

Sample of the output (with changes from latest commit):

Uses a Snowflake and the ProcessPoolExecutor to do short tasks on 1 core, 1 worker. Since it is the Pool executor, the potential and used resources will be the same as its always running.

```
[I 190709 16:23:50 managers:288] QueueManager successfully started.
    
[I 190709 16:23:50 managers:536] Pushed 0 complete tasks to the server (0 success / 0 fail).
[I 190709 16:23:50 managers:583] Stats (Tasks): Processed=0, Failed=0, Success=(N/A yet)%, Core Hours (est.)=0, Worker Core Efficiency (est.): 0.00%, Max Resource Core Efficiency (est.): 0.00%
[I 190709 16:23:50 managers:592] Acquired 10 new tasks.
[I 190709 16:24:01 managers:536] Pushed 10 complete tasks to the server (10 success / 0 fail).
[I 190709 16:24:01 managers:583] Stats (Tasks): Processed=10, Failed=0, Success=100.00%, Core Hours (est.)=0.002128599683443705, Worker Core Efficiency (est.): 74.55%, Max Resource Core Efficiency (est.): 74.55%
[I 190709 16:24:01 managers:592] Acquired 10 new tasks.
[I 190709 16:24:11 managers:536] Pushed 10 complete tasks to the server (10 success / 0 fail).
[I 190709 16:24:11 managers:583] Stats (Tasks): Processed=20, Failed=0, Success=100.00%, Core Hours (est.)=0.004118374983469645, Worker Core Efficiency (est.): 72.57%, Max Resource Core Efficiency (est.): 72.57%
[I 190709 16:24:11 managers:592] Acquired 10 new tasks.
[I 190709 16:24:21 managers:536] Pushed 10 complete tasks to the server (10 success / 0 fail).
[I 190709 16:24:21 managers:583] Stats (Tasks): Processed=30, Failed=0, Success=100.00%, Core Hours (est.)=0.006145976119571262, Worker Core Efficiency (est.): 72.40%, Max Resource Core Efficiency (est.): 72.40%
[I 190709 16:24:21 managers:592] Acquired 10 new tasks.
[I 190709 16:24:31 managers:536] Pushed 10 complete tasks to the server (10 success / 0 fail).
[I 190709 16:24:31 managers:583] Stats (Tasks): Processed=40, Failed=0, Success=100.00%, Core Hours (est.)=0.008195060094197591, Worker Core Efficiency (est.): 72.50%, Max Resource Core Efficiency (est.): 72.50%
[I 190709 16:24:31 managers:592] Acquired 10 new tasks.
[I 190709 16:24:41 managers:536] Pushed 10 complete tasks to the server (10 success / 0 fail).
[I 190709 16:24:41 managers:583] Stats (Tasks): Processed=50, Failed=0, Success=100.00%, Core Hours (est.)=0.010241767830318875, Worker Core Efficiency (est.): 72.47%, Max Resource Core Efficiency (est.): 72.47%
[I 190709 16:24:41 managers:592] Acquired 0 new tasks.
```

qcfractal/queue/executor_adapter.py
```diff
@@ -83,8 +83,8 @@ def _submit_task(self, task_spec: Dict[str, Any]) -> Tuple[Hashable, Any]:
         return task_spec["id"], task
 
     def count_running_workers(self) -> int:
-        # Have not worked through this yet for Dask
-        raise NotImplementedError
+        # Note: This number may not quite be right if its treating "worker" as a "job" or Dask-Distributed "worker"
```
dgasmith: `self.client._count_active_workers() * self.client.worker_processes` in that case?

dgasmith:
I think the status line is too long:
```
[I 190709 16:24:41 managers:583] Task stats: Complete=50, Failures=0, Success=100.00%, Pending=30
[I 190709 16:24:41 managers:583] Worker stats (est.): Core-hours=%.2f, Distribution-efficiency=%.2f, Worker-uptime=%.2f
```

For `core-hours` we can use the lovely:

```python
>>> val = 300000012340000.12341
>>> f'{val:,.2f}'
'300,000,012,340,000.12'
```

Lnaden:
The log now looks like this:

```
[I 190710 10:00:49 managers:536] Pushed 10 complete tasks to the server (10 success / 0 fail).
[I 190710 10:00:49 managers:585] Task Stats: Processed=30, Failed=0, Success=100.00%
[I 190710 10:00:49 managers:586] Worker Stats (est.): Core Hours Used=0.01, Core Usage Efficiency: 80.45%
```

In verbose mode, it includes an additional field
```
[I 190710 09:57:21 managers:536] Pushed 10 complete tasks to the server (10 success / 0 fail).
[I 190710 09:57:21 managers:585] Task Stats: Processed=40, Failed=0, Success=100.00%
[I 190710 09:57:21 managers:586] Worker Stats (est.): Core Hours Used=0.01, Core Usage Efficiency: 83.35%, Core Usage vs. Max Resources Requested: 83.35%
```

Because of rounding, the Core Hours used don't report until above 0.01 hours.

PR #314, opened 2019-07-03
PostgresHarness update

### LINKED ISSUE #309: PostgresHarness startup without PostgreSQL
Currently if PostgresHarness attempts to boot without PostgreSQL available in the path a very strange error message occurs. This should explicitly raise an error and 

Likely solution

```python
if shutil.which("pg_ctl") is None:
    raise ValueError("Could not find PostgreSQL. Please `conda install postgresql`.")
```

dgasmith:
## Description
This PR gives some face lifts to the PostgresHarness:
 - Better errors if Postgres is not found, closes #309.
 - Will ignore Postgres if already started and data directory is correct.

## Status
- [ ] Changelog updated
- [x] Ready to go


dgasmith:
@Lnaden please review!

PR #316, opened 2019-07-04
Remove dask-jobqueue workarounds

Lnaden:
This also now adds an implicit minimum version pin to Dask Jobqueue
0.5.0 and Parsl 0.8.0 should users call those.

I also went back and fixed the dask node exclusivity setting for the
other clusters to the best of my ability and the docs I could find.
I improved the docs for that setting to help users should it throw an
error.

## Status
- [ ] Changelog updated
- [x] Ready to go

### DIFF
--- a/devtools/conda-envs/adapters.yaml
+++ b/devtools/conda-envs/adapters.yaml
@@ -28,8 +28,10 @@ dependencies:
   - rdkit
   - dask
   - distributed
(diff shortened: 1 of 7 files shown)


Lnaden:
And I also re-generated the adapter files (correctly this time)

Lnaden:
Travis is now failing, is this a rebase issue?

PR #321, opened 2019-07-15
Add Server CLI Documentation

Lnaden:
## Description
This adds the `qcfractal-server` CLU documentation for all its
sub-options: init, start, config, and upgrade. This included using
the auto-doc wrapper for Pydantic `Schema`-rich classes, which had to
be moved higher up in the directory structure for easy imports. If need
be, I can move it back down into the CLI Utils instead of top level
Utils and import from there.

There are also a number of other fixes found in writing these listed
below:

* Fix Sphinx errors about missing `_static` directory
* Removed doc reference to a `database_design.jpg` file which was removed a while ago
* Renamed arg `--disable_ssl` to `--disable-ssl` to match other args (CLI hyphens vs underscore)
* Fixed bug in Parsl Adapter worker efficiency raising when it should not have
* Small typos in docstring fixes
* Fixed logger bug in PostgresHarness

## Status
- [ ] Changelog updated
- [x] Ready to go

### DIFF
--- a/qcfractal/queue/parsl_adapter.py
+++ b/qcfractal/queue/parsl_adapter.py
@@ -90,14 +90,17 @@ class ParslAdapter(BaseAdapter):
                 # Efficiency loop break
                 break
 
+        found_readable = False
         for executor_key, executor in self.client.executors.items():
             if hasattr(executor, 'connected_workers'):
                 # Should return an int
                 running += executor.connected_workers
+                found_readable = True
             elif hasattr(executor, 'max_threads') and executor_running_task_map[executor_key]:
                 running += 1
-            else:
-                raise NotImplementedError("Cannot accurately estimate consumption from executors")
+                found_readable = True
+        if not found_readable:
+            raise NotImplementedError("Cannot accurately estimate consumption from executors")
 
         return running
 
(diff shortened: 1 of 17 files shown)


qcfractal/queue/parsl_adapter.py
```diff
@@ -90,14 +90,17 @@ def count_running_workers(self) -> int:
...
             if hasattr(executor, 'connected_workers'):
                 # Should return an int
                 running += executor.connected_workers
+                found_readable = True
             elif hasattr(executor, 'max_threads') and executor_running_task_map[executor_key]:
                 running += 1
-            else:
-                raise NotImplementedError("Cannot accurately estimate consumption from executors")
+                found_readable = True
+        if not found_readable:
```
dgasmith: Just to double check, this will not actually throw an error correct and is caught in the manager?

Lnaden: Yes, the only call to that is in the `QueueManager.update` function and it does this:

```python
        try:
            running_workers = self.queue_adapter.count_running_workers()
            log_efficiency = True
        except NotImplementedError:
            running_workers = 0
            log_efficiency = False
```

PR #324, opened 2019-07-15
Optimization molecular trajectory pulling

dgasmith:
Queries a Optimization's molecular trajectory in a single pull and ensures correct ordering. Also ensures that the `get_trajectory()` method pulls in the correct order. The previous version could potentially hit issues if an optimization result was already in the database.

## Status
- [ ] Changelog updated
- [x] Ready to go


dgasmith:
@janash This is the PR I was talking about.

dgasmith:
It cannot be a partial or changed once it exists.

PR #325, opened 2019-07-17
Optimizes collections

dgasmith:
## Description
Adds a few additional optimizations for large scale collections, these are:
- [x] The compute call is now a single function with an abstracted internal function for the formation of the submission.
- [x] Fast skip ability if an entry already exists.
- [x] Tries to remove the name 'record' when it really is an 'entry', several of the times we say 'record' will be complex to remove however and have been left untouched.

## Status
- [ ] Changelog updated
- [x] Ready to go

### DIFF
--- a/qcfractal/interface/collections/collection.py
+++ b/qcfractal/interface/collections/collection.py
@@ -267,6 +267,10 @@ class BaseProcedureDataset(Collection):
         class Config(Collection.DataModel.Config):
             pass
 
+    @abc.abstractmethod
+    def _internal_compute_add(self, spec: Any, entry: Any, tag: str, priority: str) -> ObjectId:
+        pass
+
     def _pre_save_prep(self, client: 'FractalClient') -> None:
         pass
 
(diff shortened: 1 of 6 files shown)


qcfractal/interface/collections/gridoptimization_dataset.py
```diff
@@ -11,7 +11,7 @@
 from .collection_utils import register_collection
 
 
-class GORecord(BaseModel):
+class GOEntry(BaseModel):
```
Lnaden: Why the change from `Record` to `Entry` in all of these?

qcfractal/interface/collections/torsiondrive_dataset.py
```diff
@@ -32,13 +32,22 @@ class TorsionDriveSpecification(BaseModel):
...
-        records: Dict[str, TDRecord] = {}
+        records: Dict[str, TDEntry] = {}
         history: Set[str] = set()
-        specs: Dict[str, TorsionDriveSpecification] = {}
+        specs: Dict[str, TDEntrySpecification] = {}
 
         class Config(BaseProcedureDataset.DataModel.Config):
             pass
 
+    def _internal_compute_add(self, spec: Any, entry: Any, tag: str, priority: str) -> ObjectId:
```
Lnaden: So I have noticed that on this and the `GOEntrySpecification` the `_internal_compute_add` only adds them as services, which I agree is the correct way to compute them, but couldn't they have also been added as large single compute tasks in the past? Do we want to remove that ability?

qcfractal/interface/collections/gridoptimization_dataset.py
dgasmith: A `Record` is specifically the computation record. These "Entries" are the starting inputs for records. We overloaded the term, this is removing that issue.

qcfractal/interface/collections/torsiondrive_dataset.py
dgasmith: Only TD can be run as a procedure and we have yet to try that. In general, I don't think running TD as a single procedure is a good idea as it creates a chance for a failure at the very end consuming *a lot* of CPU time.

qcfractal/interface/collections/collection.py
```diff
@@ -365,24 +370,23 @@ def list_specifications(self, description=True) -> Union[List[str], 'DataFrame']
...
+        """
+        Adds an entry to the records
         """
-        lname = name.lower()
-        if lname in self.data.records:
-            raise KeyError(f"Record {name} already in the dataset.")
 
-        self.data.records[lname] = record
-        self.save()
+        self._check_entry_exists(name)
```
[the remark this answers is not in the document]
dgasmith: We would like to raise I believe.

PR #329, opened 2019-07-19
Add Logging to all API requests

doaa-altarawy:
## Description
Add Logging to all API requests

## Todos
  - [x] Fixing tornado IP issue with Nginx
  - [ ] Test on a real server with non localhost
  - [x] decide on what to save from request.body
  - [x] Add DB migrations

## Status
- [ ] Changelog updated
- [ ] Ready to go


doaa-altarawy:
I added an issue in geoip2 to request adding it to conda-forge. Currently, it's available in activisiongamescience conda channel.

dgasmith:
I don't feel too comfortable adding a 60MB file to Fractal. Can we make this an optional standalone dependance? I don't imagine too many people will want this kind of data if the IP is stored.

doaa-altarawy:
Yes, sure it can be optional.

PR #326, opened 2019-07-18
Molecule: Explicitly sets fragment options

dgasmith:
## Description
Closes #302.

## Status
- [ ] Changelog updated
- [x] Ready to go

PR #330, opened 2019-07-23
Better doc-formatter tech and model auto-doc's

Lnaden:
Refactored the doc-formatter to be more general, even though it wound
up being a bunch of if...elif statements.

This also adds the `models.rst` file which just tries to run anything
from a BaseModel class through the doc formatter.

The formatter is now wrapped around a try...except so if upstream
changes break the rather fragile flow, then it won't bring the
functioning tech down.

I also added a test for the auto-doc string to make sure everything
at least functions.

## Description
Provide a brief description of the PR's purpose here.

## Status
- [ ] Changelog updated
- [ ] Ready to go


Lnaden:
Some of these LGTM fixes may not be required as #326 appears to make LGTM happy already.

qcfractal/interface/collections/reaction_dataset.py
```diff
@@ -11,7 +11,7 @@
 
 from .collection_utils import nCr, register_collection
 from .dataset import Dataset
-from ..dict_utils import replace_dict_keys
+from qcfractal.interface.util import replace_dict_keys
```
dgasmith: Why switch this? Becomes different than all other imports.

qcfractal/interface/models/records.py
```diff
@@ -90,6 +90,10 @@ def __str__(self) -> str:
     def __repr__(self) -> str:
         return f"<{self}>"
 
+    def __eq__(self, other) -> bool:
```
dgasmith: Lets remove all of these and let LGTM complain.

qcfractal/interface/collections/reaction_dataset.py
Lnaden: Because the refactor did it. I'll revert it

qcfractal/interface/models/records.py
Lnaden: Agreed. I'll strip all those out.

Lnaden:
DNM until after 0.8

PR #335, opened 2019-07-26
Dataset: Fixes get_history with force

dgasmith:
## Description
This is a candidate to back port to QCPortal 0.8.1.

## Status
- [ ] Changelog updated
- [x] Ready to go


Lnaden:
> This is a candidate to back port to QCPortal 0.8.1.

"back port" to 0.8.1? Do you mean release a 0.8.1 based on just this, or do you mean back port to something like a 0.7.3? Or something else entirely?

dgasmith:
A fix is back ported to the previous release, in the case making a `0.8.1` patch.

PR #338, opened 2019-07-26
Add manager to base_results

doaa-altarawy:
## Description
Add manager_name to base results. This can be added automatically when tasks are marked complete or errored.

## Status
- [ ] Changelog updated
- [x] Ready to go


qcfractal/alembic/versions/4bb79efa9855_add_queue_manager_id_to_base_results.py
```diff
@@ -0,0 +1,30 @@
...
+revision = '4bb79efa9855'
+down_revision = '26cfd7b0439e'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('base_result', sa.Column('manager_name', sa.String(), nullable=True))
+    op.create_foreign_key(None, 'base_result', 'queue_manager', ['manager_name'], ['name'], ondelete='SET NULL')
```
dgasmith: Just checking did we wanted to skip making this a foreign key for speed as we are less concerned with consistency here.

qcfractal/storage_sockets/sqlalchemy_socket.py
```diff
@@ -1660,11 +1668,16 @@ def queue_mark_complete(self, task_ids: List[str]) -> int:
 
         update_fields = dict(status=TaskStatusEnum.complete, modified_on=dt.utcnow())
         with self.session_scope() as session:
-            # Update the base results
-            session.query(BaseResultORM)\
-                   .filter(BaseResultORM.id == TaskQueueORM.base_result_id)\
-                   .filter(TaskQueueORM.id.in_(task_ids))\
-                   .update(update_fields, synchronize_session=False)
+            for task_id in task_ids:
```
dgasmith: Out of curiosity is this for loop slow? It should just template the values into a single SQL query correct?

doaa-altarawy: Surely it's slower than the original code. I've just found out the SQLAlchemy supports case statement in sql:)
I'll update that.
SQLAlchemy is great!

dgasmith: Awesome :)

qcfractal/alembic/versions/4bb79efa9855_add_queue_manager_id_to_base_results.py
doaa-altarawy: It's a foreign key using a "unique" non primary key. This is valid.

dgasmith: Right, just curious if we wanted to not add another FK as the results have many FKs on them already which will potentially slow down the inserts.

doaa-altarawy: Well, it's a compromise. If you think it's going to be okay missing up with the manager_name, and will be always forgiving about it then it's okay to remove the FK constraint.
If we remove the FK, and if yo delete a manager, then the manager_name will still be there, referncing wrong ID. So, anytime you look for those managers, you put this in mind.
Also it can happen that you remove a manger and add another with the same name or ID.

In short, it's the integrity issue.

I think since we only touch the PK constraint once (when marking results as complete), it won't be a huge time issue. This constraint won't affect reads.
So I suggest to keep it.

dgasmith: Ok, sounds good to me. Just wanted to double check since we chatted about it.

dgasmith:
Looks like the SQLAlchemy CASE is failing. Is there a specific SQLAlchemy version that we need?

PR #340, opened 2019-07-29
On-Demand BaseModel Doc Generation

Lnaden:
## Description
This changes the behavior of the auto-doc formatting of BaseModels
(and their subclasses) to instead be on-demand when `__doc__` is called
by assigning `__doc__` to be a descriptor instead of a string.

In theory, this should make the speed of importing faster, but I'm not quite sure what the best way to profile it is. I could write a cPython timer `with` inside the `__init__` and see what it returns after a bunch of trials, that would reduce the noise of importing everything else, but it may also just not be worth it.

Yes, this does abuse the __doc__ method by making it not technically read only, but this does avoid having to deal with fancy meta-classing.

## Todos
  - [ ] Figure out the correct way to time this process without the noise of importing everything else (optional)

## Questions
  - [ ] Should I make the `auto_gen_docs_on_demand` and `AutoPydanticDocGenerator` fault tolerant?

## Status
- [ ] Changelog updated
- [x] Ready to go

### DIFF
--- a/qcfractal/cli/qcfractal_manager.py
+++ b/qcfractal/cli/qcfractal_manager.py
@@ -17,7 +17,7 @@ import qcfractal
 from pydantic import BaseModel, BaseSettings, validator, Schema
 
 from . import cli_utils
-from ..interface.util import doc_formatter
+from ..interface.util import auto_gen_docs_on_demand
 
 __all__ = ["main"]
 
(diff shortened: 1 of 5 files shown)


dgasmith:
I think you made it fault tolerant already?

Lnaden:
I made the actual `doc_formatter` function fault tolerant previously, but the new `auto_gen_docs_on_demand` and `AutoPydanticDocGenerator` which are called first can still throw errors. They don't depend directly on the Pydantic Models though.

One potential failure is if a user tries to set call `auto_gen_docs_on_demand` on the same class, or the instance of said class which already has its docs auto formatted.

dgasmith:
Those cases should always be caught by CI however? Either way.

Lnaden:
Should be. I'll go ahead and merge and we can fix it if it breaks.

Issue #336, opened 2019-07-26
Initial Queue Manager Stats should be more clear they need initial returns first


Lnaden:
Based on Slack conversation between @dgasmith and @jchodera

Description:
"The initial set of manager stats printed may not be useful because utilization and runtime is zero. So when grepping, there are some random zeros"

Possible solutions:
* Maybe we could have some sort of note that this will steadily increase
* Consider a secondary logging file that just has this information in it

PR #341, opened 2019-07-30
More Queue Manger QoL additions and testing

Lnaden:
## Description

Fixes a few issues assigned to me for Manager Quality of Life improvements. 

* Adds skeleton YAML generation for the Queue Manager YAML Config through `--skeleton PATH` or `--skel PATH` flag.
* A couple of different fixes to the Queue Manager stats strings to both de-duplicate the "N/A" string and be more helpful about stats not generated yet.
* Adds a test to do some inspection of the stats logs (Yay pytest `caplog` fixture, boo pytest `caplog` fixture's handler being set higher than the level we need so I have to temporarily set it manually).
    * This may be more generally helpful as well elsewhere, so we can move this function/fixture around if need be.
* I moved the check for if `new_tasks is False` or no empty slots *below* the stats texts, may or may not have been the right call as stats will be computed on every `update` call now. Thoughts?

## Status
- [ ] Changelog updated
- [x] Ready to go

### DIFF
--- a/qcfractal/cli/qcfractal_manager.py
+++ b/qcfractal/cli/qcfractal_manager.py
@@ -4,9 +4,12 @@ A command line interface to the qcfractal server.
 
 import argparse
 import signal
+import json
 import logging
 from enum import Enum
 from math import ceil
+import os
+import yaml
 
 from typing import List, Optional
 
@@ -569,6 +572,10 @@ def parse_args():
                                                                 "config file and exit. This will always show the "
                                                                 "most up-to-date schema. It will be presented in a "
                                                                 "JSON-like format.")
+    optional.add_argument("--skeleton", "--skel", type=str, help="Create a skeleton/example YAML config file at the "
+                                                                 "specified path. This does not start the manager "
+                                                                 "and instead creates a skeleton based on all the "
+                                                                 "options specified.")
 
     # Move into nested namespace
     args = vars(parser.parse_args())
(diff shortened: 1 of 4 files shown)


qcfractal/cli/qcfractal_manager.py
```diff
@@ -569,6 +572,10 @@ def parse_args():
                                                                 "config file and exit. This will always show the "
                                                                 "most up-to-date schema. It will be presented in a "
                                                                 "JSON-like format.")
+    optional.add_argument("--skeleton", "--skel", type=str, help="Create a skeleton/example YAML config file at the "
```
dgasmith: Can this default to something like `manager_config.yaml`?

qcfractal/queue/managers.py
```diff
@@ -548,49 +548,56 @@ def update(self, new_tasks: bool = True, allow_shutdown=True) -> bool:
...
         # Crunch Statistics
         self.statistics.total_failed_tasks += n_fail
         self.statistics.total_successful_tasks += n_success
         self.statistics.total_core_hours += task_cpu_hours
         na_format = ''
         float_format = ',.2f'
         if self.statistics.total_completed_tasks == 0:
-            success_rate = "(N/A yet)"
-            success_format = na_format
+            task_stats_str = "Task statistics unavailable until first tasks return"
```
dgasmith: Can this simply be `N/A` or `N/A, no tasks complete`?

qcfractal/tests/test_managers.py
```diff
@@ -58,6 +76,37 @@ def test_queue_manager_single_tags(compute_adapter_fixture):
         assert manager["username"] == CLIENT_USERNAME
 
 
+@testing.using_rdkit
+def test_queue_manager_statistics(compute_adapter_fixture, caplog):
```
dgasmith: Can this be marked as slow, doesn't seem a test that we should run every time?

qcfractal/queue/managers.py
Lnaden: I changed the way the logic works here, so if that block executes, only the "Task statistics unavailable..." line shows and the rest never trigger. This was based on #336 where a more helpful initial logger message was requested.

PR #349, opened 2019-08-05
[WIP] Add more Schema to Pydantic models

Lnaden:
## Description
The first commit here should finalize the `interface/models` directory.

Next up is the rest of the Fractal source code.

Continuation of #348 

## Todos
  - [ ] more models to add

## Status
- [ ] Changelog updated
- [x] Ready to go


qcfractal/interface/models/records.py
```diff
@@ -428,8 +428,11 @@ class OptimizationRecord(RecordBase):
         ...,
         description="The ID of the molecule which was passed in as the reference for this Optimization"
     )
-    qc_spec: QCSpecification
-    keywords: Dict[str, Any] = Schema(  # TODO: defined in Base
+    qc_spec: QCSpecification = Schema(
+        ...,
+        description="The specification of the quantum chemistry calculation to run at each point."
+    )
+    keywords: Dict[str, Any] = Schema(
```
dgasmith: A few examples where we need to rewrite the doc strings. We have: "What type of calculation this result is tied to."

Where in `common_models.py` I rewrote them to use declarative voice: "The type of calculation that is being performed (e.g., energy, gradient, Hessian, ...)."

We should try to keep the strings consistent as well, like in `common_models.py` we have specific phrasing for version which we should extend to this file.

For Enum's we could either change our documentation tech to take the Enum doctoring or we could do:

```
    driver: DriverEnum = Schema(
        ...,
        description= DriverEnum.__doc__
    )
```

I do like being explicit about this here, by calling the `__doc__` attribute.

qcfractal/interface/models/rest_models.py
```diff
@@ -1081,35 +1081,62 @@ class QueueManagerGETBody(BaseModel):
...
     class Config(RESTConfig):
         pass
 
 
 class QueueManagerGETResponse(BaseModel):
-    meta: ResponseGETMeta
-    data: List[Dict[str, Any]]
+    meta: ResponseGETMeta = Schema(
+        ...,
+        description="Standard metadata returned for fetch queries."
```
dgasmith: This might be another spot where `description=ResponseGETMeta.__doc__` would be optimal.

qcfractal/interface/models/task_models.py
```diff
@@ -37,36 +47,83 @@ class BaseResultEnum(str, Enum):
 
 
 class PythonComputeSpec(BaseModel):
-    function: str
-    args: List[Any]
-    kwargs: Dict[str, Any]
+    function: str = Schema(
+        ...,
+        description="The module and function name of a Python-callable to call. Of the form 'module.function'"
```
dgasmith: A few docs are missing periods at the end of sentences.

qcfractal/interface/models/task_models.py
```diff
@@ -37,36 +47,83 @@ class BaseResultEnum(str, Enum):
...
-    parser: str
-    status: TaskStatusEnum = "WAITING"
+    id: ObjectId = Schema(
+        None,
+        description="The Database assigned ID of the Task, if it has been assigned yet."
+    )
+
+    spec: PythonComputeSpec = Schema(
+        ...,
+        description="The Python function specification this task will execute or has executed."
```
dgasmith: Another good spot to be more declarative. "The Python specification for this task."

qcfractal/interface/models/task_models.py
```diff
@@ -37,36 +47,83 @@ class BaseResultEnum(str, Enum):
...
+        ...,
+        description="Name of the quantum chemistry program which must be present to execute this task."
+    )
+    procedure: Optional[str] = Schema(
+        None,
+        description="Name of the procedure the compute platform must be able to perform to execute this task."
+    )
+    manager: Optional[str] = Schema(
+        None,
+        description="The name of the Queue Manager which has ownership of this Task"
```
dgasmith: "The QueueManager that evaluated this task.". Ok to use past tense here as this field is only changed once touched by a manager.

qcfractal/interface/models/gridoptimization.py
```diff
@@ -179,7 +178,7 @@ class GridOptimizationRecord(RecordBase):
     # Input data
     initial_molecule: ObjectId = Schema(
         ...,
-        description="Id of the initial molecule in the database."
+        description="ID of the initial molecule in the database."
```
dgasmith: I purposefully went through and made it `Id` to match `ObjectId`, we may want to consider changing.

qcfractal/interface/models/common_models.py
```diff
@@ -39,8 +39,8 @@ def validate(cls, v):
 
 
 class DriverEnum(str, Enum):
-    """The possible driver configurations of a single quantum chemistry
-    computation.
+    """
+    The driver configuration of a single quantum chemistry computation. This is a finite set of choices.
```
dgasmith: Can you use the string that I wrote for this verbatim? "The type of calculation that is being performed (e.g., energy, gradient, Hessian, ...)."

qcfractal/interface/models/records.py
```diff
@@ -51,19 +52,19 @@ class RecordBase(BaseModel, abc.ABC):
     # Base identification
     id: ObjectId = Schema(
         None,
-        description="ID of the object on the database. This is assigned automatically when the object is fetched"
+        description="ID of the object on the database. This is assigned automatically when the object is fetched."
```
dgasmith: "This is assigned automatically by the database."

qcfractal/interface/models/records.py
```diff
@@ -277,29 +278,29 @@ class ResultRecord(RecordBase):
...
     )
     molecule: ObjectId = Schema(
         ...,
-        description="The ID of the molecule in the Database which the calculation was performed on."
+        description="The ID of the molecule in the Database which the result is computed on."
     )
     basis: Optional[str] = Schema(
         None,
-        description="The basis set which the calculation was performed with."
+        description="The basis set which the calculation is performed with."
```
dgasmith: "The basis set of the calculation (e.g., "sto-3g", "cc-pVDZ", ...)". I think this language is in one of the `common_models`.

qcfractal/interface/models/records.py
```diff
@@ -435,17 +436,17 @@ class OptimizationRecord(RecordBase):
...
     # Results
     energies: List[float] = Schema(
         None,
-        description="The list of energies at each step of the optimization, in order"
+        description="The ordered list of energies at each step of the Optimization."
     )
     final_molecule: ObjectId = Schema(
         None,
-        description="The ID of the final, optimized Molecule the Optimization procedure converged to."
+        description="The Database ID of the final, optimized Molecule the Optimization procedure converged to."
```
dgasmith: `ObjectId` again here I think.

qcfractal/interface/models/gridoptimization.py
```diff
@@ -263,28 +264,31 @@ def deserialize_key(self, key:str)->Tuple[int]:
...
+            scan_number = self.deserialize_key(scan_number)
 
         ret = []
-        for n, idx in enumerate(key):
+        for n, idx in enumerate(scan_number):
             ret.append(self.keywords.scans[n].steps[idx])
 
-        return tuple(ret)
+        # Cast to Tuple[List[float]] instead of tuple(ret) -> Tuple[float,...]
+        ret = (ret, )
```
Lnaden: This function would have failed due to undefined values, but I made the return match the call.

qcfractal/interface/models/gridoptimization.py
```diff
@@ -299,11 +303,12 @@ def get_scan_dimensions(self) -> Tuple[List[float]]:
         for scan in self.keywords.scans:
             ret.append(len(scan.steps))
 
-        return tuple(ret)
+        ret = (ret, )
```
Lnaden: I changed this return to match the Typing hint of the call, but now it causes CI to fail due to:

```
FractalServer Service Build and Iterate Error:
Traceback (most recent call last):
  File "/Users/levinaden/github/qcfractal/qcfractal/server.py", line 480, in update_services
    finished = service.iterate()
  File "/Users/levinaden/github/qcfractal/qcfractal/services/gridoptimization_service.py", line 177, in iterate
    new_points_list = expand_ndimensional_grid(self.dimensions, self.seeds, self.complete)
  File "/Users/levinaden/github/qcfractal/qcfractal/services/service_util.py", line 201, in expand_ndimensional_grid
    if new_dim >= dimensions[d]:
TypeError: '>=' not supported between instances of 'int' and 'list'
```

Which happens because the return is now a `Tuple[List[float]]` instead of `Tuple[float, ...]` which is what it was. It appears the correct call should be `Tuple[float, ...]`, but I am not sure if the fix should be revert my change and adjust the return Type hint, or change the way the dimensions are checked outside of here.

dgasmith: Should be `Tuple[float, ...]`. As we do not correctly validate type hints the current return is always the correct solution.

Lnaden: Great, I'll fix that

Lnaden:
Sure, do you want that as part of this PR or a separate one?

dgasmith:
This one would be great.

Lnaden:
I made a few changes to the Auto-doc function to be more forgiving and handle inherited auto-doc'd models, but I think this should be ready. I've attached a zip of the current output of the REST API page if you want to look at it.

[rest-api.zip](https://github.com/MolSSI/QCFractal/files/3478174/rest-api.zip)

PR #351, opened 2019-08-07
Fix some typos

mattwelborn:
## Description
This PR corrects some typos that I encountered while reading the docs. 

## Todos

## Questions
- [x]  docs/qcfractal/source/install.rst has a pip install section that hints at multiple options but provides only one. This is probably an artifact of editing, but I don't know enough to confidently change it. 

## Status
- [ ] Changelog updated
- [ ] Ready to go

### DIFF
--- a/docs/qcfractal/source/install.rst
+++ b/docs/qcfractal/source/install.rst
@@ -42,6 +42,8 @@ dependencies you would like to keep up to date:
 
 *   ``pip install qcfractal``
 
+.. XXX: only one option is listed here. 
+
 Install from Source
 -------------------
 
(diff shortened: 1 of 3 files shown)


docs/qcfractal/source/install.rst
```diff
@@ -42,6 +42,8 @@ dependencies you would like to keep up to date:
 
 *   ``pip install qcfractal``
 
+.. XXX: only one option is listed here.
```
dgasmith: Whoops, this was intended for `pip install qcfractal[extras]`. Can you rephrase to only state the singular option.

PR #354, opened 2019-08-07
Msgpack Overhaul

dgasmith:
## Description
This PR accomplishes the following:
- [x] Introduces ProtoModel as the base to all models.
- [x] Alters the storage socket to store msg pack for speed.
- [x] Changes default FractalClient submission to msgpack based rather than msgpack based.

This will not pass CI until a number of other PR's and releases are in from Engine/Elemental. This does however pass locally (!), somewhat surprised by this.

Best part: Primary SQL msgpack overhaul was 27 lines of [code](https://github.com/MolSSI/QCFractal/commit/445a679a5370896abe736fd37e604f1d0401db2c). A clear win for OOP and adapters that we might want to use as an example. @sjayellis @doaa-altarawy 

## Status
- [ ] Changelog updated
- [x] Ready to go

PR #356, opened 2019-08-08
Redesign collections in database

doaa-altarawy:
## Description
Redesign collections in database

## Todo

- [x] Add alembic migrations
## Questions
What other attributes need to be in the table? is the current formats correct?

## Status
- [ ] Changelog updated
- [x] Ready to go


doaa-altarawy:
I'm getting an error from pydantic about the records of the Dataset.
`pytest qcfractal/tests/test_collections.py::test_dataset_compute_gradient`

qcfractal/storage_sockets/models/collections_models.py
```diff
@@ -0,0 +1,237 @@
...
+from sqlalchemy.dialects.postgresql import array_agg
+from sqlalchemy.dialects import postgresql
+from sqlalchemy.sql.functions import GenericFunction
+from sqlalchemy.ext.hybrid import hybrid_property
+
+
+# class json_agg(GenericFunction):
+#     type = postgresql.JSON
+
+class json_build_object(GenericFunction):
```
dgasmith: Is it common for these classes to be snake case?

qcfractal/storage_sockets/models/collections_models.py
```diff
@@ -0,0 +1,237 @@
...
+
+    contributed_values = Column(JSON)
+
+    provenance = Column(JSON)  # TODO: in extra?
+
+    history_keys = Column(ARRAY(String, as_tuple=True))
+    history = Column(ARRAY(String, as_tuple=True))
+
+
+class DatasetRecordsORM(Base):
```
dgasmith: Just checking, are these concrete inherited tables? Only reason I ask is all collections should share an Id.

qcfractal/storage_sockets/models/collections_models.py
```diff
@@ -0,0 +1,237 @@
...
+    default_units = Column(String, nullable=True)
+    alias_keywords = Column(JSON, nullable=True)
+    default_program  = Column(String, nullable=True)
+
+    contributed_values = Column(JSON)
+
+    provenance = Column(JSON)  # TODO: in extra?
+
+    history_keys = Column(ARRAY(String, as_tuple=True))
+    history = Column(ARRAY(String, as_tuple=True))
```
dgasmith: This may not be correct for history as it is a `List[Tuple[str...]` object. Might need to be JSON as well.

qcfractal/storage_sockets/models/collections_models.py
```diff
@@ -0,0 +1,237 @@
...
+    default_keywords  = Column(JSON, nullable=True)
+
+    default_driver = Column(String, nullable=True)
+    default_units = Column(String, nullable=True)
+    alias_keywords = Column(JSON, nullable=True)
+    default_program  = Column(String, nullable=True)
+
+    contributed_values = Column(JSON)
+
+    provenance = Column(JSON)  # TODO: in extra?
```
dgasmith: Could be in `extra`, but ok to have a column for this.

qcfractal/storage_sockets/models/collections_models.py
doaa-altarawy: It's an unusual case, where there construct a method (json_agg) using classes (GenericFunction).
In the code it's used as a method.

qcfractal/storage_sockets/models/collections_models.py
doaa-altarawy: I have been getting errors from Pydantic when I used JSON. We can have a second look what was wrong. I prefer it to be json to be more flexible with other types of datasets, and also searchable

qcfractal/storage_sockets/models/collections_models.py
doaa-altarawy: No, they are just like procedures, they share ids

qcfractal/storage_sockets/models/collections_models.py
dgasmith: Ok, everything looks good then. Let's try to figure this out Monday, did you want to write migrations before merge?

doaa-altarawy: Yes of course, I marked ready to go by mistake as I should have added alembic migration to the todo.

dgasmith: You can check out #354 to see some of the migration ideas. Still shaking this out, but something.

dgasmith: Still not sure this is correct for `history` as we need a `List[Tuple[str, ...]]` per dataset. This can be JSON just fine. I would also vote slightly for `history_keys` being JSON as well incase this changes in the future. This data is tiny and we do not get much from having it more strictly type here as pydantic validates the data for us.

qcfractal/storage_sockets/models/collections_models.py
```diff
@@ -0,0 +1,236 @@
...
+
+
+class DatasetRecordsORM(Base):
+    """Association table for many to many"""
+
+    __tablename__ = 'dataset_records'
+
+    dataset_id = Column(Integer, ForeignKey('dataset.id', ondelete='cascade'), primary_key=True)
+    #TODO: check the cascase_delete with molecule
+    molecule_id = Column(Integer, ForeignKey('molecule.id', ondelete='cascade'), primary_key=True)
```
dgasmith: On the fence here with cascade delete. I guess if someone makes a bad datasets, scraps it and then wants to build another one there are molecules that would not be useful to keep around. On the flip side if we build a dataset and run computations with it, cascade delete here will be stopped as the results will be linked to this molecule as well.

If a cascade delete fails are there error involved or is it silent in nature?

dgasmith: At the same time this will be different than a reactiondataset where the molecules will not be cleaned due to their BSON structures?

qcfractal/storage_sockets/models/collections_models.py
```diff
@@ -0,0 +1,236 @@
...
+
+    records_obj = relationship(ReactionDatasetRecordsORM,
+                               lazy='selectin',
+                               cascade="all, delete-orphan",
+                               backref="reaction_dataset")
+
+    def update_relations(self, records=None, **kwarg):
+
+        self.records_obj = []
+        records = [] if not records else records
```
dgasmith: `records or []`?

qcfractal/storage_sockets/models/collections_models.py
doaa-altarawy: Ok, I'll change history and history_keys to JSON.
This cascade IN FOREIGN KEYS means: if my parent is deleted (Dataset or Molecule) then delete this DatasetRecord, not the way around. I'm sorry if this is confusing. 
This might be a rare case that we delete a molecule, but surely we might delete a dataset.
When we delete a dataset, it will delete it's records.

Cascade in RELATIONS means: delete those children when I'm (the parent) is deleted. It's different than cascade in foreign keys.

dgasmith: Just to 100% confirm. If I try to delete a molecule that is in a DatasetRecord the molecule delete will be blocked?

doaa-altarawy: We can control this.
What's implemted now is: when a molecule in the molecule table is deleted, it will be automatically removed from the DatasetRecords.

We can change that behavior to make it though an error and we won't be able to delete a molecule that has a DatasetRecord. (by simply removing the cascade=delete)

dgasmith: Ok gotcha, thats what I thought it said, I think we should remove `cascade=delete` here. So that people do not accidentally mess with their collections when removing molecules.

PR #358, opened 2019-08-09
qcfractal-server user management CLI

mattwelborn:
## Description
This PR exposes a CLI to the user/permissions management functionality in the server. Specifically, it adds three commands:

- `qcfractal-server user add [--password ...] --permissions ... username` which adds a new user
- `qcfractal-server user remove username` which deletes a user.
- `qcfractal-server user show username` which shows permissions for a user.
- `qcfractal-server user modify [--password ... | --reset-password] [--permissions ...] username` which changes a user's permissions and/or password. 

## TODO
- [x] Fix tests broken by changes to add_user function signature.
- [x] Add docs!
- [x] implement password reset and permissions update
- [x] Document the various levels of user permissions.
- [x] Add warning if user is added but authentication is turned off

## Status
- [x] Changelog updated
- [ ] Ready to go


mattwelborn:
@dgasmith Thanks for the review and comments!

> Did we want to add the ability to add or remove permissions from a user? I think this would be an --overwrite flag?

I think it makes sense to have the ability to change a user's permissions. I don't think I want to do it with `add --overwrite` for two reasons. First, I think `add --overwrite` is less intuitive than e.g. `modify`. Second, I don't want to deal with the complicate logic of how `add --overwrite` should interact with `--password`. For example, should `add --overwrite <username> <permissions>` keep the same password or generate a new one? I propose to implement:

`qcfractal-server user modify [--password ...] [--permissions ...] username`

where `--password` (inclusive) or `--permissions` must be specified. 


> It would be good to be able to generate a new password for a user incase the user looses theirs (or compromised!).

I think this could be part of the behavior of `user modify`. e.g. update the above to

`qcfractal-server user modify [--password ...] [--permissions ...] [--reset-password] username`

where now (`--password` xor `--reset-password`) or `--permissions` must be specified.

 
> Will docs go into another PR?

Good catch. Docs will go in this PR.

Will update todo accordingly.

dgasmith:
Sounds good. I believe `--password` vs `--password=mynewpassword` will be similar to the current logic where `qcfractal-server user modify --password` would autogenerate a new password?

dgasmith:
I flipped enough settings and apparently codecov is working again.

mattwelborn:
@dgasmith ping.

I think it's worth you taking another look at this because a lot has changed since you approved the PR:

- docs were added
- `user modify` was added
- tests were added

qcfractal/storage_sockets/sqlalchemy_socket.py
```diff
@@ -1959,28 +1973,28 @@ def add_user(self,
...
-            raise KeyError("Permissions settings not understood: {}".format(set(permissions) - valid_permissions))
+        if not SQLAlchemySocket.VALID_PERMISSIONS >= set(permissions):
+            raise KeyError("Permissions settings not understood: {}".format(
+                set(permissions) - SQLAlchemySocket.VALID_PERMISSIONS))
 
-        return_password = False
         if password is None:
-            password = secrets.token_urlsafe(32)
-            return_password = True
+            password = SQLAlchemySocket._generate_password()
```
dgasmith: Can you call these static methods via `self` and not the class name? This applies to all instances.

qcfractal/storage_sockets/sqlalchemy_socket.py
```diff
@@ -1959,28 +1973,28 @@ def add_user(self,
...
         tuple
-            Successful insert or not
+            A tuple of (success flag, password)
         """
-        valid_permissions = {'read', 'write', 'compute', 'queue', 'admin'}
 
         # Make sure permissions are valid
-        if not valid_permissions >= set(permissions):
-            raise KeyError("Permissions settings not understood: {}".format(set(permissions) - valid_permissions))
+        if not SQLAlchemySocket.VALID_PERMISSIONS >= set(permissions):
```
dgasmith: We typically use `_valid_permissions` for these variables. All caps are used for truly global vars.

qcfractal/storage_sockets/sqlalchemy_socket.py
mattwelborn: Will do.

qcfractal/storage_sockets/sqlalchemy_socket.py
mattwelborn: The all caps is intended to denote a constant:

https://www.python.org/dev/peps/pep-0008/#constants

rather than a class-private variable.

mattwelborn: I guess it could also be a private constant i.e. `_VALID_PERMISSIONS`

mattwelborn:
image
Huh?

qcfractal/storage_sockets/sqlalchemy_socket.py
mattwelborn: renamed to `_valid_permissions` and made into a frozenset

PR #361, opened 2019-08-12
Copy Alembic when creating staging DB

doaa-altarawy:
## Description
Copy Alembic when creating staging DB

## Status
- [ ] Changelog updated
- [x] Ready to go


dgasmith:
This seems to include the Dataset changes as well. Can you change the PR to only contain the alembic diff?

doaa-altarawy:
Ops, I thought I created the branch out of master, sorry.

PR #376, opened 2019-08-20
 Update install docs

Lnaden:
## Description
Installation docs have been updated and tested. I have also stripped out
the install from source and point only to the github for development.

## Status
- [ ] Changelog updated
- [x] Ready to go

### DIFF
--- a/docs/qcfractal/source/install.rst
+++ b/docs/qcfractal/source/install.rst
@@ -7,7 +7,7 @@ The below commands install QCFractal and its required dependencies, but *not* an
 chemistry codes nor the software to run :term:`Queue Managers <Manager>`. This is done to avoid requiring *all* software
 which *can* interface with Fractal, and instead requires the user to obtain the software they individually *require*.
 
-.. _conda_install:
+.. _conda-install:
 
 Conda
 -----
@@ -49,7 +49,7 @@ Pip
    through a native install on your device (e.g. managed clusters), a direct installer, ``yum`` install, a ``conda``
    install, or otherwise; it must be installed first or the ``Psycopg2`` package will complain about missing the
    ``pg_config``. Installation of PostgreSQL manually is beyond the scope of these instructions, so we recommend
-   either using a `Conda Install of Fractal <conda_install>`_ or contacting your systems administrator.
+   either using a :ref:`Conda install of Fractal <conda-install>` or contacting your systems administrator.
 
 If you have PosgreSQL installed already, you can also install QCFractal using ``pip``:
 
@@ -64,7 +64,7 @@ Test the Installation
 .. note::
 
    There are several optional packages Fractal can interface with for additional features such as visualization,
-   :term:`Queue Adapter`s, and services. These are not installed by default and so you can expect many of the tests
+   :term:`Queue Adapters <Queue Adapter>`, and services. These are not installed by default and so you can expect many of the tests
    will be marked with ``skip`` or ``s``.
 
 You can test to make sure that Fractal is installed correctly by first installing ``pytest``.
--- a/docs/qcfractal/source/install.rst
+++ b/docs/qcfractal/source/install.rst
@@ -1,30 +1,30 @@
 Install QCFractal
 =================
 
-You can install qcfractal with ``conda`` (recommended) or with ``pip`` (with some caveats).
+You can install QCFractal with ``conda`` (recommended) or with ``pip`` (with some caveats).
 
 The below commands install QCFractal and its required dependencies, but *not* any of the quantum
 chemistry codes nor the software to run :term:`Queue Managers <Manager>`. This is done to avoid requiring *all* software
-which *can* interface with Fractal, and instead requires the user to obtain the software they individually *require*.
+which *can* interface with QCFractal, and instead requires the user to obtain the software they individually *require*.
 
 .. _conda-install:
 
 Conda
 -----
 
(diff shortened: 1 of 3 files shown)


docs/qcfractal/source/install.rst
```diff
@@ -1,30 +1,35 @@
...
-The qcfractal package is maintained on the
+This installs QCFractal and its dependencies. The qcfractal package is maintained on the
 `conda-forge channel <https://conda-forge.github.io/>`_.
 
-
 Conda Pre-Created Environments
 ++++++++++++++++++++++++++++++
 
-Fractal can also be installed through pre-configured environments you can pull through our Conda Channel::
+Fractal can also be installed through pre-configured environments you can pull through our Conda Channel:
```
dgasmith: Can we write up a paragraph about what an environment is and give a concrete example on how to use it? There seemed to be quite a bit of confusion over these.

docs/qcfractal/source/install.rst
```diff
@@ -1,30 +1,35 @@
...
-    conda install qcfractal -c conda-forge
+.. code-block:: console
 
-The above command installs qcfractal and its required dependencies, but *not* any of the quantum
-chemistry codes nor the software to run :term:`Queue Managers <Manager>`. This is done to avoid requiring *all* software
-which *can* interface with Fractal, and instead requires the user to obtain the software they individually *require*.
+    >>> conda install qcfractal -c conda-forge
 
-The qcfractal package is maintained on the
+This installs QCFractal and its dependencies. The qcfractal package is maintained on the
```
dgasmith: We sometimes say QCFractal and qcfractal. Should we standardize to "\`\`qcfractal\`\`" here?

docs/qcfractal/source/install.rst
```diff
@@ -34,42 +39,57 @@ If you want to use a different name than the environment file, you can add a ``-
...
-    pip install -e .
+Then, run the following command:
 
-This installs all the dependencies to setup a production background in a new conda environment,
-activate the environment, and then install Fractal into development mode.
+.. code-block::
 
-Test
-----
+   >>> pytest -p qcfractal.testing --pyargs qcfractal
```
dgasmith: Cool does this work while in another directory with the `-p` option?

docs/qcfractal/source/install.rst
```diff
@@ -34,42 +39,57 @@ If you want to use a different name than the environment file, you can add a ``-
...
 
-Test qcfractal with ``pytest``::
 
-    cd qcfractal
-    pytest
+Developing from Source
+----------------------
 
+If you are a developer and want to make contributions Fractal, you can access the source code from
```
dgasmith: Nice!

docs/qcportal/source/install.rst
```diff
@@ -1,39 +1,64 @@
...
+   >>> pip install qcportal
+
+
+Test the Installation
+---------------------
+
+.. note::
+
+   QCPortal can interface with `Plotly <https://plot.ly/>`_ for additional visualization features.
+   This is not installed by default and so you can expect the visualization tests to all be marked with
```
dgasmith: It is installed with conda by default?

docs/qcfractal/source/install.rst
Lnaden: Yes. Since the `testing.py` file is importable from `qcfractal`, it does a python import as a plugin and therefore finds the missing fixtures it needs. I think we technically loose access to the `--runslow`` and ``--runexamples`` flags though. But thats an acceptable loss for now.

dgasmith: Gotcha. I think the files it is complaining about is where we forgot to do an explicit `import`. So we might be able to skip this in the future. `--runslow` would be good to keep however since the slow tests make testing a bit tedious.

docs/qcportal/source/install.rst
Lnaden: The tests were still skipped, and I see what happened here. The `S22Fixture` is currently set to always skip.

https://github.com/MolSSI/QCFractal/blob/master/qcfractal/interface/tests/test_visualization.py#L19-L26

docs/qcfractal/source/install.rst
Lnaden: If we want to keep the `--runslow`, we have to ship around the `conftest.py` file which sits at the same level as `setup.py` and not packaged with the code right now.

dgasmith: If we package it in the egg will it still work?

docs/qcportal/source/install.rst
dgasmith: But skip tests and plotly not installed are different?

https://github.com/conda-forge/qcportal-feedstock/blob/master/recipe/meta.yaml#L32

Lnaden: Yes. And I'm currently changing what I wrote here. I just saw all the viz. stuff get skipped in my testing and incorrectly assumed since they all skipped it was from a missing package similar to services in Fractal and qc codes in Engine. 

Do you want me to turn back on the `S22Fixture` now that we have upgraded everything while I am looking at it anyways?

docs/qcfractal/source/install.rst
Lnaden: Not sure. What would happen if we move the `conftest.py` file down one level into the `qcfractal` folder so it is shipped around with the package by default? We dont add it to any of the `__init__.py` files, but I think that would allow pytest to just find it and work if I am following correctly?

PR #377, opened 2019-08-20
PyTest Inner Workings Overhaul

Lnaden:
## Description
This PR makes an overhaul to the nuances of our PyTest framework.

* In order to ship around the testing configuration (the contents of `conftest.py`), this PR ports all of those features to the `testing.py` file which is now treated as a first class PyTest Plugin.
* The `pytest11` entry point for the `testing.py` file has been added to the `setup.py` file so it will be detected automatically by any installed PyTest, allowing transferring of the `--runslow` and `--runexamples` CLI flags into PyTest on its own.
    * This also enables the ability to access these flags anywhere
* Due to the removal of the [pytest.config global](https://docs.pytest.org/en/latest/deprecations.html#pytest-config-global) in the Version 5+ of PyTest, I had to rework how some of our `mark` functions behave. Instead of writing our own decorators, I instead am using the PyTest recommended method of `pytest_collection_modifyitems`, an example of which can be found [here](https://docs.pytest.org/en/latest/example/simple.html#control-skipping-of-tests-according-to-command-line-option). I have also based this off the implementation this from Dask Distributed dask/distributed#2677.
* Due to the previous change, the `mark_slow` and `mark_example` decorators are no more. We now just use `pytest.mark.slow` and `pytest.mark.example` respectively. This change has been propagated through the tests.
* I fixed a bug in the `test_visualization` where the `_has_plotly` flag was always being set to `True`.
* I have tested this with both PyTest 5 and PyTest 4.6.3

And finally, all of this, was so we could install QCFractal and run `pytest --pyargs qcfractal` from anywhere and still have access to the optional `--runslow` flag (and it solved a PyTest 5 deprecation)

## Status
- [ ] Changelog updated
- [x] Ready to go


qcfractal/interface/tests/test_visualization.py
```diff
@@ -8,6 +8,7 @@
 from . import portal
 
 try:
+    import plotly
```
dgasmith: Hah....

qcfractal/testing.py
```diff
@@ -27,6 +27,48 @@
...
+
+def pytest_configure(config):
+    import sys
+    sys._called_from_test = True
+    config.addinivalue_line("markers", "example: Mark a given test as an example which can be run")
+    config.addinivalue_line("markers", "slow: Mark a given test as slower than most other tests, needing a special "
+                                       "flag to run.")
+
+
+def pytest_unconfigure(config):
```
dgasmith: I see you have found the magic incantation. Was it on the bottom of a snake?

qcfractal/tests/test_services.py
```diff
@@ -201,7 +201,7 @@ def test_service_torsiondrive_option_energy_upper_limit(torsiondrive_fixture):
     assert pytest.approx(0.0007991274441437338, abs=1e-6) == final_energies[(-60, )]
 
 
-@mark_slow
+@pytest.mark.slow
```
dgasmith: Ah, I see this approach. Ok, should work!

qcfractal/testing.py
Lnaden: In certain interpretations, it could be. Moreover it required me to sacrifice a Lantern of Diogenes which could only be used in the daytime to summon a bridge across Cocytus to find it.

It also might be left over from the `conftest.py` and I don't know if it actually does anything. 

ONE OR MORE OF THOSE TWO STATEMENTS IS BE TRUE.

qcfractal/tests/test_services.py
Lnaden: We can also extend it with the 3 functions in the plugin in the future.

PR #385, opened 2019-08-22
Overhauls Dataset get/query fields

dgasmith:
## Description
From discussion with @mattwelborn we decided to overhaul the `get/query` functionality in the Dataset collection. This renaming is as follows:
- `get_records` will return either full ResultRecords or a project, no caching. Replaces `query`.
- `get_molecules` will return the Molecule objects for the Dataset.
- `get_values` will return the canonical values in data columns. Replaces `get_history`, but also always returns the view of the query. So `get_values("B3LYP", "sto-3g")` will now return that column while also caching it internally for speed.
- `get_contributed_values` will now return a column rather than the object representation.

The ReactionDataset tech is still being overhauled.

## Status
- [x] Changelog updated
- [x] Ready to go

### DIFF
--- a/qcfractal/interface/collections/dataset.py
+++ b/qcfractal/interface/collections/dataset.py
@@ -1,18 +1,19 @@
 """
 QCPortal Database ODM
 """
+import warnings
 from typing import Any, Dict, List, Optional, Set, Tuple, Union
 
 import numpy as np
 import pandas as pd
-import warnings
+
 from qcelemental import constants
 
-from .collection import Collection
-from .collection_utils import composition_planner, register_collection
 from ..models import ComputeResponse, Molecule, ObjectId, ProtoModel
 from ..statistics import wrap_statistics
 from ..visualization import bar_plot, violin_plot
+from .collection import Collection
+from .collection_utils import composition_planner, register_collection
 
 
 class MoleculeEntry(ProtoModel):
@@ -257,9 +258,8 @@ class Dataset(Collection):
                 query["stoich"] = query.pop("stoichiometry")
             name = self._canonical_name(**query)
             if force or (name not in self.df.columns):
-                self.df[name] = self.get_records(query.pop("method").upper(),
-                                                 projection={"return_result": True},
-                                                 **query)["return_result"]
+                data = self.get_records(query.pop("method").upper(), projection={"return_result": True}, **query)
+                self.df[name] = data["return_result"] * constants.conversion_factor('hartree', self.units)
             ret.append(self.df[name])
 
         return pd.concat(ret, axis=1)
--- a/qcfractal/interface/collections/dataset.py
+++ b/qcfractal/interface/collections/dataset.py
@@ -632,7 +632,8 @@ class Dataset(Collection):
                      indexer: Dict[str, 'ObjectId'],
                      query: Dict[str, Any],
                      projection: Optional[Dict[str, bool]] = None,
-                     merge: bool = False) -> 'pd.Series':
+                     merge: bool = False,
+                     raise_on_plan: Union[str, bool] = False) -> 'pd.Series':
         """
         Runs a query based on an indexer which is index : molecule_id
 
@@ -646,22 +647,27 @@ class Dataset(Collection):
             Description
         merge : bool, optional
             Sum compound queries together, useful for mixing results
+        raise_on_plan : Union[str, bool], optional
+            Raises a KeyError is True or string if a multi-stage plan is detected.
 
         Returns
         -------
-        Series
+        pd.Series
             A Series of the data results
 
-        Raises
-        ------
-        KeyError
-            If no records match the query
         """
         self._check_client()
         self._check_state()
 
         ret = []
-        for query_set in composition_planner(**query):
+        plan = composition_planner(**query)
+        if raise_on_plan and (len(plan) > 1):
+            if raise_on_plan is True:
+                raise KeyError("Recieved a multi-stage plan when this function does not support multi-staged plans.")
+            else:
+                raise KeyError(raise_on_plan)
+
+        for query_set in plan:
 
             # Set the index to remove duplicates
             molecules = list(set(indexer.values()))
--- a/qcfractal/interface/collections/dataset.py
+++ b/qcfractal/interface/collections/dataset.py
@@ -541,6 +541,7 @@ class Dataset(Collection):
         Series
             A Series of the data results
         """
+        self._check_client()
         self._check_state()
 
         field = field.lower()
@@ -592,10 +593,8 @@ class Dataset(Collection):
             compute_keys["keywords"],
             stoich=compute_keys.get("stoich", None))
 
+        self._check_client()
         self._check_state()
-
-        if self.client is None:
-            raise AttributeError("Dataset: Compute: Client was not set.")
 
         umols = list(set(molecules))
 
@@ -712,8 +711,7 @@ class Dataset(Collection):
             The requested KeywordSet
 
         """
-        if self.client is None:
-            raise AttributeError("Dataset: Client was not set.")
+        self._check_client()
 
         alias = alias.lower()
         program = program.lower()
@@ -878,8 +876,7 @@ class Dataset(Collection):
         if (name in self.df) and (force is False):
             return name
 
-        if self.client is None:
-            raise AttributeError("DataBase: FractalClient was not set.")
+        self._check_client()
 
         # # If reaction results
         indexer = {e.name: e.molecule_id for e in self.data.records}
--- a/qcfractal/interface/collections/dataset.py
+++ b/qcfractal/interface/collections/dataset.py
@@ -682,8 +682,7 @@ class Dataset(Collection):
 
         for query_set in plan:
 
-            if isinstance(query_set["keywords"], str):
-                query_set["keywords"] = self.data.alias_keywords[query_set["program"]][query_set["keywords"]]
+            query_set["keywords"] = self.get_keywords(query_set["keywords"], query_set["program"], return_id=True)
 
             # Set the index to remove duplicates
             molecules = list(set(indexer.values()))
@@ -848,7 +847,7 @@ class Dataset(Collection):
             self.data.default_keywords[program] = alias
         return True
 
-    def get_keywords(self, alias: str, program: str) -> 'KeywordSet':
+    def get_keywords(self, alias: str, program: str, return_id: bool = False) -> Union['KeywordSet', str]:
         """Pulls the keywords alias from the server for inspection.
 
         Parameters
@@ -857,14 +856,22 @@ class Dataset(Collection):
             The keywords alias.
         program : str
             The program the keywords correspond to.
+        return_id : bool, optional
+            If True, returns the ``id`` rather than the ``KeywordSet`` object.
+            Description
 
         Returns
         -------
-        KeywordSet
-            The requested KeywordSet
+        Union['KeywordSet', str]
+            The requested ``KeywordSet`` or ``KeywordSet`` ``id``.
 
         """
         self._check_client()
+        if alias is None:
+            if return_id:
+                return None
+            else:
+                return {}
 
         alias = alias.lower()
         program = program.lower()
@@ -872,7 +879,10 @@ class Dataset(Collection):
             raise KeyError("Keywords {}: {} not found.".format(program, alias))
 
         kwid = self.data.alias_keywords[program][alias]
-        return self.client.query_keywords([kwid])[0]
+        if return_id:
+            return kwid
+        else:
+            return self.client.query_keywords([kwid])[0]
 
     def add_contributed_values(self, contrib: ContributedValues, overwrite=False) -> None:
         """Adds a ContributedValues to the database.
--- a/qcfractal/interface/collections/dataset.py
+++ b/qcfractal/interface/collections/dataset.py
@@ -521,20 +521,24 @@ class Dataset(Collection):
 
         return name, dbkeys, history
 
-    def _query(self, indexer: str, query: Dict[str, Any], field: str="return_result") -> 'Series':
+    def _get_records(self,
+                     indexer: Dict[str, 'ObjectId'],
+                     query: Dict[str, Any],
+                     projection: Optional[Dict[str, bool]] = None,
+                     merge: bool = False) -> 'Series':
         """
         Runs a query based on an indexer which is index : molecule_id
 
         Parameters
         ----------
-        indexer : str
-            The primary index of the query
+        indexer : Dict[str, 'ObjectId']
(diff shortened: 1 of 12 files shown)


mattwelborn:
I've looked through the code for dataset, and I didn't find any issues. I'm going to next kick the tires on the new dataset functions.

qcfractal/interface/collections/dataset.py
```diff
@@ -521,44 +582,84 @@ def _default_parameters(self,
...
                 records.extend(self.client.query_results(**query_set))
 
-            records = pd.DataFrame(records, columns=["molecule", field])
+            if projection is None:
+                records = [{"molecule" : x.molecule, "record": x} for x in records]
+
+            records = pd.DataFrame.from_dict(records)
 
             df = pd.DataFrame.from_dict(indexer, orient="index", columns=["molecule"])
             df.reset_index(inplace=True)
```
mattwelborn: This throws `KeyError: 'molecule'` if no records are found. e.g.:
`ds.get_records(program="psi4", method="baloney")`
`ds.get_records(program="psi4",method='b3lyp-d3bj',subset=['baloney']) # 'b3lyp-d3bj' is a valid method for this ds`

qcfractal/interface/collections/dataset.py
```diff
@@ -521,44 +582,84 @@ def _default_parameters(self,
...
+        molecule_ids = list(set(indexer.values()))
+        for i in range(0, len(molecule_ids), self.client.query_limit):
+            molecules.extend(self.client.query_molecules(id=molecule_ids[i:i + self.client.query_limit]))
+
+        molecules = pd.DataFrame.from_dict([{"molecule_id": x.id, "molecule": x} for x in molecules])
+
+        df = pd.DataFrame.from_dict(indexer, orient="index", columns=["molecule_id"])
+        df.reset_index(inplace=True)
+
+        # Outer join on left to merge duplicate molecules
```
mattwelborn: Same issue as on line 668. KeyError on molecule_id if df is empty.

qcfractal/interface/collections/dataset.py
```diff
@@ -217,33 +242,39 @@ def _get_history(self, force: bool = False, **search: Dict[str, Optional[str]])
...
                 query["stoich"] = query.pop("stoichiometry")
-            queries.loc[name, "name"] = self.query(query.pop("method").upper(), force=force, **query)
+            name = self._canonical_name(**query)
+            if force or (name not in self.df.columns):
+                data = self.get_records(query.pop("method").upper(), projection={"return_result": True}, **query)
+                self.df[name] = data["return_result"] * constants.conversion_factor('hartree', self.units)
+            ret.append(self.df[name])
 
-        return queries
+        return pd.concat(ret, axis=1)
```
mattwelborn: `ds.get_values(program='baloney')` causes an error

mattwelborn:
Projection onto a missing field causes an internal server error. Can this be checked client-side?

dgasmith:
Unlikely, but I think we have an issue to give this a better error message. It isn't a 1-1 mapping for Record attributes to SQL fields which is what projection is really checking. I think we can improve that, but would take some time.

qcfractal/interface/collections/dataset.py
```diff
@@ -521,64 +615,136 @@ def _default_parameters(self,
...
+        plan = composition_planner(**query)
+        if raise_on_plan and (len(plan) > 1):
+            if raise_on_plan is True:
+                raise KeyError("Recieved a multi-stage plan when this function does not support multi-staged plans.")
+            else:
+                raise KeyError(raise_on_plan)
+
+        for query_set in plan:
+
+            if isinstance(query_set["keywords"], str):
```
mattwelborn: This doesn't work when `query_set["keywords"]` is an id, which gets passed as a string. Example:
```
ds = client.get_collection("Dataset","OpenFF VEHICLe Set 1")
ds.get_values(program="psi4", keywords="default")
```

`gradient_dataset_fixture` should include keyword sets to catch this in the future.

qcfractal/interface/collections/reaction_dataset.py
```diff
@@ -112,52 +164,133 @@ def _pre_save_prep(self, client):
...
+        -------
+        DataFrame
+            A DataFrame of the queried parameters
+        """
+        warnings.warn(
+            "This is function is deprecated and will be removed in 0.11.0, please use `get_values(..., )` for a instead.",
+            DeprecationWarning)
+
+        # Get default program/keywords
+        return self.get_values(method=method, basis=basis, keywords=keywords, program=program, force=force)
```
mattwelborn: There are two functions called `get_history`, which each take 5 parameters.

qcfractal/interface/tests/test_visualization.py
```diff
@@ -21,8 +21,9 @@ def live_fractal_or_skip():
     """Ensure Fractal live connection can be made"""
 
     try:
+        requests.get('https://api.qcarchive.molssi.org:443', json={}, timeout=5)
         return portal.FractalClient()
-    except requests.exceptions.ConnectionError:
+    except (requests.exceptions.ConnectionError, ConnectionRefusedError):
         return pytest.skip("Could not make a connection to central Fractal server")
```
mattwelborn: The visualization tests fail because `ReactionDataset._visualize` calls `ReactionDataset.get_history`. A few things need to be done:

1. `ReactionDataset._visualize` should call `ReactionDataset.get_values`.
2. `ReactionDataset.get_history` should call  `ReactionDataset.get_values` instead of the non-existent `ReactionDataset._get_history`.
3. There should be a test for `ReactionDataset.get_history` added, which can be removed when `ReactionDataset.get_history` is deprecated. (I think what happened is the `ReactionDataset.get_history` tests were rewritten to use `ReactionDataset.get_values`.)

qcfractal/interface/collections/reaction_dataset.py
```diff
@@ -239,16 +372,120 @@ def visualize(self,
 
         return self._visualize(metric, bench, query=query, groupby=groupby, return_figure=return_figure, kind=kind)
```
mattwelborn: Does not return the correct molecule, e.g.

```
s22 = client.get_collection("ReactionDataset", "S22")
mol = s22.get_molecules(subset='Ammonia Dimer', stoich='default')
print(mol, mol.id)
```
gives:
```
    Geometry (in Angstrom), charge = 0.0, multiplicity = 1:

       Center              X                  Y                   Z       
    ------------   -----------------  -----------------  -----------------
    O                -1.397621297827    -1.885836800365    -0.367306098301
    N                -1.464255001147     0.364182799152     0.019230098748
    C                -4.185739797656     0.369666900716     0.036095997764
    C                -3.483259802214     1.578311100122     0.250075200651
    C                -2.117950202456     1.530704799511     0.233838297543
    C                -2.077383297566    -0.863749200485    -0.189941402177
    C                -3.515603197998    -0.805195001276    -0.175758500412
    H                -5.267804501347     0.370742797362     0.041141898264
    H                -3.992033400169     2.512755999947     0.421441397921
    H                -1.492919599297     2.398409599646     0.388501799340
    H                -4.040122601443    -1.734845201024    -0.337926899246
    H                -0.426526600411     0.361212702097     0.007353801068
    N                 1.432761601549     0.363970297460    -0.015950798157
    C                 2.115419999833    -0.780344998216     0.168109900012
    C                 3.523758601997    -0.801609598128     0.154502700718
    C                 4.218589700700     0.373578298193    -0.052592901512
    C                 3.509970799813     1.561501399725    -0.244976297763
    C                 2.128013798309     1.495332398665    -0.217537401035
    H                 4.045920600345    -1.736135599652     0.307688300525
    H                 5.299942597541     0.366600900875    -0.066334898116
    H                 4.011092299506     2.502431297474    -0.413005201910
    H                 1.533987799051     2.389383699136    -0.367056501286
    N                 1.388312298093    -1.908303802027     0.419814897679
    H                 1.869471402447    -2.781277302142     0.294038502466
    H                 0.408906697955    -1.907994201608     0.130085998541
 24
```

This is not caught by the test suite, as there is not test for `ReactionDataset.get_molecules`

qcfractal/interface/collections/reaction_dataset.py
```diff
@@ -239,16 +372,120 @@ def visualize(self,
...
+        subset : Optional[Union[str, Set[str]]], optional
+            The index subset to query on
+
+        Returns
+        -------
+        Union[pd.DataFrame, 'ResultRecord']
+            The name of the queried column
+
+        Examples
+        --------
```
mattwelborn: This example needs to be updated.

qcfractal/interface/collections/reaction_dataset.py
```diff
@@ -239,16 +372,120 @@ def visualize(self,
...
+            The computational basis to query on (6-31G)
+        keywords : Optional[str], optional
+            The option token desired
+        program : Optional[str], optional
+            The program to query on
+        stoich : Union[str, List[str]], optional
+            The given stoichiometry to compute.
+        projection : Optional[Dict[str, bool]], optional
+            The attribute project to perform on the query, otherwise returns ResultRecord objects.
+        subset : Optional[Union[str, Set[str]]], optional
```
mattwelborn: The subset keyword does not appear to work:

```
s22.get_records("b3lyp", subset="Water Dimer").reset_index()["name"]
```
gives:
```
0     2-Pyridone-2-Aminopyridine Complex
1     2-Pyridone-2-Aminopyridine Complex
2     2-Pyridone-2-Aminopyridine Complex
3     2-Pyridone-2-Aminopyridine Complex
4          Adenine-Thymine Complex Stack
                     ...                
83                    Uracil Dimer Stack
84                           Water Dimer
85                           Water Dimer
86                           Water Dimer
87                           Water Dimer
Name: name, Length: 88, dtype: object
```

I think this is because `subset` is only used as an argument to `_molecule_indexer`, and `_molecule_indexer` ignores `subset`.

qcfractal/interface/collections/reaction_dataset.py
```diff
@@ -92,9 +95,58 @@ def _form_index(self):
...
+                          stoich: Union[str, List[str]],
+                          subset: Optional[Union[str, Set[str]]] = None,
+                          coefficients: bool = False) -> Tuple[Dict[Tuple[str, ...], 'ObjectId'], Tuple[str]]:
+        """Provides a {index: molecule_id} mapping for a given subset.
+
+        Parameters
+        ----------
+        stoich : Union[str, List[str]]
+            The stoichiometries, or list of stoichiometries to return
+        subset : Optional[Union[str, Set[str]]], optional
```
mattwelborn: `subset` is not used in this function.

qcfractal/interface/collections/reaction_dataset.py
mattwelborn: I think this relates to `subset` not being used in `_molecule_indexer`, see below.

dgasmith:
I figured this would be pretty rough since it was a deep and haphazard time. Let me see what I can work out.

qcfractal/interface/collections/dataset.py
dgasmith: In general do we want to support non-aliased keywords through here? That is extra functionality which may be useful, but starts to become hard to tell what is an alias and what is an Id.

qcfractal/interface/collections/reaction_dataset.py
dgasmith: Whoops, this was the heart of the issue. Fixed.

qcfractal/interface/collections/dataset.py
```diff
@@ -761,22 +912,7 @@ def list_contributed_values(self) -> List[str]:
...
-            The ContributedValues object key.
-
-        Returns
-        -------
-        ContributedValues
-            The requested ContributedValues object.
-        """
-        return self.data.contributed_values[key.lower()].copy()
-
-    def get_contributed_values_column(self, key: str) -> 'Series':
```
mattwelborn: Deprecate get_contributed_values_column

qcfractal/interface/collections/dataset.py
mattwelborn: This is still an issue. 

image

mattwelborn: I think this solves the issue:

```
-            if isinstance(query_set["keywords"], str):
-                query_set["keywords"] = self.data.alias_keywords[query_set["program"]][query_set["keywords"]]
+            query_set["keywords"] = self.data.alias_keywords[query_set["program"]].get(query_set["keywords"], query_set["keywords"])
```

qcfractal/interface/collections/reaction_dataset.py
```diff
@@ -234,21 +328,121 @@ def visualize(self,
...
+
+        self._check_client()
+        self._check_state()
+        self._validate_stoich(stoich)
+
+        indexer, names = self._molecule_indexer(stoich, subset)
+        df = self._get_molecules(indexer)
+        df.index = pd.MultiIndex.from_tuples(df.index, names=names)
+
+        if isinstance(subset, str) and isinstance(stoich, str):
```
mattwelborn: This gives incorrect output when subset is a string and stoich is a string which corresponds to more than one molecule. For example,
```
s22.get_molecules(subset='Water Dimer', stoich='default1')
```
gives only one of two molecules.

qcfractal/interface/collections/dataset.py
dgasmith: Low level and generally unused. Leaning towards just nuking it.

qcfractal/interface/collections/dataset.py
dgasmith: Keywords must be an alias in general. I have patched this slightly, but I still don't understand the use case here.

qcfractal/interface/collections/dataset.py
mattwelborn: Yeah, we can do that. Probably 0 uses in the wild.

qcfractal/interface/collections/dataset.py
mattwelborn: The problem was that this broke Dataset.get_values. It is now fixed.

qcfractal/interface/collections/reaction_dataset.py
```diff
@@ -234,21 +328,118 @@ def visualize(self,
...
+        subset : Optional[Union[str, Set[str]]], optional
+            The index subset to query on
+
+        Returns
+        -------
+        Union[pd.DataFrame, 'ResultRecord']
+            The name of the queried column
+
+        """
+
```
mattwelborn: Remove this warning.

mattwelborn:
@dgasmith I'm ready to merge this if you are.

dgasmith:
Agreed, in we go.

PR #387, opened 2019-08-23
Replace autodoc tech with Elemental's implementation

Lnaden:
## Description
Bumps the dependency on Elemental to use the 0.7's implementation of the auto-documentation tech.

Imports the `AutodocBaseSettings` into the common models from elemental.

## Status
- [ ] Changelog updated
- [x] Ready to go


qcfractal/interface/models/common_models.py
```diff
@@ -6,7 +6,7 @@
 
 from pydantic import Schema, validator
 
-from qcelemental.models import Molecule, Provenance, ProtoModel
+from qcelemental.models import Molecule, Provenance, ProtoModel, AutodocBaseSettings
```
dgasmith: Is `AutodocBaseSettings` used here?

Lnaden: No, but we had talked about having all Elemental models pull from here at one point and importing them everywhere else in Fractal. So its not used here, but its imported by __init__ and then further imported elsewhere like `queue_manager`

Lnaden:
I'm going to merge this now so its in, and if we want to change up the imports in a smaller PR we can

Issue #128, opened 2019-01-29
Procedure partial queries


dgasmith:
We will need the ability to perform partial queries on procedures. An example of this is as follows:

1) We have a *n* TorsionDrives which "own" *m* optimization runs.
2) Pulling all *m* optimizations for all *n* TorsionDrive can be too much data.
3) The TorsionDrive only wants to know about several quantities these are:
 - The `final_molecule` of each optimization
 - The `final_energy` of each optimization
 - The `final_result` of each optimization

Planned implementation:

```
GET: procedures/optimization/

Body:
{
    "meta":{},
    "data": {"ids": List[str],
             "quantity": ["final_molecule", "final_energy", "final_result", ...]}
}

Return:
{
    "meta":{"normal_data"}
    "data": Dict["id", quantity]
}
```

mattwelborn:
Is this closed by #393?

dgasmith:
Yup.

PR #393, opened 2019-08-29
Advanced Queries for more complex access to the DB

doaa-altarawy:
## Description
Advanced Queries for more complex access to the DB

## Todos/Questions
Add example queries that you need.
  - [x] Get molecule ids of the optimization procedure of a Torsion drive
  - [x] Get ACTUAL molecule of the optimization procedure of a Torsion drive --> what's the returning format needed? SQL returns tupes (not dicts), are they needed in the REST layer as dict?

## Status
- [x] Ready to go
- [x] update changelog

### DIFF
--- a/qcfractal/interface/models/rest_models.py
+++ b/qcfractal/interface/models/rest_models.py
@@ -1145,7 +1145,7 @@ register_model("queue_manager", "PUT", QueueManagerPUTBody, QueueManagerPUTRespo
 
 ## advanced procedures queries
 
-class OptimizationFinalResultsBody(ProtoModel):
+class OptimizationFinalResultBody(ProtoModel):
     class Data(ProtoModel):
         optimization_ids: QueryObjectId = Schema(
             None,
@@ -1162,7 +1162,7 @@ class OptimizationFinalResultsBody(ProtoModel):
         description="The keys with data to search the database on for Procedures."
     )
 
-class OptimizationAllResultsBody(ProtoModel):
+class OptimizationAllResultBody(ProtoModel):
     class Data(ProtoModel):
         optimization_ids: QueryObjectId = Schema(
             None,
@@ -1179,7 +1179,7 @@ class OptimizationAllResultsBody(ProtoModel):
         description="The keys with data to search the database on for Procedures."
     )
 
-class OptimizationInitialMoleculesBody(ProtoModel):
+class OptimizationInitialMoleculeBody(ProtoModel):
     class Data(ProtoModel):
         optimization_ids: QueryObjectId = Schema(
             None,
@@ -1196,7 +1196,7 @@ class OptimizationInitialMoleculesBody(ProtoModel):
         description="The keys with data to search the database on for Procedures."
     )
 
-class OptimizationFinalMoleculesBody(ProtoModel):
+class OptimizationFinalMoleculeBody(ProtoModel):
     class Data(ProtoModel):
         optimization_ids: QueryObjectId = Schema(
             None,
@@ -1213,7 +1213,7 @@ class OptimizationFinalMoleculesBody(ProtoModel):
         description="The keys with data to search the database on for Procedures."
     )
 
-class ResultsResponse(ProtoModel):
+class ResultResponse(ProtoModel):
     meta: ResponseGETMeta = Schema(
         ...,
         description=common_docs[ResponseGETMeta]
@@ -1225,7 +1225,7 @@ class ResultsResponse(ProtoModel):
         description="A List of Results found from the query per optimization id."
     )
 
-class ListResultsResponse(ProtoModel):
+class ListResultResponse(ProtoModel):
     meta: ResponseGETMeta = Schema(
         ...,
         description=common_docs[ResponseGETMeta]
@@ -1237,7 +1237,7 @@ class ListResultsResponse(ProtoModel):
         description="A List of Results found from the query per optimization id."
     )
 
-class ListMoleculesResponse(ProtoModel):
+class ListMoleculeResponse(ProtoModel):
     meta: ResponseGETMeta = Schema(
         ...,
         description=common_docs[ResponseGETMeta]
@@ -1249,7 +1249,7 @@ class ListMoleculesResponse(ProtoModel):
         description="A List of Molecules found from the query per optimization id."
     )
 
-register_custom_model("optimization", "final_results", OptimizationFinalResultsBody, ResultsResponse)
-register_custom_model("optimization", "all_results", OptimizationAllResultsBody, ListResultsResponse)
-register_custom_model("optimization", "initial_molecules", OptimizationAllResultsBody, ListMoleculesResponse)
-register_custom_model("optimization", "final_molecules", OptimizationAllResultsBody, ListMoleculesResponse)
+register_custom_model("optimization", "final_result", OptimizationFinalResultBody, ResultResponse)
+register_custom_model("optimization", "all_results", OptimizationAllResultBody, ListResultResponse)
+register_custom_model("optimization", "initial_molecule", OptimizationAllResultBody, ListMoleculeResponse)
+register_custom_model("optimization", "final_molecule", OptimizationAllResultBody, ListMoleculeResponse)
--- a/qcfractal/interface/models/rest_models.py
+++ b/qcfractal/interface/models/rest_models.py
@@ -679,6 +679,23 @@ class OptimizationBestResultsBody(ProtoModel):
         description="The keys with data to search the database on for Procedures."
     )
 
+class OptimizationAllResultsBody(ProtoModel):
+    class Data(ProtoModel):
+        opt_ids: QueryObjectId = Schema(
+            None,
+            description="List of optimization procedure Ids to fetch their ALL their results from the database."
+        )
+
+    # TODO: not yet supported
+    meta: QueryMetaProjection = Schema(
+        QueryMetaProjection(),
+        description=common_docs[QueryMetaProjection]
+    )
+    data: Data = Schema(
+        ...,
+        description="The keys with data to search the database on for Procedures."
+    )
+
 class ListResultsResponse(ProtoModel):
     meta: ResponseGETMeta = Schema(
         ...,
@@ -691,6 +708,7 @@ class ListResultsResponse(ProtoModel):
     )
 
 register_model("optimization", "best_opt_results", OptimizationBestResultsBody, ListResultsResponse)
+register_model("optimization", "best_all_results", OptimizationAllResultsBody, ListResultsResponse)
 
 
 ### Task Queue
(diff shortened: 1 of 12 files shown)


dgasmith:
For torsiondrives I likely want to return a structure like:
```
{"td_id1": {"(90, )": thing1, "(120, )": thing2, ...}, ...}
```

In general, I think people will want the Molecule objects over just the Id's. Some returns that we might want to consider: 
 - The final molecule of each "best" optimization (alternatively just the id if requested)
 - The final ResultRecord of each "best" optimization, can I apply a projection to these records as well?
 - The `return_result` of each part of the trajectory in each of the best optimizations.

doaa-altarawy:
I'm implementing the `ResturnResults` query. 

To double check (logically, not how it's implemented), given a torsion_id:
get the optimization_ids from optmization_history.
Then for each optimization, check it's trajectory, and then get the return_results values of it's results.

What's the definition of best?

doaa-altarawy:
@dgasmith

doaa-altarawy:
@dgasmith  Please see my question above. Currently, I return all `ReturnResults`

dgasmith:
The definition of best is the optimizations index by the `minimum_positions` array.

doaa-altarawy:
As we discussed, some queries will return hex strings instead of memory views of bytes, so this needs to be included in the msgpack loading in QCElemental if the data is hex string:
`bytes.fromhex(data[2:])`

qcfractal/storage_sockets/db_queries.py
```diff
@@ -0,0 +1,193 @@
...
+
+        if opt_ids is None:
+            self._raise_missing_attribute('best_opt_results', 'List of optimizations ids')
+
+        sql_statement = text("""
+            select opt_id, result.* from result
+            join (
+                select opt.opt_id, opt.result_id, max_pos from opt_result_association as opt
+                inner join (
+                        select opt_id, max(position) as max_pos from opt_result_association
```
dgasmith: `best` is defined as the `position == torsiondrive.minimum_positions`. It gets worse as `torsiondrive.minimum_positions` is a dictionary of `key: position`.

I thought we had discussed dropping torsiondrives entirely as query optimization points due to this issue?

qcfractal/storage_sockets/db_queries.py
```diff
@@ -0,0 +1,193 @@
...
+
+        sql_statement = f"""
+                select initial_molecule from optimization_procedure as opt where opt.id in
+                (
+                    select opt_id from optimization_history where torsion_id = {torsion_id}
+                )  
+                order by opt.id
+        """
+
+        return self.execute_query(sql_statement, with_keys=False)
```
dgasmith: We need to unpack and transform this to Pydantic objects at this layer as the socket layer does not do this transformation.

qcfractal/storage_sockets/sqlalchemy_socket.py
```diff
@@ -297,6 +303,48 @@ def get_query_projection(self, className, query, projection, limit, skip, exclud
 
         return rdata, n_found
 
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+    def query(self, class_name : str, query_key : str, **kwargs):
```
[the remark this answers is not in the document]
doaa-altarawy: Are they going to be related to procedures only?
I thought by sending class_name, we can also have advanced ones for collections that's used internally in the Pydantic Collections.

qcfractal/storage_sockets/db_queries.py
doaa-altarawy: This table we are looking up is not the torsion_drive history, this table is the trajectory of a single optimization_procedure. So, I'm assuming the REST gave me a list of opt_ids already.

dgasmith: Shouldn't this be moved to an `OptimizationQuery` class in this case?

qcfractal/storage_sockets/sqlalchemy_socket.py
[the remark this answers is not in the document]
dgasmith: Guessing we will need a more expansive `collection` query techniques. Keeping them separate would help with the complexity of the API I think.

qcfractal/storage_sockets/db_queries.py
doaa-altarawy: I think it's better to keep them as dicts, and then move them back to the client where they transformed into Pydantic. This means one transformation only, instead of a second one before the REST, then serialize it again before sending to the client.

dgasmith: We uniformly across the entire storage socket transform them to Pydantic objects before sending them to the client. This needs to be the standard to validate and we will have a switch to turn this off for production eventually.

qcfractal/interface/models/rest_models.py
```diff
@@ -37,36 +46,45 @@ def register_model(name: str, rest: str, body: 'ProtoModel', response: 'ProtoMod
...
 
     """
     try:
-        return __rest_models[name.lower()][rest.upper()]
+        if resource.lower() in __custom_rest_models:
+            return __custom_rest_models[resource.lower()][subresource.upper()]
+        else:
+            return __rest_models[resource.lower()][rest.upper()]
     except KeyError:
         raise KeyError(f"REST Model {name.lower()}:{rest.upper()} could not be found.")
```
dgasmith: The error needs to be expanded to include the subresource requests since this is what the user will see.

qcfractal/interface/models/rest_models.py
```diff
@@ -1122,3 +1140,115 @@ class QueueManagerPUTResponse(ProtoModel):
...
 register_model("queue_manager", "PUT", QueueManagerPUTBody, QueueManagerPUTResponse)
+
+
+## advanced procedures queries
+
+class OptimizationBestResultsBody(ProtoModel):
+    class Data(ProtoModel):
+        opt_ids: QueryObjectId = Schema(
+            None,
+            description="List of optimization procedure Ids to fetch their BEST results from the database."
```
dgasmith: I think we need to call this `Last` or `Final` results to more accurately reflect the operation.

qcfractal/interface/models/rest_models.py
```diff
@@ -1122,3 +1140,115 @@ class QueueManagerPUTResponse(ProtoModel):
...
+    meta: QueryMetaProjection = Schema(
+        QueryMetaProjection(),
+        description=common_docs[QueryMetaProjection]
+    )
+    data: Data = Schema(
+        ...,
+        description="The keys with data to search the database on for Procedures."
+    )
+
+class OptimizationFinalMoleculesBody(ProtoModel):
```
dgasmith: We usually try to keep plurality here `OptimizationInitialMoleculesBody` -> `OptimizationInitialMoleculeBody`. Same for `OptimizationBestResultsBody` -> `OptimizationFinalResultBody`.

qcfractal/interface/models/rest_models.py
```diff
@@ -1122,3 +1140,115 @@ class QueueManagerPUTResponse(ProtoModel):
 
 
 register_model("queue_manager", "PUT", QueueManagerPUTBody, QueueManagerPUTResponse)
+
+
+## advanced procedures queries
+
+class OptimizationBestResultsBody(ProtoModel):
+    class Data(ProtoModel):
+        opt_ids: QueryObjectId = Schema(
```
dgasmith: We usually try to write this out as `optimization_ids` to be verbose. What do you think?

qcfractal/storage_sockets/db_queries.py
```diff
@@ -0,0 +1,298 @@
...
+            on base_result.id = result.id
+        """)
+
+        # bind and expand ids list
+        sql_statement = sql_statement.bindparams(bindparam("opt_ids", expanding=True))
+
+        # column types:
+        columns = inspect(ResultORM).columns
+        sql_statement = sql_statement.columns(opt_ids=Integer, *columns)
+        query_result = self.execute_query(sql_statement, opt_ids=list(opt_ids))
```
dgasmith: Cool, glad this worked!

qcfractal/storage_sockets/sqlalchemy_socket.py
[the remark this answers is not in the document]
dgasmith: Can we move this to `custom_query`?

qcfractal/web_handlers.py
```diff
@@ -335,17 +335,46 @@ class ProcedureHandler(APIHandler):
...
+                ret = self.storage.query('procedure', query_type, **{**body.data.dict(), **body.meta.dict()})
         except KeyError as e:
             raise tornado.web.HTTPError(status_code=401, reason=str(e))
 
         response = response_model(**ret)
 
         self.logger.info("GET: Procedures - {} pulls.".format(len(response.data)))
         self.write(response)
+
+class OptimizationHandler(APIHandler):
```
dgasmith: This looks duplicated from above? We could also consider a `CustomQueryHandler` generic.

`localhost:7777/custom/optimization/`

qcfractal/interface/models/rest_models.py
doaa-altarawy: Yes, that's better

qcfractal/interface/models/rest_models.py
doaa-altarawy: I think Optimization being singular is good and following the rule, but the second part (i.e Molecules) indicates that the return per opt is a list of associated objects. And overall, the class applies for a list of optimizations, returning list of Molecules.

qcfractal/web_handlers.py
doaa-altarawy: This won't be quite REST, since you have to have the resource first. Also, in reality, I think all methods should be under the same object, the custom_queries are not actually custom from the user perspective, it has a different implementation in the backend, that's it.

So, I was thinking that all custom queries will be under their main REST resource. So if we add a new query for optimizaton, it will automatically go through `optimization/my_query_type/`. Same for tasks, etc.

qcfractal/interface/models/rest_models.py
dgasmith: Hmm, I always thought of it as a "Molecule per id" not "Molecule". Minor point.

qcfractal/web_handlers.py
dgasmith: Sure I like `procedure/...`. My main comment here is that this class seems extraneous. The tornado handler points to a `/optimization/...` resource at the moment.

qcfractal/interface/models/rest_models.py
doaa-altarawy: Ok, I changed them to singular.

dgasmith:
This will close #128.

dgasmith:
Can you rebase this? Should pass now.

PR #396, opened 2019-09-03
Fix bug in Dataset._compute

mattwelborn:
## Description
This PR fixes a bug in Dataset._compute that was caused by an indentation error. The bug causes the returned ComputeResponse to be truncated when the number of calculations is larger than the `query_limit`. A test has been added. 

## Status
- [ ] Changelog updated
- [x] Ready to go


dgasmith:
Maybe try retriggering travis? Merge when ready.

PR #402, opened 2019-09-18
subsets in Dataset.get_values (and molecule cache)

mattwelborn:
## Description
This PR adds a subset argument to `Dataset.get_values` and `ReactionDataset.get_values`. Caching works with subsets. 

This PR also:

- Adds a cache for molecules in `Dataset`.
- Adds a voluntary query limit of 100,000 to the client. This is intended to prevent users from accidentally calling get_molecules/values/records on the entirety of ANI-1. 
- Raises a KeyError if a string subset is requested but more than one value is returned (e.g. `s22.get_values(method='b3lyp', subset='Water Dimer')`. 

## Status
- [ ] Changelog updated
- [ ] Ready to go

### DIFF
--- a/qcfractal/interface/client.py
+++ b/qcfractal/interface/client.py
@@ -1,5 +1,4 @@
 """Provides an interface the QCDB Server instance"""
-
 import json
 import os
 import re
@@ -60,7 +59,6 @@ class FractalClient(object):
             FractalServer was not provided a SSL certificate and defaults back to self-signed
             SSL keys.
         """
-
         if hasattr(address, "get_address"):
             # We are a FractalServer-like object
             verify = address.client_verify
@@ -99,11 +97,16 @@ class FractalClient(object):
         self._headers["Content-Type"] = f'application/{self.encoding}'
         self._headers["User-Agent"] = f"qcportal/{__version__}"
 
+        self._request_counter = defaultdict(int)
+
+        ### Define all attributes before this line
+
         # Try to connect and pull general data
         self.server_info = self._automodel_request("information", "get", {}, full_return=True).dict()
 
         self.server_name = self.server_info["name"]
         self.query_limit = self.server_info["query_limit"]
+        self.voluntary_query_limit = 100_000
 
         if _isportal:
             try:
@@ -216,6 +219,7 @@ class FractalClient(object):
         Any
             The REST response object
         """
+        self._request_counter[(name, rest)] += 1
         body_model, response_model = rest_model(name, rest)
 
         # Provide a reasonable traceback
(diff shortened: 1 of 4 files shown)


qcfractal/interface/client.py
```diff
@@ -1058,3 +1062,24 @@ def modify_services(self,
         }
 
         return self._automodel_request("service_queue", "put", payload, full_return=full_return)
+
+    def apply_voluntary_query_limit(self, nqueries: int) -> None:
```
dgasmith: Private method here.

qcfractal/interface/client.py
```diff
@@ -1058,3 +1062,24 @@ def modify_services(self,
...
+        Checks that a planned number of queries is not larger than a guardrail value.
+
+        Parameters
+        ----------
+        nqueries: int
+            The number of planned queries.
+        Returns
+        -------
+        None
+        Raises
```
dgasmith: Spaces between headers here.

qcfractal/interface/collections/dataset.py
```diff
@@ -296,11 +298,16 @@ def get_values(self,
...
             Data is typically cached, forces a new query if True.
-
+        subset : Optional[Union[str, Set[str]]], optional
+            The index subset to query on
         Returns
         -------
-        DataFrame
-            A DataFrame of the queried parameters
+         Union[pd.DataFrame, Any]
+            Either a DataFrame of indexed values or a single value if a single subset string was provided.
```
dgasmith: Ditto here, there must be spaces between sections.

qcfractal/interface/collections/dataset.py
```diff
@@ -622,20 +639,41 @@ def _get_molecules(self, indexer: Dict[str, 'ObjectId']) -> 'pd.Series':
...
-            molecules.extend(self.client.query_molecules(id=molecule_ids[i:i + self.client.query_limit]))
 
-        molecules = pd.DataFrame.from_dict([{"molecule_id": x.id, "molecule": x} for x in molecules])
-        if len(molecules) == 0:
+        # Partition request into cached and uncached molecules
+        cached_ids = []
+        cached_molecules = []
+        uncached_ids = []
+        uncached_molecules = []
+        for molecule_id in molecule_ids:
```
dgasmith: It might be better to cache at the client level. Where for all queries that include `id` we cache say 100k of them. This would be more universal and  would make all of this simpler.

mattwelborn: This depends on whether we want an in-memory cache for molecules loaded from a `.download()`-ed HDF5 file or just for network-retrieved molecules.

mattwelborn:
This is getting quickly out of data, and will need to be mostly redone after HDF5.

PR #403, opened 2019-09-18
Fix for Dataset.get_values+DFTD3

mattwelborn:
## Description
This PR fixes a bug in `Dataset.get_values` for the case of `method = "*-d3"`. 

## Status
- [ ] Changelog updated
- [ ] Ready to go

### DIFF
--- a/qcfractal/interface/collections/dataset.py
+++ b/qcfractal/interface/collections/dataset.py
@@ -327,7 +327,7 @@ class Dataset(Collection):
             name = self._canonical_name(**query)
             if force or (name not in self.df.columns):
                 self._column_metadata[name] = query
-                data = self.get_records(query.pop("method").upper(), projection={"return_result": True}, **query)
+                data = self.get_records(query.pop("method").upper(), projection={"return_result": True}, merge=True, **query)
                 self.df[name] = data["return_result"] * constants.conversion_factor('hartree', self.units)
 
             ret.append(self.df[name])
@@ -984,7 +984,8 @@ class Dataset(Collection):
               keywords: Optional[str]=None,
               program: Optional[str]=None,
               projection: Optional[Dict[str, bool]]=None,
-              subset: Optional[Union[str, Set[str]]] = None) -> Union[pd.DataFrame, 'ResultRecord']:
+              subset: Optional[Union[str, Set[str]]] = None,
+              merge: bool = False) -> Union[pd.DataFrame, 'ResultRecord']:
         """Queries full ResultRecord objects from the database.
 
         Parameters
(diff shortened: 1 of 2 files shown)


qcfractal/tests/test_collections.py
```diff
@@ -159,6 +159,35 @@ def test_dataset_compute_response(fractal_compute_server):
     assert len(response.ids) == 2
 
 
+def test_dataset_d3(fractal_compute_server):
+    testing.check_has_module("psi4")
```
dgasmith: `testing.check_has_module("psi4")` is only needed for fixtures. Can you use the current -d3 dataset to test this out rather than building a new one?

qcfractal/interface/collections/dataset.py
```diff
@@ -1001,6 +1002,8 @@ def get_records(self,
             The attribute project to perform on the query, otherwise returns ResultRecord objects.
         subset : Optional[Union[str, Set[str]]], optional
             The index subset to query on
+        merge : bool
+            Merge multiple results into one (as in the case of DFT-D3)
```
dgasmith: We may need to mention that this will only work for `return_results`.

qcfractal/tests/test_collections.py
```diff
@@ -159,6 +159,35 @@ def test_dataset_compute_response(fractal_compute_server):
...
+    # Add two helium dimers to the DB at 4 and 8 bohr
+    HeDimer = ptl.Molecule.from_data([[2, 0.343, 0, -4.123], [2, 0, 0, 4.123]], dtype="numpy", units="bohr", frags=[1])
+    ds.add_entry("HeDimer", HeDimer)
+    ds.set_default_program("psi4")
+    ds.add_keywords("scf_default", "psi4", ptl.models.KeywordSet(values={}), default=True)
+
+    ds.save()
+
+    ds.compute("B3LYP", "sto-3g")
+    ds.compute("B3LYP-D3", "sto-3g")
```
dgasmith: I don't see any new tests that are not in `test_compute_reactiondataset_dftd3`?

mattwelborn: This test works on Dataset instead of ReactionDataset. The bug was present in Dataset only, so the previous test did not catch it.

qcfractal/tests/test_collections.py
```diff
@@ -322,6 +292,29 @@ def test_rectiondataset_dftd3_molecules(reactiondataset_dftd3_fixture_fixture):
     assert set(stoichs) == set(mols["stoichiometry"])
 
 
+def test_dataset_d3(reactiondataset_dftd3_fixture_fixture):
```
dgasmith: Can we call this `dftd3` like the other DFTD3 test? Otherwise merge when ready.

PR #409, opened 2019-09-20
Skip broken intel-openmp

mattwelborn:
## Description
This PR skips intel-openmp version 2019.5-281, which breaks geometric+psi4. See also, https://github.com/ContinuumIO/anaconda-issues/issues/11294.

### DIFF
--- a/devtools/conda-envs/openff.yaml
+++ b/devtools/conda-envs/openff.yaml
@@ -31,7 +31,7 @@ dependencies:
   - rdkit
   - geometric>=0.9.3
   - torsiondrive
-  - intel-openmp!=2019.5-281
+  - intel-openmp!=2019.5
 
 # QCArchive includes
   - qcengine>=0.9.0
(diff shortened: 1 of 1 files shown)


dgasmith:
This was kind of pain to google, the only way I found it is a new tool: https://www.gitmemory.com/msarahan

Kind of an interesting tool: https://www.gitmemory.com/molssi/qcfractal

devtools/conda-envs/openff.yaml
```diff
@@ -31,6 +31,7 @@ dependencies:
   - rdkit
   - geometric>=0.9.3
   - torsiondrive
+  - intel-openmp!=2019.5-281
```
dgasmith: Do you want to skip all of 2019.5 or just 281?

mattwelborn: It turns out that "2019.5-281" doesn't register with conda as a version, so now I'm skipping "2019.5".

PR #411, opened 2019-09-25
Dataset.list_values and modified get_values interface

mattwelborn:
## Description
This PR introduces a new `list_values` function for datasets to replace `list_history` and `list_contributed_values`. It also changes `get_values` to return contributed values in addition to native values (controlled by the `native` argument).

In addition, this PR:
- Adds a `name` selector to `get_values`.
- Adds rudimentary units support to `get_values`.
- Bumps deprecation warnings to version v0.12

## Todos
- [x] Run yapf
- [x] check for things that depend on _column_metadata 

## Status
- [x] Changelog updated
- [ ] Ready to go

### DIFF
--- a/qcfractal/interface/collections/dataset.py
+++ b/qcfractal/interface/collections/dataset.py
@@ -1088,6 +1088,8 @@ class Dataset(Collection):
 
         for cv_name, theory_level_details in cvs:
             spec = {"name": cv_name}
+            for k in self.data.history_keys:
+                spec[k] = "Unknown"
             # ReactionDataset uses "default" as a default value for stoich, but many contributed datasets lack a stoich field
             if "stoichiometry" in self.data.history_keys:
                 spec["stoich"] = "default"
--- a/qcfractal/interface/collections/dataset.py
+++ b/qcfractal/interface/collections/dataset.py
@@ -214,19 +214,41 @@ class Dataset(Collection):
                 "driver": driver}
 
         if native in {True, None}:
-            df = self.list_records(dftd3=False, **spec)
+            df = self._list_records(dftd3=False)
             df['native'] = True
             ret.append(df)
 
         if native in {False, None}:
-            df = self._list_contributed_values(force=force, **spec)
+            df = self._list_contributed_values()
             df['native'] = False
             ret.append(df)
 
         ret = pd.concat(ret)
-        ret.reset_index(inplace=True)
+
+        # Filter
+        ret.fillna("None", inplace=True)
+        ret = self._filter_records(ret, **spec)
         ret.set_index(["native"] + list(self.data.history_keys[:-1]), inplace=True)
         ret.sort_index(inplace=True)
+        return ret
+
+    @staticmethod
+    def _filter_records(df: pd.DataFrame, **spec):
+        """
+        Helper for filtering records on a spec. Note that `None` is a wildcard while `"None"` matches `None` and NaN.
+        """
+        ret = df.copy()
+        for key, value in spec.items():
+            if value is None:
+                continue
+            elif isinstance(value, str):
+                value = value.lower()
+                ret = ret[ret[key].fillna("None").str.lower() == value]
+            elif isinstance(value, (list, tuple)):
+                query = [x.lower() for x in value]
+                ret = ret[ret[key].fillna("None").str.lower().isin(query)]
+            else:
+                raise TypeError(f"Search type {type(value)} not understood.")
         return ret
 
     def list_history(self,
@@ -259,15 +281,13 @@ class Dataset(Collection):
     def list_records(self,
                      dftd3: bool = False,
                      pretty: bool = True,
-                     **search: Dict[str, Optional[str]]) -> 'DataFrame':
+                     **search: Dict[str, Optional[str]]) -> pd.DataFrame:
         """
         Lists specifications of available records, i.e. method, program, basis set, keyword set, driver combinations
         `None` is a wildcard selector. To search for `None`, use `"None"`.
 
         Parameters
         ----------
-        dftd3: bool, optional
-            Include dftd3 program record specifications in addition to composite DFT-D3 record specifications
         pretty: bool
             Replace NaN with "None" in returned DataFrame
         **search : Dict[str, Optional[str]]
@@ -279,11 +299,29 @@ class Dataset(Collection):
             Record specifications matching **search.
 
         """
+        ret = self._list_records(dftd3=dftd3)
+        ret = self._filter_records(ret, **search)
+        if pretty:
+            ret.fillna("None", inplace=True)
+        return ret
 
+    def _list_records(self, dftd3: bool = False) -> pd.DataFrame:
+        """
+        Lists specifications of available records, i.e. method, program, basis set, keyword set, driver combinations
+        `None` is a wildcard selector. To search for `None`, use `"None"`.
+
+        Parameters
+        ----------
+        dftd3: bool, optional
+            Include dftd3 program record specifications in addition to composite DFT-D3 record specifications
+
+        Returns
+        -------
+        DataFrame
+            Record specifications matching **search.
+
+        """
         show_dftd3 = dftd3
-
-        if not (search.keys() <= set(self.data.history_keys) | {"name"}):
-            raise KeyError("Not all query keys were understood.")
 
         history = pd.DataFrame(list(self.data.history), columns=self.data.history_keys)
 
@@ -291,7 +329,6 @@ class Dataset(Collection):
         if history.shape[0] == 0:
             ret = history.copy()
             ret['name'] = None
-            ret.set_index(list(self.data.history_keys[:-1]), inplace=True)
             return ret
 
         # Build out -D3 combos
@@ -316,29 +353,9 @@ class Dataset(Collection):
                                                                  keywords=row["keywords"],
                                                                  stoich=row.get("stoichiometry", None),
                                                                  driver=row["driver"]), axis=1)
-        # Find the returned subset
-        for key, value in search.items():
-            if value is None:
-                continue
-            elif value is "None":
-                ret = ret[ret[key].isna()]
-            elif isinstance(value, str):
-                value = value.lower()
-                ret = ret[ret[key].str.lower() == value]
-            elif isinstance(value, (list, tuple)):
-                query = [x.lower() for x in value]
-                ret = ret[ret[key].str.lower().isin(query)]
-            else:
-                raise TypeError(f"Search type {type(value)} not understood.")
-
         if show_dftd3 is False:
             ret = ret[ret["program"] != "dftd3"]
 
-        if pretty:
-            ret.fillna("None", inplace=True)
-
-        ret.set_index(list(self.data.history_keys[:-1]), inplace=True)
-        ret.sort_index(inplace=True)
         return ret
 
     def get_values(self,
@@ -451,7 +468,6 @@ class Dataset(Collection):
             return pd.DataFrame(columns=['index']).set_index('index')
 
         queries = self._form_queries(method=method, basis=basis, keywords=keywords, program=program, name=name)
-
         names = []
         for _, query in queries.iterrows():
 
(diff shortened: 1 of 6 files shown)


qcfractal/interface/collections/dataset.py
```diff
@@ -6,6 +6,7 @@
 
 import numpy as np
 import pandas as pd
+import pint
```
dgasmith: We cannot import pint, the 0.5 second load time is too long without lazy loading. You will need to catch this error in a different way.

qcfractal/interface/collections/dataset.py
```diff
@@ -263,56 +338,86 @@ def list_records(self,
...
+            A DataFrame of values with columns corresponding to methods and rows corresponding to molecule entries.
+            Contributed (native=False) columns are marked with "(contributed)" and may include units in square brackets
+            if their units differ in dimensionality from the Dataset's default units.
         """
-
-        queries = self.list_records(**search, dftd3=True, pretty=False).reset_index()
-        if queries.shape[0] > 10:
-            raise TypeError("More than 10 queries formed, please narrow the search.")
-
+        spec = locals()
```
dgasmith: Kind of resisted something as opaque as locals for some time.

qcfractal/interface/collections/dataset.py
```diff
@@ -335,40 +442,63 @@ def get_values(self,
         DataFrame
             A DataFrame of the queried parameters
         """
+        spec = locals()
+        spec.pop("force")
+        spec.pop("self")
+        # TODO: is this somewhere in elemental?
+        au_units = {'energy': 'hartree',
```
dgasmith: Not yet, just assumed for now.

qcfractal/interface/collections/dataset.py
mattwelborn: LGTM also is not a fan of locals. I'll go ahead and spell it out explicitly.

qcfractal/interface/collections/dataset.py
mattwelborn: Okay, I did the following workaround:

```
            try:
                ret[column_name] *= constants.conversion_factor(
                    data.units, self.units)
            except ValueError as e:
                # This is meant to catch pint.errors.DimensionalityError without importing pint, which is too slow
                if e.__class__.__name__ == "DimensionalityError":
                    ret.rename(columns={column_name: f"{column_name} [{data.units}]"}, inplace=True)
                else:
                    raise
```

dgasmith: Nasty, but ok.

qcfractal/interface/collections/dataset.py
```diff
@@ -942,52 +1075,91 @@ def add_contributed_values(self, contrib: ContributedValues, overwrite=False) ->
...
+            try:
+                ret[column_name] *= constants.conversion_factor(
+                    data.units, self.units)
+            except ValueError as e:
+                # This is meant to catch pint.errors.DimensionalityError without importing pint, which is too slow
+                if e.__class__.__name__ == "DimensionalityError":
+                    ret.rename(columns={column_name: f"{column_name} [{data.units}]"}, inplace=True)
+                else:
+                    raise
+        ret.set_index("index", inplace=True)
```
dgasmith: So here we will not change the underlying `ds.df`? In this case it appears we do not add the metadata to `self._column_metadata[name] = query` or similar. This could make the `visualization` function difficult.

qcfractal/interface/statistics.py
```diff
@@ -47,11 +47,10 @@ def mean_unsigned_relative_error(value, bench, **kwargs):
 _return_series = ['ME', 'MUE', 'MURE', 'WMURE']
 
 
-def wrap_statistics(description, df, value, bench, **kwargs):
-
+def wrap_statistics(description, ds, value, bench, **kwargs):
     # Get benchmark
     if isinstance(bench, str):
-        rbench = df[bench]
+        rbench = ds.get_values(name=bench).iloc[:, 0]
```
dgasmith: Shouldn't this be `rbench = ds.get_values(name=bench)[bench]`?

qcfractal/interface/tests/test_dataset.py
```diff
@@ -232,15 +231,15 @@ def test_nbody_rxn(nbody_ds):
...
+               ("gradient", "p1", "m2", "", "")] # yapf: disable
 
     for h in history:
         ds._add_history(driver=h[0], program=h[1], method=h[2], basis=h[3], keywords=h[4])
 
     assert ds.list_records().shape[0] == 5
     assert ds.list_records(program="P1").shape[0] == 4
-    assert ds.list_records(basis=None).shape[0] == 3
-    assert ds.list_records(keywords=None).shape[0] == 1
+    assert ds.list_records(basis="").shape[0] == 3
```
dgasmith: Basis does need to be None for things like PM6/TorchANI/UFF/etc.

qcfractal/tests/test_collections.py
```diff
@@ -110,17 +141,66 @@ def test_gradient_dataset_get_records(gradient_dataset_fixture):
...
+    client, ds = gradient_dataset_fixture
+
+    # List contributed values
+    df = ds.list_values(native=False).reset_index()
+    assert df.shape == (3, 7)
+    assert set(df.columns) == {*ds.data.history_keys, "name", "native"}
+    assert set(map(lambda x: x.lower(), df['name'])) == set(ds.data.contributed_values.keys())
+
+    df = ds.list_values(driver="graDieNt", native=False).reset_index()
+    assert set(map(lambda x: x.lower(), df['name'])) == set(["gradient", "all details"])
```
dgasmith: `{x.lower() for x in df['name']}`, slightly shorter and uses simpler comprehension. Also recommend using literals: `{"gradient", "all details"}`.

qcfractal/tests/test_collections.py
```diff
@@ -132,6 +212,114 @@ def test_gradient_dataset_statistics(gradient_dataset_fixture):
...
+    ds.save()
+
+    yield client, client.get_collection("dataset", "ds_contributed")
+
+
+def test_dataset_contributed_units(contributed_dataset_fixture):
+    _, ds = contributed_dataset_fixture
+
+    assert "[hartree]" not in ds.get_values(name="Fake Energy").columns[0]
+    assert "[hartree/bohr]" in ds.get_values(name="Fake Gradient").columns[0]
```
dgasmith: Interesting, you wouldn't lean towards keeping this in metadata?

mattwelborn: I'm okay with it going in metadata. Is there a way to attach metadata to columns of a Dataframe? Will it show up in a Jupyter notebook?

dgasmith: This is the `self. _column_metadata`. Pandas will be able to hold onto this soonish.

qcfractal/interface/collections/dataset.py
```diff
@@ -169,6 +165,69 @@ def _add_history(self, **history: Dict[str, Optional[str]]) -> None:
...
+            The computational basis (6-31G)
+        keywords : Optional[str], optional
+            The keyword alias
+        program : Optional[str], optional
+            The underlying QC program
+        driver : Optional[str], optional
+            The type of calculation (e.g. energy, gradient, hessian, dipole...)
+        name : Optional[str], optional
+            The canonical name of the data column
+        native: Optional[bool], optional
```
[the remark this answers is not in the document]
mattwelborn: I wanted a name where True is the default, to emphasize that the default operating mode of QCFractal is through native compute, which is why I didn't use `contributed`. `local` and `indigenous` qualify though.

qcfractal/interface/collections/dataset.py
mattwelborn: Added caching to ds.df and column_metadata for contributed values.

qcfractal/interface/tests/test_dataset.py
mattwelborn: `"None"` may be passed to search for fields which are `None`.

mattwelborn:
Thanks for all of the comments @dgasmith! If you're okay with `native`, this is ready to go once I hit it with yapf.

mattwelborn:
Dummy comment attempting to trigger codecov.

qcfractal/interface/collections/dataset.py
```diff
@@ -263,56 +341,83 @@ def list_records(self,
...
+        program : Optional[str], optional
+            The underlying QC program
+        driver : Optional[str], optional
+            The type of calculation (e.g. energy, gradient, hessian, dipole...)
+        name : Optional[str], optional
+            Canonical name of the record. Overrides the above selectors.
+        native: Optional[bool], optional
+            True: only include data computed with QCFractal
+            False: only include data contributed from outside sources
+            None: include both
```
dgasmith: I wish there was a good way to template doc strings, I have so far failed and wasted a great deal of time doing so.

qcfractal/interface/collections/dataset.py
```diff
@@ -263,56 +341,83 @@ def list_records(self,
...
-                query["stoich"] = query.pop("stoichiometry")
-            name = self._canonical_name(**query)
-            if force or (name not in self.df.columns):
-                data = self.get_records(query.pop("method").upper(), projection={"return_result": True}, **query)
-                self.df[name] = data["return_result"] * constants.conversion_factor('hartree', self.units)
-            ret.append(self.df[name])
+        if native in {True, None}:
+            spec_nodriver = spec.copy()
+            driver = spec_nodriver.pop("driver")
+            if driver is not None and driver != self.data.default_driver:
```
dgasmith: Makes sense.

qcfractal/interface/collections/dataset.py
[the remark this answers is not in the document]
dgasmith: Still not sure if `native` is right, can we raise this as an issue to be resolved before 0.10 (Monday)? A little caution here is warranted since this is likely *the* keyword that people will specify a lot for ML.

qcfractal/interface/collections/dataset.py
```diff
@@ -917,22 +1045,31 @@ def get_keywords(self, alias: str, program: str, return_id: bool = False) -> Uni
...
             The ContributedValues to add.
         overwrite : bool, optional
             Overwrites pre-existing values
+
+        Raises
+        ------
+        ValueError:
+            If contrib's indexes do not match those of the dataset's entries
+        KeyError:
+            If values with the same name already exist, and overwrite is False.
```
dgasmith: I usually do not write out `Raises`. Any time I do it feel awkward, seems to be a general policy. Since we have been neglecting them everywhere, perhaps a removal here is in order.

qcfractal/interface/tests/test_dataset.py
```diff
@@ -235,12 +234,12 @@ def test_database_history():
...
+               ("gradient", "p1", "m2", None, None)]  # yapf: disable
 
     for h in history:
         ds._add_history(driver=h[0], program=h[1], method=h[2], basis=h[3], keywords=h[4])
 
     assert ds.list_records().shape[0] == 5
     assert ds.list_records(program="P1").shape[0] == 4
-    assert ds.list_records(basis=None).shape[0] == 3
-    assert ds.list_records(keywords=None).shape[0] == 1
+    assert ds.list_records(basis="None").shape[0] == 3
```
dgasmith: I see, a bit awkward and not a "user first" strategy. Still if someone say `ds.list_records(method="ANI1")` that is rather clear, I think its worth it.

qcfractal/tests/test_collections.py
```diff
@@ -132,6 +212,132 @@ def test_gradient_dataset_statistics(gradient_dataset_fixture):
...
+    ds.save()
+
+    yield client, client.get_collection("dataset", "ds_contributed")
+
+
+def test_dataset_contributed_units(contributed_dataset_fixture):
+    _, ds = contributed_dataset_fixture
+
+    assert ds._column_metadata[ds.get_values(name="Fake Energy").columns[0]]["units"] == "kcal / mol"
+    assert ds._column_metadata[ds.get_values(name="Fake Gradient").columns[0]]["units"] == "hartree/bohr"
```
dgasmith: Careful here since you have `hartree/bohr` and `kcal / mol`, granted doesn't matter or our units system. I wonder if these things should live as `pint` units here rather than strings. A problem for another time, but worth thinking about.

PR #418, opened 2019-09-30
HDF5 views for Dataset and ReactionDataset, supporting list_values and get_values

mattwelborn:
## Description
This WIP PR adds client-side HDF5 views for Dataset and ReactionDataset, supporting the `list_values` and `get_values` methods. There is not a public interface for attaching views to `{Reaction_,}Dataset` since views cannot yet replace the server.

In addition, this PR:
- introduces a h5py dependency
- parameterizes collection test fixtures to run with and without a view
- adds request counting to client for testing purposes

image

## Todos
- [x] Rename `_get_values_from_records` to `_get_native_values`
- [x] Clean up/refactor duplicated code
- [x] Ensure tests pass on Travis
- [x] This PR should not be merged until after the 0.11 release.

## Questions
- [x] How to handle h5py dependency

## Status
- [ ] Changelog updated
- [ ] Ready to go


mattwelborn:
@doaa-altarawy @dgasmith I think this is ready for review. 

@Lnaden could you check that I added the h5py dependency correctly?

qcfractal/interface/collections/__init__.py
```diff
@@ -9,3 +9,4 @@
 from .optimization_dataset import OptimizationDataset
 from .reaction_dataset import ReactionDataset
 from .torsiondrive_dataset import TorsionDriveDataset
+from .dataset_view import HDF5View
```
Lnaden: add the empty line here

mattwelborn:
Thanks for taking a look, @Lnaden.

qcfractal/interface/collections/dataset_view.py
```diff
@@ -0,0 +1,259 @@
...
+            },
+            "dipole": {
+                "dtype": np.dtype("float64"),
+                "shape": (n_records, 3)
+            }
+        }
+
+        def _write_dataset(dataset, column, entry_dset):
+            assert column.shape[1] == 1
+            for i, name in enumerate(entry_dset):
```
dgasmith: Looping through this data seems slow, is there any way to do this in a vectorized way?

mattwelborn: As far as I could tell, the only data which can be written in a vectorized way are fixed length. Unfortunately, most of our large objects (geometries, gradients, hessians, orbitals) will be variable length. For now, I'm not worrying too much about write performance since writes should be rare.

dgasmith: Sounds good!

PR #420, opened 2019-09-30
Protocols

dgasmith:
## Description
Begins to allow protocols to propagate from Fractal to Engine. This first pass only focuses on optimizations, but will include Wavefunction results in the next release.

This PR will not pass until a new version of Elemental is released. This PR also provides updates to the database to ensure the protocol and wavefunction data can be safely propagated.

## Status
- [ ] Changelog updated
- [ ] Ready to go


mattwelborn:
LGTM, but I don't (yet) know much about how the Schemas work.

dgasmith:
Failing due to a QCElemental/QCEngine version pin mismatch. Going to go ahead and merge.

PR #428, opened 2019-10-03
Dataset view get_molecules

mattwelborn:
## Description
This PR implements local-view-backed `{Reaction_,}Dataset.get_molecules`. 

Misc changes:
- Added (~very slow~) QCArchive-backed QM3 and S22 test fixtures. ~While painful~, these find lots of edge cases (redundant history columns, missing data, multiple stoichiometries, ...) ~These tests are very slow. They add about 10 minutes to the testing process. We may wish to create a new mark `very_slow`, and not run them in CI.~
- Added `get_entries` function to get the dataset index and molecule ids. `_molecule_indexer` now uses this function. The entry/molecule index is stored in the HDF5, but must be loaded into memory.

image

## Benchmarking
These benchmarks were run on the QM9 dataset of 133,885 molecules with 8 energy columns.

image
`list_values` is slightly faster without HDF5 (`force=True`) because the history is stored in memory. However, it's super fast in both cases.

image
`get_values` is unsurprisingly wildly faster with the HDF5 view. The server-backed version was killed after 10 minutes.

image
`get_molecules` is faster with the HDF5 view, but not by an extreme amount. This is because the cost is asymptotically dominated by de-serialization of the Molecules. For the same reason, json and msgpack show the same performance.

image
msgpack saves a little space over json. If space is a real concern, the first change that needs to be made is to de-duplicate the geometry in the `molecule/schema` and `molecule/geometry` datasets.

## Todos
- [x] Document `get_entries`
- [x] Document `force` where it was added.

## Questions
- [x] ~Once more, the new tests make CI 10 minutes longer. Is this okay?~

## Status
- [x] Changelog updated - No significant user-facing changes.
- [ ] Ready to go

### DIFF
--- a/qcfractal/interface/collections/dataset_view.py
+++ b/qcfractal/interface/collections/dataset_view.py
@@ -97,9 +97,9 @@ class HDF5View(DatasetView):
         self._entries: pd.DataFrame = None
 
     def list_values(self) -> pd.DataFrame:
-        df = pd.DataFrame()
         with self._read_file() as f:
             history_keys = self._deserialize_field(f.attrs['history_keys'])
+            df = pd.DataFrame(columns=history_keys + ["name", "native"])
             for dataset in f['value'].values():
                 row = {k: self._deserialize_field(dataset.attrs[k]) for k in history_keys}
                 row["name"] = self._deserialize_field(dataset.attrs["name"])
@@ -379,6 +379,8 @@ class HDF5View(DatasetView):
     # Methods for serializing into HDF5 data fields
     @staticmethod
     def _serialize_data(data: Any) -> np.ndarray:
+        # h5py v3 will support bytes,
+        # but for now the workaround is variable-length np unit8
         return np.fromstring(serialize(data, 'msgpack-ext'), dtype='uint8')
 
     @staticmethod
--- a/qcfractal/interface/collections/dataset_view.py
+++ b/qcfractal/interface/collections/dataset_view.py
@@ -5,9 +5,10 @@ import warnings
 from contextlib import contextmanager
 
 from .dataset import Dataset, MoleculeEntry
+from ..models import ObjectId
 from .reaction_dataset import ReactionEntry
 import pathlib
-from typing import Union, List, Tuple, Any
+from typing import Union, List, Tuple, Dict, Any
 import numpy as np
 import pandas as pd
 import h5py
@@ -51,7 +52,7 @@ class DatasetView(abc.ABC):
             A Dataframe with specification of available columns.
         """
     @abc.abstractmethod
-    def get_values(self, queries: List[Tuple[str]]) -> Tuple[pd.DataFrame, List[str]]:
+    def get_values(self, queries: List[Tuple[str]]) -> Tuple[pd.DataFrame, Dict[str, str]]:
         """
         Get value columns.
 
@@ -65,7 +66,7 @@ class DatasetView(abc.ABC):
             A Dataframe whose columns correspond to each query and a list of units for each column.
         """
     @abc.abstractmethod
-    def get_molecules(self, indexes: List['ObjectId']) -> List[Molecule]:
+    def get_molecules(self, indexes: List[Union[ObjectId, int]]) -> List[Molecule]:
         """
         Get a list of molecules using a molecule indexer
 
@@ -122,7 +123,7 @@ class HDF5View(DatasetView):
         # for some reason, pandas makes native a float column
         return df.astype({"native": bool})
 
-    def get_values(self, queries: List[Tuple[str]]) -> Tuple[pd.DataFrame, List[str]]:
+    def get_values(self, queries: List[Tuple[str]]) -> Tuple[pd.DataFrame, Dict[str, str]]:
         units = {}
         with self._read_file() as f:
             ret = pd.DataFrame(index=f["entry/entry"][()])
@@ -157,10 +158,13 @@ class HDF5View(DatasetView):
 
         return ret, units
 
-    def get_molecules(self, indexes: List['ObjectId']) -> List[Molecule]:
+    def get_molecules(self, indexes: List[Union[ObjectId, int]]) -> List[Molecule]:
         with self._read_file() as f:
             mol_schema = f['molecule/schema']
-            ret = [Molecule(**self._deserialize_data(mol_schema[i])) for i in indexes]
+            ret = [
+                Molecule(**self._deserialize_data(mol_schema[int(i) if isinstance(i, ObjectId) else i]))
+                for i in indexes
+            ]
         return ret
 
     def get_entries(self) -> pd.DataFrame:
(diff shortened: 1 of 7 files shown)


qcfractal/interface/collections/dataset_view.py
```diff
@@ -185,10 +231,67 @@ def _write_dataset(dataset, column, entry_dset):
...
+            mol_shape = (len(molecules), )
+            mol_geometry = molecule_group.create_dataset("geometry",
+                                                         shape=mol_shape,
+                                                         dtype=vlen_double_t,
+                                                         **dataset_kwargs)
+            mol_symbols = molecule_group.create_dataset("symbols",
+                                                        shape=mol_shape,
+                                                        dtype=vlen_utf8_t,
+                                                        **dataset_kwargs)
+            mol_schema = molecule_group.create_dataset("schema", shape=mol_shape, dtype=utf8_t, **dataset_kwargs)
```
dgasmith: Do we want to worry about additional fields like charge/multiplicity or leave that for another day?

mattwelborn: No time like the present!

qcfractal/interface/collections/dataset_view.py
```diff
@@ -162,8 +162,8 @@ def get_molecules(self, indexes: List[Union[ObjectId, int]]) -> List[Molecule]:
         with self._read_file() as f:
             mol_schema = f['molecule/schema']
             ret = [
-                Molecule(**self._deserialize_data(mol_schema[int(i) if isinstance(i, ObjectId) else i]))
-                for i in indexes
+                Molecule(**self._deserialize_data(mol_schema[int(i) if isinstance(i, ObjectId) else i]),
```
dgasmith: Hows the new performance? I suspect it may not have changed much since the molecules should carry around a flag that they are validated.

qcfractal/interface/collections/dataset_view.py
```diff
@@ -1,15 +1,19 @@
...
-import json
+import pathlib
 import warnings
 from contextlib import contextmanager
+from typing import Any, Dict, List, Tuple, Union
 
-from .dataset import Dataset
-import pathlib
-from typing import Union, List, Tuple
+import h5py
```
dgasmith: Would it be feasible to make h5py optional in a `qcportal-core` or the like? It doesn't seem to be that way, but perhaps worth checking in on.

qcfractal/tests/test_collections.py
```diff
@@ -732,10 +745,56 @@ def test_compute_reactiondataset_keywords(fractal_compute_server):
...
 
 
+def live_fractal_or_skip():
+    """Ensure Fractal live connection can be made"""
+    import requests
+    try:
+        requests.get('https://api.qcarchive.molssi.org:443', json={}, timeout=5)
+        return ptl.FractalClient()
+    except (requests.exceptions.ConnectionError, ConnectionRefusedError):
+        return pytest.skip("Could not make a connection to central Fractal server")
```
dgasmith: It might be good to add this to `testing.py` as it will be useful in other places.

qcfractal/tests/test_collections.py
```diff
@@ -732,10 +745,56 @@ def test_compute_reactiondataset_keywords(fractal_compute_server):
     assert kw.values["scf_type"] == "df"
 
 
+def live_fractal_or_skip():
+    """Ensure Fractal live connection can be made"""
+    import requests
```
dgasmith: Fine to have requests as a top level import as it is imported into the project anyway.

qcfractal/tests/test_collections.py
```diff
@@ -564,7 +573,11 @@ def test_rectiondataset_dftd3_energies(reactiondataset_dftd3_fixture_fixture):
 def test_rectiondataset_dftd3_molecules(reactiondataset_dftd3_fixture_fixture):
     client, ds = reactiondataset_dftd3_fixture_fixture
 
-    mols = ds.get_molecules()
+    request_made = not ds._use_view(False)
+    ds._clear_cache()
+
+    with monitor_requests(client, "molecule", request_made=request_made):
```
dgasmith: I didn't realize that monitor requests asserted as well. Could we change the name of this to `check_requests_monitor` and add error messages to the asserts so that it say something like "Made requests when none where expected!". Some sort of name/assert change in that category.

dgasmith:
Do you have any idea why the S22/QM3 tests are so slow? We pull down the S22 set for visualization and it takes a few seconds. `get_values` on 133k rows should only take ~20-30 seconds lets I checked as well, something may be wrong here in the timings.

3kb per entry isn't bad, likely half that without the schema molecule? As you get to larger molecules msgpack will asymptotically be ~2.2x smaller than JSON in addition to being much faster. msgpack/json for ~9 atom systems is pretty close together.

qcfractal/interface/collections/dataset_view.py
mattwelborn: 3.6s vs 4.3s for 10,000 molecules, so not a huge deal but still something.

qcfractal/interface/collections/dataset_view.py
mattwelborn: I will revisit this when I get to `.download()` TODO comment added.

mattwelborn:
S22/QM3 tests now only pull b3lyp columns, and take about 10s total. 

I agree with your concern about file size. I'll see what I can figure out about the object sizes inside the hdf5 file.

mattwelborn:
QM9 file size breakdown:
Schema: 235 Mb  
Geometry: 45 Mb
Energy column: 1 Mb each (x8)
Charge: 1 Mb
Multiplicity: 0.5 Mb
Entry names: 2 Mb
Molecule ID: 1 Mb. 

Two things are clear:
1. The Schema is by far the largest data.
2. The total file size (432 MB) is 50% larger than the sum of its parts. 
Both of these issues warrant further investigation, but for now I will say they are beyond the scope of this PR. 

(Worth noting that ANI-I is 32-bit floats, so 14 GB is a better comparison.)

dgasmith:
@Lnaden You have expressed concerns over some fun(ish) HDF5 aspects. Do you have any thoughts here?

Lnaden:
I think we are okay. @mattwelborn and I sat down last Monday and worked through the dependencies, so long as you dont explicitly flag you need the MPI version, this works fine and has a small-ish footprint.

dgasmith:
What about the data organization and *size* of data? @Lnaden?

Lnaden:
The size of the data is not so much of an issue as much as considering the standard views people will access. The chunk sizes for this data in each dimension should be about the size of how we expect users to access the data. E.g. if we expect people to access some dimension in chunks of 50, then that should be our target chunk size (or some factor there of if 50 would be too large.)

Organization is not something I think I can speak well to as I am not the most familiar with how ANI-I is most optimally structured.

dgasmith:
@Lnaden Specifically this item:

> The total file size (432 MB) is 50% larger than the sum of its parts.

PR #429, opened 2019-10-04
OptimizationDataset Protocols

dgasmith:
## Description
Adds protocols to the optimization dataset.

## Status
- [ ] Changelog updated
- [x] Ready to go

### DIFF
--- a/qcfractal/interface/collections/optimization_dataset.py
+++ b/qcfractal/interface/collections/optimization_dataset.py
@@ -5,6 +5,8 @@ from typing import Any, Dict, List, Optional, Set, Union
 
 import pandas as pd
 
+import qcelemental as qcel
+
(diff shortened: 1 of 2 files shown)


qcfractal/interface/collections/optimization_dataset.py
```diff
@@ -69,15 +77,25 @@ def add_specification(self,
...
+        protocols : Optional[Dict[str, Any]], optional
+            Protocols for this specification.
         overwrite : bool, optional
             Overwrite existing specification names
 
+        Returns
+        -------
+        None
+            Description
+
```
mattwelborn: This is an odd bit of documentation to include. At least `Description` should be deleted.

dgasmith: Whoops, auto doc got me.

PR #430, opened 2019-10-07
Portal type hints

mattwelborn:
## Description
This PR seeks to incrementally improve type hints in Portal, especially in modules used in the QCArchive examples. The motivation for this is twofold:

1. Help users to explore the portal functionality introduced in the examples. ("Can I pass a list of methods to `get_values`?") 
2. Find bugs with type checkers. Like the fact that you can totally pass a list of methods to `get_values`.

Non-goal: make Fractal pass mypy. We're nowhere near that.

## Status
- [ ] Changelog updated
- [ ] Ready to go

### DIFF
--- a/qcfractal/interface/client.py
+++ b/qcfractal/interface/client.py
@@ -4,15 +4,21 @@ import json
 import os
 import re
 from collections import defaultdict
-from typing import Any, Dict, List, Optional, Union
+from typing import Any, DefaultDict, Dict, List, Optional, Tuple, Union
 
 import pandas as pd
 import requests
+
 from pydantic import ValidationError
+from qcfractal.interface.models import TaskRecord
 
 from .collections import collection_factory, collections_name_map
-from .models import GridOptimizationInput, Molecule, ObjectId, TorsionDriveInput, build_procedure
-from .models.rest_models import ComputeResponse, rest_model
+from .collections.collection import Collection
+from .models import (GridOptimizationInput, KeywordSet, Molecule, ObjectId, ResultRecord, TorsionDriveInput,
+                     build_procedure)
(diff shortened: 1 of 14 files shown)


qcfractal/interface/client.py
```diff
@@ -4,15 +4,21 @@
...
 
 from .collections import collection_factory, collections_name_map
-from .models import GridOptimizationInput, Molecule, ObjectId, TorsionDriveInput, build_procedure
-from .models.rest_models import ComputeResponse, rest_model
+from .collections.collection import Collection
+from .models import (GridOptimizationInput, KeywordSet, Molecule, ObjectId, ResultRecord, TorsionDriveInput,
+                     build_procedure)
+from .models.rest_models import (ComputeResponse, KeywordGETResponse, MoleculeGETResponse, ProcedureGETResponse,
+                                 QueryObjectId, QueryProjection, QueryStr, ResultGETResponse, ServiceQueueGETResponse,
+                                 TaskQueueGETResponse, rest_model)
```
dgasmith: What about something like: https://github.com/samuelcolvin/pydantic/blob/master/pydantic/main.py#L23?

Also please do `isort` this when you get the chance.

mattwelborn: Was not aware of `TYPE_CHECKING`, it makes a lot of sense, and I will update this PR accordingly.

mattwelborn:
@dgasmith This is ready for review. 

Changing to TYPE_CHECKING for imports leads to an known issue with LGTM (Semmle/ql#2014). Until that's fixed, I've disabled the alert. The TYPE_CHECKING blocks are also marked as `pragma: no cover` (nedbat/coveragepy#831).

PR #433, opened 2019-10-09
Server-side Dataset Views

mattwelborn:
## Description
This PR implements server-side HDF5 views for `portal.collections.Dataset` and `ReactionDataset`. To recap, views currently provide fast implementations of `get_values`, `list_values`, `get_molecules`, and `get_entries`. Two access modes to views are provided.

1. `Dataset.download()` downloads the view and then uses that view locally to handle requests.
2. `RemoteView` sends requests to new REST endpoints on the server. This is the default operating mode if the server indicates that it is supported for a given dataset (using the `view_available` column). These endpoints use server-side HDF5 views to compose responses. Data are sent over the wire as feather-serialized pandas DataFrames.

In service of RemoteView, four new REST GET endpoints have been added:
- `/collection/[collection_id]/view/entry`
- `/collection/[collection_id]/view/list`
- `/collection/[collection_id]/view/value`
- `/collection/[collection_id]/view/molecule`

The construction of views, their storage, and updates to the collections table are a power-user only feature. However, the process is fairly simple (see `docs/qcfractal/server_view.rst`), so we might consider adding a CLI for our own convenience.

A future PR will address subsets in `get_values` and `get_entries`, and remove the requirement that the entries table be stored in memory client side. (It can be many GB.)

## Other changes
This PR changes `get_values` to chunk queries to views into one request.

This PR introduces a dependency on `pyarrow`, the python binding to the C++ implementation of Apache Arrow. This PR introduces a testing dependency on `requests-mock` for testing `Dataset.download()`.

This PR adds a new REST GET endpoint at `/collection/[collection_id]`. This endpoint takes the same REST model as GET `/collection`, but additionally searches the collection table on `collection_id`.

@doaa-altarawy This PR changes the how REST models are registered. `register_model` now takes a regex rather than a string and `rest_model` calls match on that regex. This change was made to accommodate the new endpoints (regex e.g.: `"collection/[0-9]+/view/entry"`).

## ANI-Milestone Roadmap
image

## Todos
- [x] Checksum blob

## Status
- [x] Changelog updated
- [x] Ready to go


qcfractal/interface/collections/collection.py
```diff
@@ -242,7 +242,9 @@ def save(self, client: Optional['FractalClient'] = None) -> 'ObjectId':
                 raise KeyError(f"Error adding collection: \n{response.meta.error_description}")
             self.data.__dict__["id"] = response.data
         else:
-            client.add_collection(self.data.dict(), overwrite=True)
+            response = client.add_collection(self.data.dict(), overwrite=True, full_return=True)
+            if response.meta.success is False:
+                raise KeyError(f"Error updating collection: \n{response.meta.error_description}")
```
dgasmith: Thanks this has been needed.

qcfractal/interface/collections/dataset.py
```diff
@@ -104,6 +107,40 @@ class DataModel(Collection.DataModel):
         history: Set[Tuple[str, str, str, Optional[str], Optional[str]]] = set()
         history_keys: Tuple[str, str, str, str, str] = ("driver", "program", "method", "basis", "keywords")
 
+    def set_view(self, path: Union[str, Path]) -> None:
+        from . import HDF5View
+        self._view = HDF5View(path)
+
+    def download(self, local_path: Optional[Union[str, Path]] = None, verify: bool = True) -> None:
+        """ Download a remote view """
+        import hashlib
```
dgasmith: Native lib, can be imported at the top.

qcfractal/interface/collections/dataset.py
```diff
@@ -104,6 +107,40 @@ class DataModel(Collection.DataModel):
...
+            self._view_tempfile = tempfile.NamedTemporaryFile()  # keep temp file alive until self is destroyed
+            local_path = self._view_tempfile.name
+
+        r = requests.get(self.data.view_url, stream=True)
+        with open(local_path, 'wb') as fd:
+            for chunk in r.iter_content(chunk_size=8192):
+                fd.write(chunk)
+
+        if verify:
+            remote_checksum = requests.get(self.data.view_url + ".blake2b").text.rstrip()
```
dgasmith: Could this be a metadata blob rather than a single file? Might help us in the future.

qcfractal/interface/collections/dataset_view.py
```diff
@@ -3,31 +3,29 @@
 import pathlib
 import warnings
 from contextlib import contextmanager
-from typing import Any, Dict, Iterator, List, Tuple, Union
+from typing import TYPE_CHECKING, Any, Dict, Iterator, List, Tuple, Union
```
dgasmith: I don't think we need to import this or I am very wrong here?

qcfractal/interface/models/rest_models.py
```diff
@@ -891,7 +902,7 @@ class ListMoleculeResponse(ProtoModel):
...
 
 
-register_custom_model("optimization", "final_result", OptimizationFinalResultBody, ResultResponse)
-register_custom_model("optimization", "all_results", OptimizationAllResultBody, ListResultResponse)
-register_custom_model("optimization", "initial_molecule", OptimizationAllResultBody, ListMoleculeResponse)
-register_custom_model("optimization", "final_molecule", OptimizationAllResultBody, ListMoleculeResponse)
+register_model(r"optimization/final_result", "GET", OptimizationFinalResultBody, ResultResponse)
+register_model(r"optimization/all_results", "GET", OptimizationAllResultBody, ListResultResponse)
+register_model(r"optimization/initial_molecule", "GET", OptimizationAllResultBody, ListMoleculeResponse)
+register_model(r"optimization/final_molecule", "GET", OptimizationAllResultBody, ListMoleculeResponse)
```
dgasmith: @doaa-altarawy Can you take a look here. Seems fine, but best to be safe.

qcfractal/storage_sockets/view.py
```diff
@@ -0,0 +1,122 @@
...
+        path: Union[str, Path]
+            Directory containing dataset views
+        """
+        self._path = pathlib.Path(path)
+        if not self._path.is_dir():
+            raise ValueError(f"Path in ViewHandler must be a directory, got: {self._path}")
+
+    def view_path(self, collection_id: int) -> pathlib.Path:
+        """
+        Get path of a view
```
dgasmith: Small comment: we usually try to have these as a complete sentence with a period.

qcfractal/interface/collections/dataset.py
mattwelborn: I don't completely understand the proposal. Is the idea to put the checksum in the collections table? I think that would be a good idea, if that's what you mean.

qcfractal/interface/collections/dataset_view.py
mattwelborn: `TYPE_CHECKING` is in the namespace of the `typing` module (https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING). Or maybe you mean that there was no `if TYPE_CHECKING:` block in this file when this comment was made.

dgasmith: Yup, I am just [foolish](https://github.com/samuelcolvin/pydantic/blob/master/pydantic/main.py#L10).

qcfractal/interface/collections/dataset.py
dgasmith: If I understand this correctly the `.blake2b` is a text file at this location. It might be more flexible for this to be a JSON-serialized dictionary so that we can have arbitrary fields.

mattwelborn: You understand correctly, and I'm happy to make a metadata blob instead. That said, what do you think about keeping that metadata blob in a column in the collections table rather than in a file on the http file server?

dgasmith: It's more of a future proofing item. If the JSON blob just has a `blake2b` field then its fine, but what if we need something else? It would require new architecture then.

qcfractal/interface/models/rest_models.py
```diff
@@ -46,49 +40,52 @@ def _register_model(is_custom_queries: bool, name: str, rest: str, body: ProtoMo
...
-        if resource.lower() in __custom_rest_models:
-            return __custom_rest_models[resource.lower()][subresource.upper()]
-        else:
-            return __rest_models[resource.lower()][rest.upper()]
-    except KeyError:
-        sub = subresource.lower() if subresource else ''
-        raise KeyError(f"REST Model {rest.upper()} {resource.lower()}:{sub} could not be found.")
+    rest = rest.upper()
+    matches = []
+    for model_re in __rest_models.keys():
```
dgasmith: Considering we can get up to 100k REST requests a day right now, does this slow us down appreciably? Might be best to add a `lru_cache`.

qcfractal/tests/test_collections.py
```diff
@@ -475,7 +479,7 @@ def test_reactiondataset_check_state(fractal_compute_server):
     assert bench.loc["He2"][0] == contrib["values"]["He2"]
 
 
-@pytest.fixture(scope="module", params=[True, False], ids=["with view", "without view"])
+@pytest.fixture(scope="module", params=["download view", "no view", "remote view"])
```
dgasmith: Might want to consider `download_view` here to keep it a bit pythonic.

qcfractal/tests/test_collections.py
```diff
@@ -969,6 +970,68 @@ def test_s22_view_identical(s22_fixture):
...
+    with requests_mock.Mocker(real_http=True) as m:
+        path = pathlib.Path(tmp_path_factory.mktemp('test_collections'), 'ds_gradient_remote.hdf5')
+        view = ptl.collections.HDF5View(path)
+        view.write(ds)
+
+        fake_url = "https://qcarchiveviews.com/gradient_ds.h5"
+        ds.data.__dict__["view_url"] = fake_url
+        assert ds.data.id == ds.save()
+
+        m.get(fake_url, body=open(path, 'rb'))
```
dgasmith: Is the straight open here safe?

devtools/conda-envs/adapters.yaml
```diff
@@ -20,11 +20,14 @@ dependencies:
   - alembic
   - tqdm
   - h5py
+  - pyarrow>=0.13.0
+  - double-conversion>=3.0.0
```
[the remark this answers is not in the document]
mattwelborn: Pyarrow is used in qcfractal/interface/collections/dataset_view.py explicitly and in qcfractal/storage/view.py implicitly. double-conversion is a pyarrow dependency that does not install correctly in some Travis pipelines. This is probably pyarrow's fault, but it's hard to track down. (Maybe @Lnaden could suggest a strategy for debugging this.)

We could consider making these dependencies optional in addition to the h5py dependency, and just raise if views are attempted and they aren't installed.

devtools/conda-envs/adapters.yaml
```diff
@@ -20,11 +20,14 @@ dependencies:
...
 
 # Test depends
   - pytest
   - pytest-cov
   - codecov
+  - requests-mock
```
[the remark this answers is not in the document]
mattwelborn: requests-mock is used in test_collections to mock file downloads in the dataset fixtures. It depends only on requests and six. Let me know if you feel strongly about it.

qcfractal/tests/test_collections.py
mattwelborn: The resolved issues I can think of are:
- We wrote the file at path, so we can trust it.
- The file at path is not open elsewhere.
- The request is mocked, so we don't have to worry about MitM or other network security concerns. 

Let me know if you have a different concern.

qcfractal/config.py
```diff
@@ -75,6 +75,13 @@ class Config(SettingsCommonConfig):
         pass
 
 
+class ViewSettings(ConfigSettings):
+    """
+    HDF5 view settings
+    """
+    directory: str = Schema(None, description="Location of frozen-view data.")
```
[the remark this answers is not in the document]
mattwelborn: Makes sense. Now defaults to `base_folder/view`. I wanted views to be off by default, so I added an `enable: bool = False` config option.

qcfractal/config.py
```diff
@@ -194,6 +202,13 @@ def database_uri(self, safe=True, database=None):
 
         return uri
 
+    @property
+    def view_path(self):
+        if self.view.directory is None:
+            return None
```
doaa-altarawy: What is the use case when the view directory is None?

mattwelborn: Previously, `None` meant disable views. I've added an `enable` flag in addition to the view path to make this more clear.

qcfractal/interface/models/rest_models.py
```diff
@@ -408,6 +407,95 @@ class CollectionPOSTResponse(ProtoModel):
...
+    meta: EmptyMeta = Schema(EmptyMeta(), description=common_docs[EmptyMeta])
+    data: Data = Schema(..., description="Information about which entries to return.")
+
+
+class CollectionViewEntryGETResponse(ProtoModel):
+    meta: CollectionViewGETResponseMeta = Schema(..., description=str(get_base_docs(CollectionViewGETResponseMeta)))
+    data: Optional[bytes] = Schema(..., description="Feather-serialized bytes representing a pandas DataFrame.")
+
+
+register_model("collection/[0-9]+/view/entry", "GET", CollectionViewEntryGETBody, CollectionViewEntryGETResponse)
```
doaa-altarawy: I don't think there is a need for the `/view/` in all urls

qcfractal/storage_sockets/sqlalchemy_socket.py
```diff
@@ -851,23 +849,28 @@ def add_collection(self, data: Dict[str, Any], overwrite: bool = False):
...
-                        name: str = None,
-                        return_json: bool = True,
-                        with_ids: bool = True,
-                        limit: int = None,
-                        projection: Dict[str, Any] = None,
+                        collection: Optional[str] = None,
+                        name: Optional[str] = None,
+                        col_id: Optional[int] = None,
+                        limit: Optional[int] = None,
+                        projection: Optional[Dict[str, Any]] = None,
```
doaa-altarawy: This can cause a big problem if it's called with no parameters, `get_collection()`. The use of no filters at all should raise an exception. No `get_` method should allow getting all the DB unintentionally.
However, `get_collection(collection=collection_type)` is okay.

qcfractal/interface/models/rest_models.py
```diff
@@ -12,32 +15,24 @@
...
-def register_model(resource: str, rest: str, body: ProtoModel, response: ProtoModel) -> None:
-    _register_model(False, resource, rest, body, response)
-
-
-def register_custom_model(resource: str, rest: str, body: ProtoModel, response: ProtoModel) -> None:
-    _register_model(True, resource, rest, body, response)
-
-
-def _register_model(is_custom_queries: bool, name: str, rest: str, body: ProtoModel, response: ProtoModel) -> None:
+def register_model(name: str, rest: str, body: ProtoModel, response: ProtoModel) -> None:
```
doaa-altarawy: That's better :)

devtools/conda-envs/adapters.yaml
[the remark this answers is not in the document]
dgasmith: It is is certainly worth thinking about, QCFractal depends are not too big of an issue. I do worry a bit about portal however trying to make sure it doesn't become a monolith. The flip side is we become hugely parametrizable for what is really core functionality.

Any other comments here are welcome, something we might want to run through Slack too.

qcfractal/config.py
```diff
@@ -75,6 +75,14 @@ class Config(SettingsCommonConfig):
         pass
 
 
+class ViewSettings(ConfigSettings):
+    """
+    HDF5 view settings
+    """
+    enable: bool = Schema(False, description="Enable frozen-views.")
+    directory: str = Schema(None, description="Location of frozen-view data. If None, defaults to base_folder/view.")
```
dgasmith: `base_folder/views` (plural)?

qcfractal/interface/models/rest_models.py
```diff
@@ -46,49 +41,53 @@ def _register_model(is_custom_queries: bool, name: str, rest: str, body: ProtoMo
...
+    if name not in __rest_models:
+        __rest_models[name] = {}
 
-    models_dict[name][rest] = (body, response)
+    __rest_models[name][rest] = (body, response)
 
 
-def rest_model(resource: str, rest: str, subresource: str = None) -> Tuple[ProtoModel, ProtoModel]:
-    """Aquires a REST Model
+@functools.lru_cache(1000, typed=True)
```
dgasmith: Thanks, unlikely to really be needed, but it is a harmless change.

qcfractal/storage_sockets/sqlalchemy_socket.py
dgasmith: We do need the ability to call `get_collections` without a query, but with a projection which is how the client side `list_collections` work. IIRC `get_molecules` and the like are uncapped, up for changing this where we add an explicit `list_collections` storage socket function as well.

qcfractal/storage_sockets/view.py
```diff
@@ -0,0 +1,134 @@
...
+            Dictionary corresponding to requested REST model
+        """
+        meta = {"errors": [], "success": False, "error_description": False, "msgpacked_cols": []}
+
+        if not self.view_exists(collection_id):
+            meta["success"] = False
+            meta["error_description"] = f"View not available for collection #{collection_id}"
+            return {"meta": meta, "data": None}
+
+        view = HDF5View(self.view_path(collection_id))
```
dgasmith: Is this pretty fast or should it be opened init?

qcfractal/tests/test_collections.py
dgasmith: I mean to make sure we close the file handle. `with open(...) as handle: `. In a test so not worried, but likely worth a comment of why we are not doing that perhaps.

qcfractal/alembic/versions/a2f76bb7be65_updates_collections_table_with_owner_.py
```diff
@@ -21,6 +21,7 @@ def upgrade():
     op.add_column('collection', sa.Column('description', sa.String(), nullable=True))
     op.add_column('collection', sa.Column('owner', sa.String(length=100), nullable=True))
     op.add_column('collection', sa.Column('view_url', sa.String(), nullable=True))
+    op.add_column('collection', sa.Column('view_metadata', sa.JSON(), nullable=True))
```
mattwelborn: @dgasmith Could you please spot-review this two-line change to the migration?

qcfractal/storage_sockets/view.py
mattwelborn: Should be very fast. The HDF5view constructor just sets `self._path = path`. The get methods in HDF5view open and close the HDF5 file every time they are called, which I benchmarked at 100 µs on my laptop. I believe this is the intended use of HDF5 (rather than keeping the file open throughout a session).

mattwelborn:
"I mean to make sure we close the file handle. with open(...) as handle: . In a test so not worried, but likely worth a comment of why we are not doing that perhaps."

Makes total sense. Fixed.

qcfractal/storage_sockets/sqlalchemy_socket.py
mattwelborn: I'm going to make an issue and kick this down the road.

qcfractal/interface/collections/dataset.py
mattwelborn: Now a collections table column

qcfractal/alembic/versions/a2f76bb7be65_updates_collections_table_with_owner_.py
dgasmith: Looks good.

mattwelborn:
Thanks for the reviews @dgasmith @doaa-altarawy !

PR #438, opened 2019-10-14
add thread and ternary plots

loriab:
## Description
**DON'T MERGE** as this is only rewrites of matplotlib routines that function on data. Presently *no* integration with QCA.

## Todos
  - [x] plotly threads plot
  - [x] plotly ternary plot
  - [ ] plotly threads plot of qcf data
  - [ ] plotly ternary plot of qcf data
  - [ ] find a decent RGB colormap in plotly -- their "Jet" is very dark so that red/blue are hard to distinguish

Screen Shot 2019-10-14 at 11 09 36 AM
Screen Shot 2019-10-14 at 11 08 52 AM

## Questions
- [ ] Using the plotly built-in ternary would be nice, but I don't see plotting two at once or turning one upside down. If anyone sees a better path, let me know.

## Status
- [ ] Changelog updated
- [ ] Ready to go

### DIFF
--- a/qcfractal/interface/visualization.py
+++ b/qcfractal/interface/visualization.py
@@ -229,3 +229,280 @@ def scatter_plot(traces: List[Dict[str, Any]],
     figure = go.Figure(data=data, layout=layout)
 
     return _configure_return(figure, "qcportal-violin", return_figure)
+
+
+def threads(data, labels, color=None, title='', xlimit=4.0, mae=None, mape=None,
+    mousetext=None, mouselink=None, mouseimag=None, mousetitle=None, mousediv=None,
+    labeled=True, view=True,
+    saveas=None, relpath=False, graphicsformat=['pdf']):
+    """Generates a tiered slat diagram between model chemistries with
+    errors (or simply values) in list *data*, which is supplied as part of the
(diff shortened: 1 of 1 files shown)


dgasmith:
Cool! This looks great. 

I agree that we need to hand code the ternaries from Plotly basics rather than their built in function.

You may want to turn off the grid lines, [like so.](https://plot.ly/python/axes/#toggle-grid-and-zerolines-with-update-axis-methods).

qcfractal/interface/visualization.py
```diff
@@ -229,3 +229,280 @@ def scatter_plot(traces: List[Dict[str, Any]],
     figure = go.Figure(data=data, layout=layout)
 
     return _configure_return(figure, "qcportal-violin", return_figure)
+
+
+def threads(data, labels, color=None, title='', xlimit=4.0, mae=None, mape=None,
```
dgasmith: We usually use `mue/mure` over `mae/mape`.

loriab:
turn off the blue bkgd + white grid lines for both ternary and threads or just the former?

dgasmith:
For ternary for sure. For the threads you may want to consider shifting the chart +/- 0.5 to have the grid correctly line up, may with a view.

PR #440, opened 2019-10-17
Wavefunction Computation & Storage

### LINKED ISSUE #117: Electrostatic potential and electric field on a grid
For our Open Force Field collaboration, we would like for calculations to be able to output electrostatic potential (ESP) and electric field (EF) values on a grid.  

There are a few standard ways of generating grids for evaluating ESPs, such as Merz-Singh-Kollman grids, which are standard in the AMBER procedure, and Christopher Bayly's FCC grids.  The [respyte package](https://github.com/lpwgroup/respyte), which @hyejang is developing in my group, should contain methods for easily generating these grids for a provided molecular structure (i.e. atomic symbols and xyz coordinates).  Our current idea is for QCArchive to interface with `respyte` to generate the grid points, then call the QC code to evaluate the ESPs and electric fields.  

The potential applications of this feature would be the development of a wide range of electrostatic models, including point-charge, multipole, and polarizable.  To support all of these, we should be able to compute ESPs and EFs for any structure (including optimized, constrained-optimized and un-optimized.)  For developing polarizable models, it would be highly helpful if we could place external point charges near the molecule.  The external point charge would polarize the electron density which would affect the ESP/EF.  However, I think the output ESP/EF should not contain the "direct" contribution from the external point charge, as it may be very close to the grid points and result in very large values. 

Note: QCArchive would not be responsible for producing the charge model (i.e. running RESP), as we would do that locally.  Once we have decided on a good charge model and a more solid procedure, we could talk about having QCArchive carry out the RESP fitting in a separate feature.

dgasmith:
## Description
Adds the ability to evaluate wavefunctions via protocols and store them in the database for further use. This also moves the `FractalClient.add_compute` function to a more kwarg centric evaluator.

This requires Psi4 1.4 to be released and QCEngine 0.12.0 to be released before it can work with the current testing infrastructure. Tests should be padded in such a way that this isn't an issue however.

Closes #117 as we will store the fundamental data to evaluate ESPs at a later date.

## Status
- [ ] Changelog updated
- [x] Ready to go


qcfractal/alembic/versions/1bd2f7711e68_wavefunction_table.py
```diff
@@ -0,0 +1,53 @@
...
+import qcfractal
+
+
+# revision identifiers, used by Alembic.
+revision = '1bd2f7711e68'
+down_revision = 'a2f76bb7be65'
+branch_labels = None
+depends_on = None
+
+
```
mattwelborn: Doesn't need to be in this PR, but we should get localized orbitals in. They're useful for education!

qcfractal/interface/client.py
```diff
@@ -732,17 +734,17 @@ def add_compute(self,
...
             The KeywordSet ObjectId to use with the given compute
-        molecule : Union[str, Molecule, List[Union[str, Molecule]]]
+        molecule : Union['ObjectId', 'Molecule', List[Union['ObjectId', 'Molecule']]]
             The Molecules or Molecule ObjectId's to compute with the above methods
-        priority : str, optional
+        priority : Optional[str], optional
             The priority of the job {"HIGH", "MEDIUM", "LOW"}. Default is "MEDIUM".
-        tag : str, optional
+        tag : Optional[str], optional
             The computational tag to add to your compute, managers can optionally only pull
```
mattwelborn: Please add protocols to docstring

qcfractal/interface/models/records.py
```diff
@@ -284,6 +287,58 @@ def check_method(cls, v):
     def check_basis(cls, v):
         return prepare_basis(v)
 
+    def get_wavefunction(self, key: Union[str, List[str]]) -> Any:
+        """
+        Pulls down the Wavefunction data associated with the copmutation.
```
mattwelborn: typo

qcfractal/interface/models/rest_models.py
```diff
@@ -575,6 +575,25 @@ def ensure_list_of_dict(cls, v):
...
+    data: Data = Schema(
+        ..., description="Data of the Wavefunction Get field: consists of a ObjectId of the Wavefunction object to fetch.")
+
+
+class WavefunctionStoreGETResponse(ProtoModel):
+    meta: ResponseGETMeta = Schema(..., description=common_docs[ResponseGETMeta])
+    data: Dict[str, Any] = Schema(..., description="The entries of the Wavefunction object requested.")
+
+
+register_model("wavefunctionstore", "GET", WavefunctionStoreGETBody, WavefunctionStoreGETResponse)
```
mattwelborn: another nitpick: consider naming the endpoint "wavefunction" to match the names of the other models ("molecule", "result", etc.)

qcfractal/procedures/procedures.py
```diff
@@ -159,12 +168,35 @@ def parse_output(self, result_outputs):
...
             rdata = data["result"]
-            stdout, stderr, error = self.storage.add_kvstore([rdata["stdout"], rdata["stderr"],
-                                                              rdata["error"]])["data"]
+            stdout, stderr, error = self.storage.add_kvstore([rdata["stdout"], rdata["stderr"], rdata["error"]])["data"]
             rdata["stdout"] = stdout
             rdata["stderr"] = stderr
             rdata["error"] = error
 
+            # Store Wavefunction data
+            if data["result"].get("wavefunction", False):
```
mattwelborn: In 2021, we can use the walrus operator here!

mattwelborn:
Looks great to me. I'm very excited to see wavefunction data in the database. My only gripe is that `WavefunctionStore` doesn't match the naming convention of other tables, models, and endpoints (molecule, keywords, collection, ...).

qcfractal/interface/models/rest_models.py
dgasmith: This seems much more analogous to the KVStore than the `result` or `molecule`. This is why the API and REST endpoints match other stores. What do you think?

mattwelborn: I didn't fully appreciate that, and I am okay with WavefunctionStore now.

PR #442, opened 2019-10-18
Entries and values subsets

mattwelborn:
## Description
This PR implements subsets in `get_values` and `get_entries` for `Dataset` and `ReactionDataset`. It also defers downloading of entries (`Dataset.data.records`) and contributed values until they are requested. 

After discussion with @doaa-altarawy, the syntax for projections has been changed/corrected. All projections are affirmative lists. For example, `projection = {a: True, b: True}` is now `projection = ["a", "b"]`. `projection = {k: False}` is no longer supported.

Because of substantial changes to the collections table, tests involving the QCA server will fail. I have created a local staging server and verified that these tests pass. This PR is therefore fully [WOMM certified](https://blog.codinghorror.com/the-works-on-my-machine-certification-program/).
 
## Roadmap
image

## Todos
- [x] Docstrings

## Status
- [x] Changelog updated
- [x] Ready to go


qcfractal/storage_sockets/models/sql_base.py
```diff
@@ -49,7 +48,7 @@ def to_dict(self, exclude=None):
         class_inspector = inspect(self.__class__)
         # add hybrid properties
         for key, prop in class_inspector.all_orm_descriptors.items():
-            if isinstance(prop, hybrid_property):
+            if isinstance(prop, hybrid_property) and (key not in tobe_deleted_keys):
```
doaa-altarawy: Revert back, this condition is needed.

mattwelborn: Thanks for picking this up!

qcfractal/storage_sockets/sqlalchemy_socket.py
```diff
@@ -853,8 +853,9 @@ def get_collections(self,
                         name: Optional[str] = None,
                         col_id: Optional[int] = None,
                         limit: Optional[int] = None,
-                        projection: Optional[Dict[str, Any]] = None,
-                        skip: int = 0) -> Dict[str, Any]:
+                        projection: Optional[List[str]] = None,
+                        skip: int = 0,
+                        heavy: bool = True) -> Dict[str, Any]:
```
doaa-altarawy: Add docs what is heavy means plz.

qcfractal/tests/test_collections.py
```diff
@@ -27,6 +28,62 @@ def check_requests_monitor(client, request, request_made=True, kind="get"):
         assert after == before, f"Requests were made when none were expected. Request type: {request} ({kind})"
 
 
+def df_equals(df1, df2, sort=False):
```
dgasmith: Perhaps should go into `testing` or even elemental testing at some point.

qcfractal/interface/client.py
```diff
@@ -500,7 +500,8 @@ def list_collections(self, collection_type: Optional[str] = None, aslist: bool =
             df.sort_index(inplace=True)
             return df
 
-    def get_collection(self, collection_type: str, name: str, full_return: bool = False) -> 'Collection':
+    def get_collection(self, collection_type: str, name: str, full_return: bool = False,
+                       heavy: bool = None) -> 'Collection':
```
dgasmith: Maybe `complete` or `full`?

qcfractal/interface/collections/collection.py
```diff
@@ -172,7 +172,11 @@ def from_json(cls, data: Dict[str, Any], client: 'FractalClient' = None) -> 'Col
 
         name = data.pop('name')
         # Allow PyDantic to handle type validation
-        return cls(name, client=client, **data)
+        ret = cls(name, client=client, **data)
+        for optional_field in ["records", "contributed_values"]:
+            if optional_field in ret.data.fields and optional_field not in data:
+                ret.data.__dict__[optional_field] = None
```
dgasmith: Somewhat surprised this isn't handled by Pydantic initialization. A bit dangerous to have this so low as `["records", "contributed_values"]` are specific to the Dataset subclasses and not generic Collections.

qcfractal/interface/collections/dataset.py
```diff
@@ -1224,19 +1265,35 @@ def _list_contributed_values(self) -> pd.DataFrame:
...
-    def _get_contributed_values(self, force: bool = False, **spec) -> pd.DataFrame:
+    def _subset_in_cache(self, column_name: str, subset: Set[str]) -> bool:
+        try:
+            return not self.df.loc[subset, column_name].isna().any()
+        except (KeyError):
+            return False
+
+    def _update_cache(self, new_data: pd.DataFrame) -> None:
+        new_df = pd.DataFrame(index=set(self.df.index) | set(new_data.index),
+                              columns=set(self.df.columns) | set(new_data.columns))
```
dgasmith: Always a fan of set logic.

qcfractal/storage_sockets/sqlalchemy_socket.py
```diff
@@ -887,7 +888,15 @@ def get_collections(self,
         collection_class = get_collection_class(collection)
         query = format_query(collection_class, lname=name, collection=collection, id=col_id)
 
-        # try:
+        if not heavy and projection is None:
+            projection = collection_class.col()
+            for item in [
+                    "contributed_values", "records", "records_obj", "update_relations", "lname", "extra",
```
doaa-altarawy: `lname` is already excluded in all queries.

This doesn't look general enough, it seems customized to one use case that doesn't need extra nor records nor contributed_values. Also, and most importantly, records are specific to Datasets not all kinds of collections.
I'll push some code to the PR.

qcfractal/tests/test_storage.py
```diff
@@ -515,14 +560,9 @@ def test_results_get_dual(storage_results):
...
     """See new changes in design here"""
 
-    ret = storage_results.get_results(method="M2", program="P2", projection={"return_result", "id"})["data"][0]
-    assert set(ret.keys()) == {"id", "return_result"}
-    assert ret["return_result"] == 15
-
-    # Note: explicitly set with_ids=False to remove ids
-    ret = storage_results.get_results(method="M2", program="P2", with_ids=False,
-                                      projection={"return_result"})["data"][0]
-    assert set(ret.keys()) == {"return_result"}
```
doaa-altarawy: Why is this test removed?

doaa-altarawy:
I couldn't get `test_visualization` and `test_collections` to pass. What should I do to make it work?

mattwelborn:
@doaa-altarawy `test_visualization` and `test_collections` contain tests that depend on the QCA server. They work on my local staging server.

qcfractal/tests/test_storage.py
mattwelborn: oops. Fixed.

qcfractal/interface/collections/collection.py
dgasmith: On second thought can we move this to the `Dataset` or `ReactionDataset` class?

qcfractal/interface/collections/dataset.py
```diff
@@ -163,16 +161,35 @@ def download(self, local_path: Optional[Union[str, Path]] = None, verify: bool =
...
-    def _form_index(self) -> None:
-        self._entry_index = pd.DataFrame([[entry.name, entry.molecule_id] for entry in self.data.records],
-                                         columns=["name", "molecule_id"])
+    def _get_data_records_from_db(self):
+        self._check_client()
+        response = self.client.get_collection(self.__class__.__name__.lower(),
+                                              self.name,
+                                              full_return=False,
+                                              heavy=True)
+        self.data.__dict__["records"] = response.data.records
```
dgasmith: Can we add a note that this only works because `get_collection` builds a validated Dataset object and so `response.data.records` is correctly coerced to a Pydantic object.

qcfractal/interface/collections/dataset.py
```diff
@@ -1217,19 +1267,36 @@ def _list_contributed_values(self) -> pd.DataFrame:
 
         return ret
 
-    def _get_contributed_values(self, force: bool = False, **spec) -> pd.DataFrame:
-        queries = self._filter_records(self._list_contributed_values().rename(columns={'stoichiometry': 'stoich'}),
-                                       **spec)
+    def _subset_in_cache(self, column_name: str, subset: Set[str]) -> bool:
+        try:
+            return not self.df.loc[subset, column_name].isna().any()
+        except (KeyError):
```
dgasmith: Doesn't need `()`.

qcfractal/testing.py
```diff
@@ -533,10 +536,73 @@ def sqlalchemy_socket_fixture(request, postgres_server):
...
+        print("Failed to connect to localhost")
+        try:
+            return pytest.xfail("Until QCA is migrated to v0.12.0")
+            requests.get('https://api.qcarchive.molssi.org:443', json={}, timeout=5)
+            return FractalClient()
+        except (requests.exceptions.ConnectionError, ConnectionRefusedError):
+            return pytest.skip("Could not make a connection to central Fractal server")
+
+
+def df_equals(df1, df2, sort=False):
```
dgasmith: We want to keep this inline with QCElemental testing naming schema, so `df_compare` I believe.

qcfractal/testing.py
```diff
@@ -533,10 +536,73 @@ def sqlalchemy_socket_fixture(request, postgres_server):
...
+            df2 = df2.reindex(sorted(df2.columns), axis=1)
+        elif isinstance(df2, pd.Series):
+            df2 = df2.sort_index()
+
+    def element_equal(e1, e2):
+        if isinstance(e1, np.ndarray):
+            if not np.array_equal(e1, e2):
+                return False
+        elif isinstance(e1, Molecule):
+            if not e1.get_hash() == e2.get_hash():
```
dgasmith: We should build this into the Molecule class.

mattwelborn: Agreed. Resolved because we have an issue in elemental now.

mattwelborn:
@dgasmith Can I get a re-review? I will do include/exclude for other REST models in a separate PR.

qcfractal/interface/client.py
```diff
@@ -500,7 +500,11 @@ def list_collections(self, collection_type: Optional[str] = None, aslist: bool =
             df.sort_index(inplace=True)
             return df
 
-    def get_collection(self, collection_type: str, name: str, full_return: bool = False) -> 'Collection':
+    def get_collection(self,
+                       collection_type: str,
+                       name: str,
+                       full_return: bool = False,
```
dgasmith: Full_return/heavy is no longer used.

qcfractal/interface/client.py
```diff
@@ -520,8 +525,13 @@ def get_collection(self, collection_type: str, name: str, full_return: bool = Fa
         """
 
         payload = {"meta": {}, "data": {"collection": collection_type, "name": name}}
-        response = self._automodel_request("collection", "get", payload, full_return=True)
+        if include is None:
+            if collection_type.lower() in ["dataset", "reactiondataset"]:  #XXX
```
dgasmith: It would be best if these were collection properties, but we can adjust later.

PR #444, opened 2019-10-23
Feature/rp integration

andre-merzky:
## Description
Hi Daniel, thanks for your support with this RCT adaptor.  It now represents a state where some of the tests pass, and where it makes sense to do the func/exec conversion for the workload.

## Todos
  - [ ] some TODOs are noted in the adaptor code

## Questions
Alas, I lost the set of commands needed to run the unit tests.  I do something like:
```shell
$ source ~/miniconda/etc/profile.d/conda.sh
$ conda activate radical
$ pg_ctl -D /tmp/pg_data -l logfile start
$ pytest -vsk test_adapter
```
That seems to collect the tests, but does not run any of them and just hangs.  Any idea what step I am missing?

## Status
- [ ] requires completion, unit testing, integration testing

### DIFF
--- a/qcfractal/queue/radical_adapter.py
+++ b/qcfractal/queue/radical_adapter.py
@@ -1,0 +1,126 @@
+
+"""
+Queue adapter for Radical-Pilot
+"""
+
+import math
+import time
+import logging
+import traceback
+
+import radical.pilot as rp
+
+from typing import Any, Callable, Dict, Hashable, Optional, Tuple
+
+from qcelemental.models import FailedOperation
+
+from .base_adapter import BaseAdapter
+
+
+# ------------------------------------------------------------------------------
+#
+class RadicalAdapter(BaseAdapter):
+    """An Adapter for Radical-Pilot.
+    """
+
+    # --------------------------------------------------------------------------
+    #
+    def __init__(self, client: Any, logger: Optional[logging.Logger] = None, **kwargs):
+        BaseAdapter.__init__(self, client, logger, **kwargs)
+
+        self._session = client['session']
+        self._pmgr    = client['pmgr']
+        self._umgr    = client['umgr']
+        self._config  = client['config']
+        self._tasks   = list()
+        self._gc      = list()
+
+        # number of managers we can concurrently run
+        self._nmgrs = math.floor(self._config['pd']['cores'] /
+                                 self._config['mgr']['cores'])
+
+        print(self._config['pd'])
+      # pd = rp.ComputePilotDescription(self._config['pd'])
+      # self._pilot = self._pmgr.submit_pilots(pd)
+
+        pd = rp.ComputePilotDescription(self._config['pd'])
+        self._pilot = self._pmgr.submit_pilots(pd)
+        self._umgr.add_pilots(self._pilot)
+
+
+    # --------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


dgasmith:
Thanks for making the PR! A few notes:
- `radical.pilot` will not be a deployment dependency as users will choose the workload manager to use.
- The strategy we have often used is to use a function-local import rather than a file global one. As those functions can only be reached if an object of the import is provided it is a pretty safe way to organize things.

andre-merzky:
@dgasmith : thanks for reminding me of how to run the adaptor in a separate process, that helped.  I forgot that we need to set `export RADICAL_UTILS_NOATFORK=1`.  That will be fixed in the next RP release, but for now it's needed to run anything RP.

qcfractal/queue/radical_adapter.py
```diff
@@ -0,0 +1,122 @@
...
+
+      # # Form run tuple
+      # func = self.get_app(task_spec["spec"]["function"])
+      # task = func(*task_spec["spec"]["args"], **task_spec["spec"]["kwargs"])
+
+        cuds = list()
+        for i in range(0, 128):
+
+            cud = rp.ComputeUnitDescription()
+            cud.executable    = '/bin/sleep'
```
dgasmith, line 68: This will need to be the QCEngine CLI `qcengine`

qcfractal/queue/radical_adapter.py
```diff
@@ -0,0 +1,122 @@
...
+            cud.arguments     = ['2']
```
dgasmith, line 69: `qcengine.compute(model, "program")` is the `task_spec["spec"]["args"]` field so it will be:

```python
model, program = task_spec["spec"]["args"]
cud.arguments  = [model.json(), program]
```

qcfractal/queue/radical_adapter.py
```diff
@@ -0,0 +1,122 @@
...
+        # NOTE: why is this a private method, unlike the others?
+        # TODO: bulks
+
+      # # Form run tuple
+      # func = self.get_app(task_spec["spec"]["function"])
+      # task = func(*task_spec["spec"]["args"], **task_spec["spec"]["kwargs"])
+
+        cuds = list()
+        for i in range(0, 128):
+
```
dgasmith, line 66:
```python
if not task_spec["function"].split(".")[0].startswith("qcengine"):
    raise Exception("Only QCEngine is hooked in!")
```

qcfractal/queue/radical_adapter.py
```diff
@@ -0,0 +1,122 @@
...
+    def acquire_complete(self) -> Dict[str, Any]:
+
+        # FIXME: limit on stdout
+        done = {task for task in self._tasks
+                              if task.uid not in self._gc and
+                                 task.state in rp.FINAL}
+
+        self._gc += [task.uid for task in done]
+
+        return {task.name : task.stdout for task in done}
```
dgasmith, line 97: This will be raw JSON so we need to load it back in.
```python
return {task.name : json.loads(task.stdout) for task in done}
```

bennybp:
With v0.50, this feature is not applicable anymore. We have moved to a Parsl-only compute manager, although different kinds of managers are still possible in the future.

PR #445, opened 2019-10-23
Staging: copy specific datasets and results with datasets 

mattwelborn:
This PR adds a small update to create_staging.py which allows specific named datasets to be copied, and for results corresponding to a dataset to be copied. It hits a lot of lines because I added yapf to my precommit hooks.


dgasmith:
Forgot a pydantic pin. Will fix this tomorrow.

PR #447, opened 2019-10-23
Plaintext views

mattwelborn:
## Description
This PR implements plaintext view writing for Dataset and ReactionDataset. Plaintext views present headline results and molecules in CSV and XYZ format, respectively. These will be downloadable from the website. 

An example plaintext view for S22 is attached.
[S22.tar.gz](https://github.com/MolSSI/QCFractal/files/3764333/S22.tar.gz)

## Roadmap
This PR is part of the ANI-1 Milestone. 
image

## TODO
- [x] ~Change XYZ -> XYZ+ after MolSSI/QCElemental#152~

## Status
- [ ] Changelog updated
- [ ] Ready to go


qcfractal/alembic/versions/a2f76bb7be65_updates_collections_table_with_owner_.py
```diff
@@ -21,6 +21,7 @@ def upgrade():
     op.add_column('collection', sa.Column('description', sa.String(), nullable=True))
     op.add_column('collection', sa.Column('owner', sa.String(length=100), nullable=True))
     op.add_column('collection', sa.Column('view_url', sa.String(), nullable=True))
+    op.add_column('collection', sa.Column('view_url_plaintext', sa.String(), nullable=True))
```
dgasmith: Could we come up with a name for this? It isn't quite plaintext in some ways. Nothing is coming to me at the moment, but I will think on it.

Also what if we add the `view_url` as `view_url_hdf5`? Or we can leave it as is and consider it *the* internal QCArchive internal format.

@loriab Any ideas on a name for XYZ + CSV version of shipping around datasets?

qcfractal/interface/collections/dataset_view.py
```diff
@@ -495,3 +498,90 @@ def _deserialize(data: bytes, msgpacked_cols: List[str]) -> pd.DataFrame:
...
+            readme_path = temppath / "README"
+
+            mol_path.mkdir()
+            if "stoichiometry" in ds.data.history_keys:
+                molecules = ds.get_molecules(stoich=list(ds.valid_stoich),
+                                             force=True)  #TODO: change to function after pr 442
+            else:
+                molecules = ds.get_molecules(force=True)
+            for molecule in molecules['molecule']:
+                molecule.to_file(mol_path / f"{molecule.id}.xyz")
```
dgasmith: Should we try to name these similar to the entry names? It would make it a bit more readable.

qcfractal/interface/collections/dataset_view.py
```diff
@@ -495,3 +498,90 @@ def _deserialize(data: bytes, msgpacked_cols: List[str]) -> pd.DataFrame:
...
+            ds._disable_query_limit = ds_query_limit_state
+            df = ds.list_values(force=True).reset_index().set_index("name")
+            df["units"] = ds.units
+            for name in df.index:
+                if name in ds._column_metadata:
+                    if "units" in ds._column_metadata[name]:
+                        df['units'] = ds._column_metadata[name]["units"]
+            df.to_csv(list_path)
+
+            open(readme_path, 'w').write(self._readme(ds))
```
dgasmith: `with open`

qcfractal/interface/collections/dataset_view.py
```diff
@@ -495,3 +498,90 @@ def _deserialize(data: bytes, msgpacked_cols: List[str]) -> pd.DataFrame:
...
+        raise NotImplementedError()
+
+    @staticmethod
+    def _readme(ds) -> str:
+        # TODO: citations once we add that column
+        ret = f"""Name: {ds.name}
+
+Tagline: {ds.data.tagline}
+Tags: {", ".join(ds.data.tags)}
+Downloaded from: {ds.client.server_information()['name']}
```
dgasmith: We need units here.

qcfractal/tests/test_collections.py
```diff
@@ -1020,6 +1021,31 @@ def test_view_download_mock(gradient_dataset_fixture, tmp_path_factory):
                 ds.download(verify=True)
 
 
+def test_gradient_dataset_plaintextview_write(gradient_dataset_fixture, tmpdir):
+    _, ds = gradient_dataset_fixture
+    ptl.collections.PlainTextView(tmpdir / "test.tar.gz").write(ds)
```
dgasmith: Should we implement this similar to `ds.to_file("test.tar.gz", dtype="plaintext")`?

qcfractal/interface/collections/dataset_view.py
```diff
@@ -495,3 +498,90 @@ def _deserialize(data: bytes, msgpacked_cols: List[str]) -> pd.DataFrame:
...
+    def __init__(self, path: Union[str, pathlib.Path]) -> None:
+        """
+        Parameters
+        ----------
+        path: Union[str, pathlib.Path]
+            File path of view
+        """
+        if isinstance(path, str):
+            path = pathlib.Path(path)
+        self._path = path
```
dgasmith:
```
if not path.endswith(".tar.gz")
    path = path + ".tar.gz"
```

qcfractal/interface/collections/dataset_view.py
```diff
@@ -471,7 +474,7 @@ def list_values(self) -> pd.DataFrame:
         self._check_response_meta(response.meta)
         return self._deserialize(response.data, response.meta.msgpacked_cols)
 
-    def write(self, ds: Dataset) -> None:
+    def write(self, ds: Dataset) -> NoReturn:
```
dgasmith: Bleh, kind of wish this wasn't a thing.

qcfractal/interface/collections/dataset_view.py
mattwelborn: Units are in value_descriptions.csv to account for the possibility that columns have different dimensionality.

qcfractal/interface/collections/dataset_view.py
mattwelborn: Entry names are not great paths since they can contain spaces and other naughty characters. Additionally, names don't work well for reaction datasets with complicated stoichiometry. `entries.csv` provides a join table between entries and molecule ids.

qcfractal/alembic/versions/a2f76bb7be65_updates_collections_table_with_owner_.py
mattwelborn: I'm okay with changing `view_url` to `view_url_hdf5`.

The point of XYZ + CSV is that they are familiar, broadly compatible, and inspect-able with a text editor -- at the cost of performance. We could call it:
- `xyz+csv`
- `tar`
- `plain`

mattwelborn:
Thanks for the review @dgasmith!

Something seems to be going wrong on travis. Is it a Pydantic 1.0 issue?

dgasmith:
Yes, a 1.0 issue.

qcfractal/interface/collections/dataset_view.py
dgasmith: Does the phrase "join table" conflict with "familiar, broadly compatible, and inspect-able"? I agree that entry names are difficult (but can be converted), but are likely more readable. I wonder if we could consider a hybrid `entry_name.replace(" ", "_")... + stoich + mol_uid`?

qcfractal/interface/collections/dataset_view.py
dgasmith: Goodp oint.

qcfractal/alembic/versions/a2f76bb7be65_updates_collections_table_with_owner_.py
dgasmith: `ascii`?

mattwelborn: `text`?

dgasmith: Hmm, `plaintext` + `xyz+csv` is still my favorites.

dgasmith: Still developer facing so some thought is needed. Happy to merge this as is if you wish.

qcfractal/interface/collections/dataset_view.py
mattwelborn: Good idea. I think I have done this with correct path normalization.

qcfractal/alembic/versions/a2f76bb7be65_updates_collections_table_with_owner_.py
mattwelborn: `TextTables`?

dgasmith: I give in on plaintext, I don't think we will have a `plaintextv2`.

PR #449, opened 2019-10-26
Inspect ORM classes with column types, fixes #441

### LINKED ISSUE #441: sql_base.Base.col() should not return methods
sql_base.Base.col() should not return methods such as 'update_relations'

doaa-altarawy:
## Description
Inspect ORM classes with column types, fixes #441  
Needed for exclude and include in queries.

## Questions
- [ ] Question1

## Status
- [x] Ready to go

### DIFF
--- a/qcfractal/storage_sockets/sqlalchemy_socket.py
+++ b/qcfractal/storage_sockets/sqlalchemy_socket.py
@@ -2408,7 +2408,9 @@ class SQLAlchemySocket:
             else:
                 current = db_ver.first()
 
-            return current.to_dict(exclude=['id'])
+            ver = current.to_dict(exclude=['id'])
+
+        return ver
 
     def get_total_count(self, className, **kwargs):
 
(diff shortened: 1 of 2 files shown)


dgasmith:
@doaa-altarawy Do we need any tests here? If not, feel free to merge.

qcfractal/storage_sockets/sqlalchemy_socket.py
```diff
@@ -2408,7 +2408,7 @@ def check_lib_versions(self):
             else:
                 current = db_ver.first()
 
-        return current.to_dict(exclude=['id'])
+            return current.to_dict(exclude=['id'])
```
dgasmith: Feels a bit awkward to return inside a `with` statement. Can we case these to dict inside we `with` and then return outside?

doaa-altarawy: ok, I'll change and merge

PR #450, opened 2019-10-28
Update ContributedValues model

mattwelborn:
## Description
This PR updates Dataset's ContributedValues model for the 0.12 release. 

@doaa-altarawy Is this enough for you to go on when implementing the contributed values table as we discussed today? Let me know if there is any more information or clarification I can provide.

### DIFF
--- a/qcfractal/interface/models/common_models.py
+++ b/qcfractal/interface/models/common_models.py
@@ -201,6 +201,7 @@ class KeywordSet(ProtoModel):
 class Citation(ProtoModel):
     """ A literature citation.  """
     _citation: str  # hand-formatted citation in ACS style. In the future, this could be bibtex, rendered to different formats.
+    _bibtex: Optional[str]  # bibtex blob for later use with bibtex-renderer
     doi: Optional[str]
     url: Optional[str]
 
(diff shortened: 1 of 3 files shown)


qcfractal/interface/models/common_models.py
```diff
@@ -196,3 +196,14 @@ def __init__(self, **data):
 
     def get_hash_index(self):
         return hash_dictionary(self.values.copy())
+
+
+class Citation(ProtoModel):
+    """ A literature citation.  """
+    _citation: str  # hand-formatted citation in ACS style. In the future, this could be bibtex, rendered to different formats.
+    doi: Optional[str]
+    url: Optional[str]
```
dgasmith: This should optionally allow full definitions:
```
author: List[str]
title: str
journal: str
year: int
volume: int
number: int
pages Tuple[int, Optional[int]]
month: int
doi: str
note: str
key: str
```

mattwelborn: The plan is to store the citation object as a JSON blob in the DB to allow for future improvement. At this stage, we will pre-render citations in `_citation`, but later we will have the option of using the citation manager tech to render citations from bibtex.

dgasmith: Sounds good, but lets make sure we have the option and correct places to keep all of the required metadata currently. Going back and having to re-enter the information seems rough.

doaa-altarawy:
Please put back the doi, then will merge.

mattwelborn:
Now part of #460.

PR #452, opened 2019-10-29
Use inlcude/exclude in queries with hybrid prop

doaa-altarawy:
## Description
Add exclude/include functionality in the backend Database layer
`include` is still called projection, and it takes a list (not dict), although dict works by grabbing the keys.

This works as expected. If any include (projection list) or exclude is specified, then only requested fields will be fetched/excluded.

## Status
- [x] Ready to go

### DIFF
--- a/qcfractal/storage_sockets/sqlalchemy_socket.py
+++ b/qcfractal/storage_sockets/sqlalchemy_socket.py
@@ -283,18 +283,66 @@ class SQLAlchemySocket:
         return limit if limit and limit < self._max_limit else self._max_limit
 
     def get_query_projection(self, className, query, projection, limit, skip, exclude=None):
+        # Todo: projection to be renamed to include_only
+
+        if projection and exclude:
+            raise AttributeError(f'Either projection (include) or exclude can be '
+                                 'used, not both at the same query. '
+                                 'Given projection: {prjection}, exclude: {exclude}')
+
+        prop, hybrids, relationships = className._get_col_types()
+
+        # build projection from include or exclude
+        _projection = []
+        if projection:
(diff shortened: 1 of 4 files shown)


qcfractal/storage_sockets/sqlalchemy_socket.py
```diff
@@ -283,18 +283,66 @@ def get_limit(self, limit: Optional[int]) -> int:
         return limit if limit and limit < self._max_limit else self._max_limit
 
     def get_query_projection(self, className, query, projection, limit, skip, exclude=None):
+        # Todo: projection to be renamed to include_only
+
+        if projection and exclude:
+            raise AttributeError(f'Either projection (include) or exclude can be '
+                                 'used, not both at the same query. '
+                                 'Given projection: {prjection}, exclude: {exclude}')
+
```
mattwelborn: Typo in "projection"

mattwelborn:
@doaa-altarawy Thank you for this! The code looks good to me. I checked it out (in a git sense), and it passes my simple storage tests in PR#442. I'm going to merge this into #442 and make sure that everything works in that context.

dgasmith:
@mattwelborn Does #442 have tests for include/exclude? If so we can merge this without tests for now.

Do we want to deprecate the name `projection` for `include` as well?

mattwelborn:
@dgasmith #442 has tests for include/exclude. My plan is to deprecate `projection` for `include` in all of the REST models. 

@doaa-altarawy chose to keep `projection` in `sqlalchemy_socket.py`, so it would be good to check with her before making this change everywhere.

PR #457, opened 2019-10-31
Change projection to include/exclude

mattwelborn:
## Description
This PR renames projection to include in the rest models and sqlalchemy socket. While exclude functionality is supported, it is not tested in all cases and not exposed in Portal except where needed. 

## Status
- [ ] Changelog updated
- [x] Ready to go


qcfractal/tests/test_storage.py
```diff
@@ -269,86 +269,6 @@ def test_collections_overwrite(storage_socket):
     assert ret == 1
 
 
-def test_collections_projection(storage_socket):
```
mattwelborn: Duplicate test.

PR #458, opened 2019-10-31
Collections: Add metadata field to DataModel

mattwelborn:
## Description
This PR adds a metadata field to collections. It uses the extras functionality; there is no schema migration.

## Status
- [ ] Changelog updated
- [x] Ready to go

### DIFF
--- a/qcfractal/interface/collections/collection.py
+++ b/qcfractal/interface/collections/collection.py
@@ -77,6 +77,8 @@ class Collection(abc.ABC):
         view_metadata: Optional[Dict[str, str]] = None
         view_available: bool = False
 
+        metadata: Optional[Dict[str, Any]] = {}
+
     def __str__(self) -> str:
         """
         A simple string representation of the Collection.
--- a/qcfractal/tests/test_collections.py
+++ b/qcfractal/tests/test_collections.py
@@ -1373,28 +1373,9 @@ def test_get_collection_no_records_ds(fractal_compute_server):
 
 def test_collection_metadata(fractal_compute_server):
     client = ptl.FractalClient(fractal_compute_server)
-    name = "test_collection_metadata"
-    ds = ptl.collections.Dataset(name, client=client)
 
-    class MlMetadata(ProtoModel):
-        data_points: int
-        elements: List[str]
-        theory_level: str
-        sampling: str
-        labels: List[str]
-
-    meta = MlMetadata(data_points=133_885,
-                      elements=['C', 'O', 'N', 'F', 'H'],
-                      theory_level="DFT",
-                      sampling="Minima",
-                      labels=[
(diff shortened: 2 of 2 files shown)


qcfractal/interface/collections/collection.py
```diff
@@ -77,6 +77,8 @@ class DataModel(ProtoModel):
         view_metadata: Optional[Dict[str, str]] = None
         view_available: bool = False
 
+        metadata: Optional[Dict[str, Any]] = {}
```
dgasmith: We could consider growing this over time `Union[MlMetaData, Dict[str, Any], None]`. As long as there is a concrete key in the models we can automatically cast them to the correct ProtoModel.

A key could be something like `type: constr('mytype') = 'mytype'`.

qcfractal/tests/test_collections.py
```diff
@@ -1368,3 +1369,32 @@ def test_get_collection_no_records_ds(fractal_compute_server):
     ds.get_entries()
     assert len(ds.data.records) == 1
     assert ds.data.records[0].name == "He1"
+
+
+def test_collection_metadata(fractal_compute_server):
```
dgasmith: I usually like more tests, but this seems like a lot of volume to test one field?

mattwelborn: Fair enough, how about now?

qcfractal/tests/test_collections.py
```diff
@@ -1368,3 +1369,13 @@ def test_get_collection_no_records_ds(fractal_compute_server):
...
+    client = ptl.FractalClient(fractal_compute_server)
+
+    ds = ptl.collections.Dataset("test_collection_metadata", client=client)
+    ds.data.metadata["data_points"] = 133_885
```
dgasmith: This test indicates that `metadata` is not `Optional` which implies a `None` value is acceptable. Good example of how Pydantic is clarifying our thinking on `Optional`.

mattwelborn: It's a little tricky since existing collections will not have metadata. I wanted to allow for this possibility, but also to make new collections have an empty metadata dict going forward. Test has been updated accordingly.

dgasmith: Lets do a data migration in this case. Add a new `metadata` column with `{}` as default. Should save us some headaches in the future.

mattwelborn: I thought we decided against this due to the partially migrated state of the collections table on molssi10

dgasmith: Need to chat about this in person, but I think we need to hand alter the tables back to v0.11 schema anyways.

dgasmith:
Issue #465 spawned from this.

PR #462, opened 2019-11-01
Dataset: add a progress bar for download

mattwelborn:
## Description
This PR adds a progress bar to dataset's download function, since this sometimes can take a while.

- [ ] Changelog updated
- [x] Ready to go

### DIFF
--- a/qcfractal/interface/collections/dataset.py
+++ b/qcfractal/interface/collections/dataset.py
@@ -10,8 +10,10 @@ from typing import TYPE_CHECKING, Any, Dict, List, Optional, Set, Tuple, Union
 
 import numpy as np
 import pandas as pd
-import pydantic
 import requests
+from tqdm import tqdm
+
+import pydantic
 from qcelemental import constants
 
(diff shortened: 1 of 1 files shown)


qcfractal/interface/collections/dataset.py
```diff
@@ -155,9 +161,21 @@ def download(self, local_path: Optional[Union[str, Path]] = None, verify: bool =
...
         r = requests.get(self.data.view_url_hdf5, stream=True)
+
+        pbar = None
+        try:
+            if progress_bar:
+                file_length = int(r.headers.get('content-length'))
+
+                pbar = tqdm(total=file_length, initial=0, unit='B', unit_scale=True)
+        except Exception as e:
+            print(e)
```
mattwelborn: I would like to turn the exception into a warning here, but I'm not sure how.

dgasmith: Why do you need the exception? You would just warn if you hit the execution right?

PR #463, opened 2019-11-01
client.list_collection filter on owner and visibility

mattwelborn:
## Description
This PR adds functionality to `FractalClient.list_collections` to filter on owner and visibility. By default,
- collections with `visibility=False` are not shown
- only collections with `owner="default"` are shown, unless the client is logged in, in which cae only collections with `owner=self.username` are shown.  A special value of `"*"` (open to changing this) allows collections to not be filtered on owner. 

When new collection are added, their owner is set to the username that created them. During migration from 0.11.0, the default value for collection owners is `"default"`. 

## Status
- [x] Changelog updated
- [ ] Ready to go


qcfractal/alembic/versions/a2f76bb7be65_updates_collections_table_with_owner_.py
```diff
@@ -20,6 +20,7 @@ def upgrade():
     # ### commands auto generated by Alembic - please adjust! ###
     op.add_column('collection', sa.Column('description', sa.String(), nullable=True))
     op.add_column('collection', sa.Column('owner', sa.String(length=100), nullable=True))
+    op.excecute("UPDATE collection SET owner=default")
```
mattwelborn: @dgasmith @doaa-altarawy Could one of you please verify that (i) this is correct SQL and (ii) this change to the migration won't mess with our current partially migrated state on molssi10.

qcfractal/interface/collections/collection.py
```diff
@@ -238,6 +238,10 @@ def save(self, client: Optional['FractalClient'] = None) -> 'ObjectId':
 
         self._pre_save_prep(client)
 
+        if self.data.owner is None:
```
dgasmith: Can this be a Collection default?

qcfractal/interface/collections/collection.py
```diff
@@ -238,6 +238,10 @@ def save(self, client: Optional['FractalClient'] = None) -> 'ObjectId':
...
+            if self.client.username is not None:
+                self.data.__dict__["owner"] = self.client.username
```
dgasmith: Pretty sure we don't want to do this until we have fleshed out what `owner` really means?

qcfractal/alembic/versions/a2f76bb7be65_updates_collections_table_with_owner_.py
dgasmith: Should be fine.

qcfractal/interface/collections/collection.py
mattwelborn: The idea was to make this work:
```
client = ptl.FractalClient(**login)
ds = ptl.collections.Dataset(name="foo", client=client)
ds.save()
client.list_collections()
```

qcfractal/tests/test_collections.py
```diff
@@ -1157,7 +1157,7 @@ def test_reactiondataset_dftd3_dataset_plaintextview_write(reactiondataset_dftd3
 def test_collection_query(fractal_compute_server):
     client = ptl.FractalClient(fractal_compute_server)
 
-    ds = ptl.collections.Dataset("CAPITAL", client)
+    ds = ptl.collections.Dataset("CAPITAL", client, owner="default")
```
doaa-altarawy: If it's the `owner="default"`, then we shouldn't specify it in the call., it should be the default.

mattwelborn: It's only the default if the client is not logged in.

doaa-altarawy: Even If I'm logged in, I won't like to write this in every query. I should always get the same behavior and use the same code either when I'm logged in or not and get the same results.

dgasmith: I'm not really a fan of having this set as the logged in user as it could lead to confusion. What if we change this to `group` to remove the single-user ownership notion and give us grounds for integration into the groups framework that we sort of have planned?

mattwelborn: Per discussion with @dgasmith, we will:
- change owner to group
- the default value for group for new datasets will be "default"
- the default value for group for client.list_collections will be "default"
- there will be no username-related funny business

qcfractal/alembic/versions/a2f76bb7be65_updates_collections_table_with_owner_.py
```diff
@@ -19,7 +18,8 @@
 def upgrade():
     # ### commands auto generated by Alembic - please adjust! ###
     op.add_column('collection', sa.Column('description', sa.String(), nullable=True))
-    op.add_column('collection', sa.Column('owner', sa.String(length=100), nullable=True))
+    op.add_column('collection', sa.Column('group', sa.String(length=100), nullable=True))
+    op.excecute("UPDATE collection SET group=default")
```
dgasmith: After the execute add this so there must be a group: `op.alter_column('collection', 'group', nullable=False)`

PR #467, opened 2019-11-06
Migration: Updates contrib values migration

dgasmith:
## Status
- [ ] Changelog updated
- [x] Ready to go


qcfractal/alembic/versions/a2f76bb7be65_updates_collections_table_with_owner_.py
```diff
@@ -22,10 +22,6 @@ def upgrade():
     op.add_column('collection', sa.Column('group', sa.String(), nullable=False, server_default='default'))
     op.alter_column('collection', 'group', server_default=None)
 
-    op.add_column('collection', sa.Column('metadata', sa.JSON(), nullable=True))
-    op.execute("UPDATE collection SET metadata = '{}'::json")
-    op.alter_column('collection', 'metadata', nullable=False)
-
```
doaa-altarawy: This is on purpose, right?

dgasmith: Yes, @mattwelborn SQLAlchemy does not allow you to have a `metadata` field.d

mattwelborn: do you mean that "metadata" is a reserved name?

dgasmith: A reserved class attribute, yes.

PR #468, opened 2019-11-06
Fix citation model

mattwelborn:
## Description
This PR fixes the citation model:

- underscore-prefixed fields are renamed since pydantic does not treat these as fields
- everything is now optional
- bibtex is a string because unpacking it into a dictionary is too fancy

Sorry for not catching this earlier. 

## Status
- [x] Ready to go

### DIFF
--- a/qcfractal/interface/models/common_models.py
+++ b/qcfractal/interface/models/common_models.py
@@ -200,7 +200,8 @@ class KeywordSet(ProtoModel):
 
 class Citation(ProtoModel):
     """ A literature citation.  """
-    acs_citation: str = None  # hand-formatted citation in ACS style. In the future, this could be bibtex, rendered to different formats.
+    acs_citation: Optional[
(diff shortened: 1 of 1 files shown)


qcfractal/interface/models/common_models.py
```diff
@@ -200,11 +200,11 @@ def get_hash_index(self):
 
 class Citation(ProtoModel):
     """ A literature citation.  """
-    _acs_citation: str  # hand-formatted citation in ACS style. In the future, this could be bibtex, rendered to different formats.
-    _bibtex: Optional[Dict[str, str]]  # bibtex blob for later use with bibtex-renderer
-    doi: Optional[str]
-    url: Optional[str]
+    acs_citation: str = None  # hand-formatted citation in ACS style. In the future, this could be bibtex, rendered to different formats.
```
dgasmith: `Optional[str]`

mattwelborn: Thanks for catching this. I'm clearly struggling with this particular model :P

PR #470, opened 2019-11-07
Dataset: don't scale non-default units

mattwelborn:
## Description
This PR fixes a bug in Dataset which caused data columns whose units do not match those of the dataframe to be incorrectly scaled. For example, in a dataset whose `default_driver` is energy but which contains a contributed values column of dipole moments, setting `self.units = "hartree" (from `self.units="kcal/mol"`) would cause the dipole moments to be scaled by 1/627.  

## Status
- [ ] Changelog updated
- [x] Ready to go

### DIFF
--- a/qcfractal/interface/collections/dataset.py
+++ b/qcfractal/interface/collections/dataset.py
@@ -11,11 +11,10 @@ from typing import TYPE_CHECKING, Any, Dict, List, Optional, Set, Tuple, Union
 import numpy as np
 import pandas as pd
 import requests
-from tqdm import tqdm
-
 from pydantic import Field, validator
 from qcelemental import constants
 from qcelemental.models.types import Array
+from tqdm import tqdm
 
 from ..models import Citation, ComputeResponse, ObjectId, ProtoModel
 from ..statistics import wrap_statistics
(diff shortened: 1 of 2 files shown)


qcfractal/interface/collections/dataset.py
```diff
@@ -1171,8 +1174,19 @@ def units(self):
 
     @units.setter
     def units(self, value):
-
-        self.df *= constants.conversion_factor(self._units, value)
+        for column in self.df.columns:
+            try:
+                self.df[column] *= constants.conversion_factor(self._column_metadata[column]["units"], value)
+                if self._column_metadata[column]["units"] != self._units:
```
mattwelborn: there might be a better way to check this to not trigger on "kcal/mol" != "kcal / mol"

dgasmith:
```python
>>> qcel.constants.Quantity('kcal/mol') == qcel.constants.Quantity('kcal / mol')
True
```

dgasmith: Or more concrete:
```python
>>> qcel.constants.Quantity('kcal/mol') == qcel.constants.Quantity('kilocal / mole')
True
```

PR #477, opened 2019-11-11
Testing and LGTM un-regressions

mattwelborn:
## Description
This PR:
1) Removes the xfail from tests involving the QCA server now that it is updated to >=0.12.0. 
2) Runs `make format`
3) Removes `LGTM py/unused-import` annotations since Semmle/ql#2014 is resolved.

## Changelog description
This PR does not need a change log line.

## Status
- [x] Ready to go

### DIFF
--- a/qcfractal/interface/collections/dataset.py
+++ b/qcfractal/interface/collections/dataset.py
@@ -10,10 +10,9 @@ import numpy as np
 import pandas as pd
 import requests
 from pydantic import Field, validator
-from tqdm import tqdm
-
 from qcelemental import constants
 from qcelemental.models.types import Array
+from tqdm import tqdm
 
 from ..models import Citation, ComputeResponse, ObjectId, ProtoModel
 from ..statistics import wrap_statistics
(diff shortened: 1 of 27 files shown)


qcfractal/interface/collections/torsiondrive_dataset.py
```diff
@@ -12,7 +12,7 @@
 from .collection_utils import register_collection
 
 if TYPE_CHECKING:  # pragma: no cover
-    from ..models import Molecule  # lgtm[py/unused-import] (https://github.com/Semmle/ql/issues/2014)
```
mattwelborn: It looks like autoflake is unaware of `if TYPE_CHECKING` in the same way that LGTM was. I guess we're going to have to substitute the lgtm annotation for a pyflakes annotation...

dgasmith: Maybe turn off `auto flake` for now? Its nice, but not needed.

qcfractal/interface/collections/dataset.py
```diff
@@ -10,10 +10,9 @@
 import pandas as pd
 import requests
 from pydantic import Field, validator
-from tqdm import tqdm
-
 from qcelemental import constants
 from qcelemental.models.types import Array
+from tqdm import tqdm
```
dgasmith: Unsure why this order was changed?

mattwelborn: I think it alphabetizes.

PR #481, opened 2019-11-13
Improve ALCF and Parsl Support

WardLT:
## Description
Improves support for running tasks at ALCF with Parsl

Specific changes include:
- Adding support for Cobalt scheduler to Parsl adapters
- task_per_worker now maps to "max_workers" on the HTEx Parsl executor, which properly controls the number of concurrent tasks per worker
- Parsl provider sets the number of nodes per block, which makes it possible to request >1 node per job
- The maximum number of "blocks" requested by the Parsl provider is now set by the maximum number of workers and the number of nodes per block (which defines the number of workers per block)

## Changelog description
Improves support for Parsl at Argonne Leadership Computer Facility

## Status
- [x] Ready to go. I have tested this on Theta, not sure if I need more tests for the CI

### DIFF
--- a/qcfractal/cli/qcfractal_manager.py
+++ b/qcfractal/cli/qcfractal_manager.py
@@ -426,9 +426,10 @@ class ParslLauncherSettings(AutodocBaseSettings):
 
 class ParslProviderSettings(SettingsBlocker):
     """
-    Settings for the Parsl Provider class. Valid values for this field are functions of your cluster.scheduler and no
-    linting is done ahead of trying to pass these to Parsl.
-    Please see the docs for the provider information
+    Settings for the Parsl Provider class. Valid values for this field depend on your choice of  cluster.scheduler and
+    are defined in `the Parsl docs for the providers
+    <https://parsl.readthedocs.io/en/stable/userguide/execution.html#execution-providers>`_ with some minor exceptions.
+    The initializer function for the Parsl settings will indicate which
 
     NOTE: The parameters listed here are a special exception for additional features Fractal has engineered or
     options which should be considered for some of the edge cases we have discovered. If you try to set a value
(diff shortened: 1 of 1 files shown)


WardLT:
Is the error from Travis CI a transient/random error? It does not seem like the failing tests should have been affected by my changes.

qcfractal/cli/qcfractal_manager.py
```diff
@@ -440,6 +442,14 @@ class ParslProviderSettings(SettingsBlocker):
 
     """
 
+    def __init__(self, **kwargs):
+        if 'nodes_per_block' in kwargs:
+            raise ValueError('``nodes_per_block`` is named ``nodes_per_job`` to keep nomenclature consistent')
```
dgasmith: We may want to keep kwargs the same us the underlying adapter. A bit on the fence about this.

WardLT: I was on the fence as well and would be happy with "block" or "job".

What sent me over to "nodes_per_job" is that the documentation for the different adapters took awhile to understand and required frequent consulting with the nomenclature definitions. So, I did not want to further introduce confusion by adding new terms unnecessarily.

dgasmith: We don't see to use the `job` notion that often. Perhaps leaving it as `nodes_per_block` with an explanation may be best. Something like "`nodes_per_block` allows a single job to span multiple nodes with a worker on each node, this is useful to limit the number of jobs submitted to the resource scheduler". Though that highlights the block/job issue, but I do like attempting to keep the scheduler specific kwargs equal to their own specifications.

We could also consider adding ``nodes_per_job`` to the common settings and enforce everyone besides Parsl being one.

@mattwelborn

WardLT: I like that idea of it being a common setting best, as notions of requesting batches of nodes in a single job is not a Parsl-specific thing. Is it more of a "cluster" common setting though?

dgasmith: My reasoning for the `common` setting is this is where the amount of resources requested is specified. `cluster` seems a bit more ancillary about the "how".

dgasmith: I should note that we should probably shake up the names and organization once more before 1.0. The shake up is not deep since most seem fine, but worth another reconsideration.

We should also support building one of these in the python layer as a first class citizen. This likely requires interrogation of the queue adapter objects for number of cores/memory that they have set internally.

WardLT:
Should I provide an example YML configuration for the documentation? I can do so in another PR

dgasmith:
It should be automatically documented in the settings description. Give it a try and make a PR with an example if everything works?

PR #483, opened 2019-11-15
Faster views

mattwelborn:
## Description
This PR makes dataset views faster by correctly setting the index in the dataframe used to cache entries. This is especially important for ANI-1 where getting 1,000 values now takes about 0.5s compared to 30s before (after a burn-in period of indexing the entries the first time `get_values` is called.).

## Changelog description
Improved performance of very large dataset views.

## Status
- [x] Ready to go


mattwelborn:
I've tested this on a staging server, and it makes ANI-1 very fast.

PR #485, opened 2019-11-16
Add client-side tag filtering to list_collections

mattwelborn:
## Description
This PR adds client-side tag filtering to `FractalClient.list_collections()`. This lets us do things like `client.list_collections(tag="machine learning")`.

## Changelog description
FractalClient.list_collections() now takes an optional argument `tag` which allows collections to be filtered their tags.

## Status
- [ ] Ready to go

### DIFF
--- a/qcfractal/interface/client.py
+++ b/qcfractal/interface/client.py
@@ -484,6 +484,7 @@ class FractalClient(object):
         aslist: bool = False,
         group: Optional[str] = "default",
         show_hidden: bool = False,
+        tag: Optional[Union[str, List[str]]] = None
     ) -> pd.DataFrame:
         """Lists the available collections currently on the server.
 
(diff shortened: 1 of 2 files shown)


qcfractal/interface/client.py
```diff
@@ -524,7 +527,13 @@ def list_collections(
             df = df[df["visibility"]]
         if group is not None:
             df = df[df["group"].str.lower() == group.lower()]
-        df.drop(["visibility", "group"], axis=1, inplace=True)
+        if tag is not None:
+            if isinstance(tag, str):
+                tag = [tag]
+            tag = {t.lower() for t in tag}
+            df = df[df.apply(lambda x: len({t.lower() for t in x['tags']} & tag) > 0, axis=1)]
```
mattwelborn: @dgasmith someday, all of dataset will be a pandas one-liner.

dgasmith: Or in SQL :)

PR #502, opened 2019-12-08
Docker: add gcp to qcarchive_worker_openff and switch to stable psi

mattwelborn:
## Description
This PR adds gcp to the qcarchive_worker_openff docker image.

## Changelog description
(Probably not needed.)

## Status
- [ ] Ready to go


dgasmith:
Intermittent failures still not fixed on the TD Datasets. Similar issues, might consider moving away from RDKit if this instability stays.

A bit odd that we can run quantum chemistry flawlessly, but MD is causing issues.

PR #506, opened 2019-12-12
RDS Fix

dgasmith:
## Description
See below.

## Changelog description
Fixes metadata corruption on repeated `visualize` calls.

## Status

- [x] Code base linted
- [x] Ready to go

### DIFF
--- a/qcfractal/tests/test_collections.py
+++ b/qcfractal/tests/test_collections.py
@@ -1207,6 +1207,7 @@ def test_missing_collection(fractal_compute_server):
 
 
 @pytest.mark.slow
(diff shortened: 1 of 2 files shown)


qcfractal/tests/test_collections.py
```diff
@@ -1207,6 +1207,7 @@ def test_missing_collection(fractal_compute_server):
 
 
 @pytest.mark.slow
+@pytest.mark.xfail(reason="Flaky on Travis CI")
```
mattwelborn: Can you please add the issue # to this message and close #508?

PR #507, opened 2019-12-12
Collection: add new molecules in chunks

mattwelborn:
This PR adds molecules in chunks inside `Collection.save()`.

## Changelog description
Fixed an issue that could arise when adding large numbers of molecules to a collection.

## Status

- [x] Code base linted
- [ ] Ready to go

### DIFF
--- a/qcfractal/interface/collections/collection.py
+++ b/qcfractal/interface/collections/collection.py
@@ -266,7 +266,10 @@ class Collection(abc.ABC):
             flat_map_keys.append(k)
             flat_map_mols.append(v)
 
-        mol_ret = client.add_molecules(flat_map_mols)
+        CHUNK_SIZE = client.query_limit
+        mol_ret = []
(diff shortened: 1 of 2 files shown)


qcfractal/interface/collections/collection.py
```diff
@@ -266,7 +266,10 @@ def _add_molecules_by_dict(client, molecules):
             flat_map_keys.append(k)
             flat_map_mols.append(v)
 
-        mol_ret = client.add_molecules(flat_map_mols)
```
mattwelborn: Perhaps we should have a loop like this implemented somewhere in FractalClient.

dgasmith: I like some sort of wrapper:

```python
def chunk_list(data_list, chunk_size, tqdm=None):
    if tqdm is None:
        tqdm = is_jupyter()

   for x in range(0, len(data_list), chunk_size):
        yield data_list[x : x + chunk_size]

    # handle tqdm

```

Issue #515, opened 2019-12-17
Docker Images and Conda Envs


dgasmith:
It would be nice to build the docker images from our conda envs and keep them in sync so that the two can be created from either side. Either a airflow or GHA trigger I think?

Bonus for a small part of the docs listing available images.

@jchodera

mattwelborn:
Could you please unpack "created from either side"?

dgasmith:
The conda and Docker images should be as identical as possible. This means that we would be able to create a `qcfractal_oss_worker` environment from a conda environment file uploaded to our QCArchive conda channel (https://anaconda.org/QCArchive) as well as a `qcfractal_oss_worker` Docker image on our MolSSI Docker channel where the Docker image would be created using `conda env qcarchive/qcfractal_oss_worker` so that the two are nearly identical.

mattwelborn:
The conda envs don't include fractal and the Docker images do. Are you envisioning adding four new Docker images corresponding the conda envs without fractal, but tagged corresponding to the Fractal version they came from? On the flip side, there would be two new conda envs corresponding to the Docker images, which contain specific versions of fractal?

dgasmith:
I am thinking of the [prod-envs](https://github.com/MolSSI/QCFractal/blob/master/devtools/prod-envs/qcf-manager-openff.yaml) not the `conda-envs` where we get this folder and the Docker images in sync. The `conda-envs` might need to be renamed `ci-envs` or similar to be a bit more explicit.

Apologies for the confusion there, we have many envs and images running around.

jchodera:
What about the [`qcfractal/qcf-snowflake`](https://anaconda.org/QCArchive/qcf-snowflake/files) environment described in the [QCFractal Quickstart](http://docs.qcarchive.molssi.org/projects/QCFractal/en/stable/quickstart.html) docs? Would be great if this could be kept in sync too, since it's super easy to reproducibly create the correct environment this way.

dgasmith:
Yup, the snowflake environment is part of the `prod-envs` folder. We have all of the pieces in place, the only item that we need is a bit more automation to spawn the updating of these files on conda + Docker rebuild.

I think this is likely a good candidate for some internal Airflow workflows. If it isn't Airflow we would need to spin this off into (yet) another repo that can be spawned via GHA in a sane manner.

dgasmith:
Probably closed by #518. Will close once everything is up.

mattwelborn:
Happiness is a green DAG:
[...]

PR #518, opened 2019-12-18
devtools: sync dockerfiles and prod_envs

mattwelborn:
## Description
This PR updates the prod_envs to match the dockerfiles. The Dockerfiles are updated to use the prod_envs, pulled down from Anaconda Cloud. 

Once this is merged, #515 can be addressed by making an airflow corresponding to:
- Pull latest qcfractal
- Upload each of the prod envs to anaconda cloud
- Edit each dockerfile to pin QCFractal to desired version
- Build, tag, and push each docker image to docker hub

## Changelog description
Conda environments now correspond to docker images. 

## Status

- [x] Code base linted
- [x] Ready to go


devtools/prod-envs/qcfractal.yaml
```diff
@@ -0,0 +1,11 @@
+name: qcfractal
+channels:
+  - defaults
+  - conda-forge
+dependencies:
+  - python
+  - qcfractal
```
mattwelborn: Should this one also contain `qcfractal-dashboard`?

dgasmith: I would say not until this is stable, but then yea lets include it. Can we cap `python = 3.7`?

PR #521, opened 2019-12-19
Queue: ProcessPoolExecutor now correctly counts running tasks

### LINKED ISSUE #517: Process Pool Manager Active Task Tracking
**Describe the bug**
Completed tasks remain "active" with the process pool manager so that "active" tasks always grows.

**To Reproduce**
View the active tasks of a process pool manager over time.

**Additional context**
The issue appears to stem from the fact that `pool._queue_count` is always increasing and does not reset under any circumstance of a given task. `pool._pending_work_items appears` to work correctly and perhaps a solution to active tasks is `min(pool._max_workers, len(pool._pending_work_items))` which from some testing fixes this issue.

Relevant line: `executor_adapter.py:39`.

dgasmith:
## Description
Fixes #517, be correctly counting tasks. The test is good to have, but I worry that it is slightly unstable as testing a fully asynchronous framework in such a fashion may be risky. Note the `time.sleep(1)` to allow Python to fully process the evaluation of the task.

Edit: The current version was platform dependent. Simply dropping the ability to check current active tasks now as this metric is rarely useful. Workload managers #1 goal is fully satisfy the current workspan so we can assume that active tasks is roughly equivalent to `min(current_active_slots, current_queue_depth)`. This isn't a very useful value at the moment and is omitted. This PR still fixes #517.

## Changelog description
Fixes an issue where ProcessPoolExecutor returned the incorrect number of currently running tasks.

## Status

- [x] Code base linted
- [x] Ready to go


dgasmith:
Expected this to hit some deeper issues, will look at fixing when I can build some envs.

dgasmith:
@mattwelborn or @Lnaden  Please review!

PR #522, opened 2019-12-19
Collections: Allows subset status queries, fixes #514

### LINKED ISSUE #514: ProcedureDataset Status Failure
**Describe the bug**
`ProcedureDataset.status()` will fail for sparsely mapped specifications.

**To Reproduce**
Evaluate any procedure dataset on a subset and then query the status of the results.

```python
ProcedureDataset.compute("default", subset={"subset1"})
ProcedureDataset.compute.status("default")
```

**Expected behavior**
`ProcedureDataset.compute.status()` should work and return results. If `default=False` `DataFrame.value_counts()` should be checked to ensure that None's are correctly counted.

**Additional context**
The issue is [here](https://github.com/MolSSI/QCFractal/blob/master/qcfractal/interface/collections/collection.py#L585), we need a function like the following to apply:

```python
def get_status(item):
    try:
        return item.status.value
    except:
        return None
    
status = ds.df.apply(lambda col: col.apply(get_status))
```

@ChayaSt

dgasmith:
## Changelog description
Fixes a bug where `ProcedureDataset.status()` failed for specifications where only a subset was computed.

## Status

- [x] Code base linted
- [x] Ready to go

### DIFF
--- a/qcfractal/interface/collections/collection.py
+++ b/qcfractal/interface/collections/collection.py
@@ -584,7 +584,7 @@ class BaseProcedureDataset(Collection):
             def get_status(item):
                 try:
                     return item.status.value
-                except:
+                except AttributeError:
                     return None
 
             # apply status by column then by row
--- a/qcfractal/interface/collections/collection.py
+++ b/qcfractal/interface/collections/collection.py
@@ -581,8 +581,15 @@ class BaseProcedureDataset(Collection):
                 for spec in specs:
                     list_specs.append(self.query(spec))
 
+            def get_status(item):
+                try:
+                    return item.status.value
+                except:
+                    return None
+
             # apply status by column then by row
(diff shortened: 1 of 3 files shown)


qcfractal/interface/collections/collection.py
```diff
@@ -581,8 +581,15 @@ def status(
...
 
+            def get_status(item):
+                try:
+                    return item.status.value
+                except:
+                    return None
+
             # apply status by column then by row
-            df = self.df[list_specs].apply(lambda col: col.apply(lambda entry: entry.status.value))
+            df = self.df[list_specs].apply(lambda col: col.apply(get_status))
```
mattwelborn: `lambda entry: entry.status.__dict__.get("value", None)`?

dgasmith: Doesn't work when `entry` is NaN.

PR #525, opened 2019-12-21
Docker: fix entrypoint of openff worker

mattwelborn:
## Description
This PR fixes `ENTRYPOINT` of the qcarchive_worker_openff worker. (Conda and Docker are not friends.)

## Changelog description
(None needed)

## Status

- [x] Code base linted
- [x] Ready to go


dgasmith:
Rebase this as well.

doaa-altarawy:
Just a comment here. In general, we don't need conda environments in docker containers because the global conda is our environment. This simplifies things a lot.

dgasmith:
In this case we do since we want the conda image to be identical to the Conda-env file we ship around. It isn't possible to remint conda default to an env file (which may change base python etc) without some theatrics.

PR #532, opened 2020-01-02
Fixes Pytest Coverage

dgasmith:
## Description
Coverage was bumped to 5.0 which now uses a SQL database rather than a file for `.coverage`. We switch to `--parallel-mode` here to build a new database per process which is then automatically merged.

I have also deleted the examples folder as it is noticeably out of data and we have similar examples many places in the documentation.

## Changelog description
Fixes a testing subprocess routine when coverage is enabled for coverage 5.0 breaking changes.

## Status

- [x] Code base linted
- [x] Ready to go


mattwelborn:
Will this affect engine or elemental?

dgasmith:
It would only effect engine if you ran coverage on your CLI tests which you do not appear to do.

PR #536, opened 2020-01-03
[WIP] Dataset: don't pull records before save if no records were changed.

mattwelborn:
## Description
This PR allows for updating dataset metadata without pulling and re-writing records. This is especially helpful for large datasets like QM9. 

## Changelog description
Improved the performance of `Dataset.save()` if no new records have been added.

## Status

- [x] Code base linted
- [ ] Ready to go

### DIFF
--- a/qcfractal/tests/test_storage.py
+++ b/qcfractal/tests/test_storage.py
@@ -1321,3 +1321,53 @@ def test_collections_include_exclude(storage_socket):
     storage_socket.del_collection(collection=collection, name=name)
     storage_socket.del_collection(collection=collection, name=name2)
     storage_socket.del_molecules(mol_insert["data"])
+
+
+def test_collections_metadata_update(storage_socket):
+
+    collection = "Dataset"
+    name = "Dataset_tcmu"
+
+    # Add two waters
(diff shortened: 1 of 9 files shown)


dgasmith:
Looks like tests are failing due to the first changes.

qcfractal/tests/test_storage.py
```diff
@@ -1321,3 +1321,53 @@ def test_collections_include_exclude(storage_socket):
...
+    assert len(ret["data"]) == 1
+    assert set(ret["data"][0].keys()) == include
+    assert len(ret["data"][0]["records"]) == 2
+
+    ret = storage_socket.add_collection(db2, overwrite=True)
+    ret = storage_socket.get_collections(collection=collection, name=name, include=include)
+    assert ret["meta"]["success"] is True
+    assert len(ret["data"]) == 1
+    assert set(ret["data"][0].keys()) == include
+    assert len(ret["data"][0]["records"]) == 2
```
mattwelborn: I think this test isolates how this fails. When we call add_collection without "records", records somehow get deleted. I would guess this relates to deep sqlalchemy magic and the fact that "records" is a computed column.

doaa-altarawy:
No work is done on this anymore, so closing it.

PR #543, opened 2020-01-14
Use versions to correctly detect and upgrade db, closes #509

### LINKED ISSUE #509: Upgrade pathways
**Describe the bug**
This is a subtle bug due to the interaction between SQLAlchemy, our version checker, and the upgrade comment. At the moment if you try to boot a server with an outdated version of the database SQLAlchemy will create new tables before the database version check comes into play since we call `Base.metadata.create_all(self.engine)` *before* `self.check_lib_versions()`.

**To Reproduce**
1) Initialize a database
2) Upgrade QCFractal
3) Try to start a server which will fail due to mismatch versions (SQAlchemy builds new tables and columns)
4) Upgrade the server (Fails since new tables and columns are built)

**Expected behavior**
Trying to start the server with an outdated database should not update the database schema.

**Additional context**

@muammar @doaa-altarawy

doaa-altarawy:
## Description
Use versions to correctly detect and upgrade db. This was not implemented yet, although version tables existed.

Different test scenarios:
  - Init a new fractal-server, then start  --> should work
  - Init a new fractal-server, run upgrade, then start --> should work
  - Existing fractal-server, then pip update qcfractal, then start server --> won't start and give msg to upgrade
  - Existing fractal-server, then pip update qcfractal, run upgrade, then start server --> should work

Using a newer DB with older QCFractal version is not supported, and both need to be updated.

## Questions: 
 Any other test scenarios you can think ok?

closes issue #509 

## Changelog description
 No change needed.

## Status
- [x] Code base linted
- [ ] Ready to go
`

### DIFF
--- a/qcfractal/storage_sockets/sqlalchemy_socket.py
+++ b/qcfractal/storage_sockets/sqlalchemy_socket.py
@@ -66,6 +66,9 @@ from .models import Base
 if TYPE_CHECKING:
     from ..services.service_util import BaseService
 
+# for version checking
+import qcelemental, qcfractal, qcengine
+
 _null_keys = {"basis", "keywords"}
 _id_keys = {"id", "molecule", "keywords", "procedure_id"}
 _lower_func = lambda x: x.lower()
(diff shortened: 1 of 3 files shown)


qcfractal/storage_sockets/sqlalchemy_socket.py
```diff
@@ -2672,14 +2684,15 @@ def _copy_users(self, record_list: Dict):
     def check_lib_versions(self):
         """Check the stored versions of elemental and fractal"""
 
+        # check if versions table exist
+        if not self.engine.dialect.has_table(self.engine, "versions"):
+            return None
```
dgasmith: Why doesn't this cause errors when creating a new database and connecting for the first time?

doaa-altarawy: This line checks if the table is created, and won't cause an error because there is no attempt to connect to any table or any kind of queries. This is used when you want to check the DB version, but you don't want to use ORM when tables are not created yet. So this line is important in the case of: 
`qcfractal-server init`, then `qcfractal-server start`

PR #544, opened 2020-01-14
Backup/Restore CLI

### LINKED ISSUE #510: CLI Database Backup/Restore
**Is your feature request related to a problem? Please describe.**
To make it easier to backup and restore the database CLI solutions should be implemented.

**Describe the solution you'd like**
`qcfractal-server backup file.backup` to backup the database. Automatically saving snapshots should be considered, but seems overkill.

`qcfractal-server restore file.backup` to restore from a file. Would require the usual equivalence test of the best folder from the CLI (similar to resetting the folder) in order to overwrite the database.

@doaa-altarawy @mattwelborn What do you think?

dgasmith:
## Description
Provides backup and restore command to the `qcfractal-server` CLI in order to... backup and restore databases. This is only tested locally at the moment as a unit test would be quite involved.

Closed #510.

## Changelog description
Provides backup and restore functionality from the `qcfractal-server` CLI.

## Status

- [x] Code base linted
- [x] Ready to go

### DIFF
--- a/qcfractal/cli/qcfractal_server.py
+++ b/qcfractal/cli/qcfractal_server.py
@@ -177,7 +177,11 @@ def parse_args():
 
     # Restore
     restore = subparsers.add_parser("restore", help="Restores the database from a backup file.")
+    restore.add_argument("filename", default=None, type=str, help="The filename to restore from.")
     restore.add_argument("--base-folder", **FractalConfig.help_info("base_folder"))
+    restore.add_argument(
+        "--force", action="store_true", help="If True, do not ask if the user wishes to delete the current database."
+    )
 
     ### Move args around
     args = vars(parser.parse_args())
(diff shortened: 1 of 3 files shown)


qcfractal/cli/qcfractal_server.py
```diff
@@ -481,6 +486,53 @@ def server_dashboard(args, config):
     app.run_server(debug=True)
 
 
+def server_backup(args, config):
+    psql = standard_command_startup("backup", config)
+
+    print("\n>>> Starting backup, this may take several hours for large databases (100+ GB)...")
```
mattwelborn: Might be helpful to print the user's database size here.

qcfractal/postgres_harness.py
```diff
@@ -338,6 +343,54 @@ def init_database(self) -> None:
...
+        if filename is None:
+            filename = os.path.realpath(f"{self.config.database.database_name}.bak")
+
+        # fmt: off
+        cmds = [
+            shutil.which("pg_dump"),
+            "-p", str(self.config.database.port),
+            "-d", self.config.database.database_name,
+            "-Fc", # Custom postgres format, fast
+            "--file", filename
```
mattwelborn: Consider doing this in a faux-atomic way to avoid writing incomplete backups during failures. For example, write the backup to a temporary file, then move it to `filename` when done.

qcfractal/cli/qcfractal_server.py
```diff
@@ -149,10 +161,28 @@ def parse_args():
...
+        help="The filename to dump the backup to, defaults to 'database_name.bak'.",
+    )
+    backup.add_argument("--base-folder", **FractalConfig.help_info("base_folder"))
+
+    # Restore
+    restore = subparsers.add_parser("restore", help="Restores the database from a backup file.")
+    restore.add_argument("filename", default=None, type=str, help="The filename to restore from.")
+    restore.add_argument("--base-folder", **FractalConfig.help_info("base_folder"))
+    restore.add_argument(
+        "--force", action="store_true", help="If True, do not ask if the user wishes to delete the current database."
```
doaa-altarawy: Since we are using `--overwrite-config` to overwrite the installation, would it be better to also use `--overwrite-db` or just `--overwrite` here?
The word overwrite has no ambiguity.

qcfractal/cli/qcfractal_server.py
dgasmith: Can do.

qcfractal/postgres_harness.py
dgasmith: On the fence on this one. I like it because the user can see that there is something actually happening, but in rare cases in can result in incomplete files.

qcfractal/cli/qcfractal_server.py
dgasmith: Since we always overwrite the `db` in the current setup I differentiated this with the word `--force` here to skip the check. This was mostly added so that I could quickly tests things and should perhaps be removed at this point. Thoughts?

dgasmith: I have removed this option for now.

qcfractal/postgres_harness.py
dgasmith: I have left the current version for now.

mattwelborn: You could write a file in the same directory called `f"{filename}.in_progress"` and then move it when done.

dgasmith: Good point, added.

PR #552, opened 2020-01-28
Changes to set default driver for running properites computation

ahurta92:
## Description
The purpose of the PR is to be able to set a Dataset's default driver to something other than the default "energy".  I was hoping to compute dipole moments with nwchem on my Dataset.  In order to do so, I need to use the "properties" driver.  These changes allow me to do that.  It also allows get_values to return energy values when the driver is "properties".  As a note, I wasn't able to set the default driver using ds.data.default_driver="properties" because the Datamodel class is immutable.

## Changelog description
Creates set_default_driver() function to be able to set the default driver on Dataset collections.  

## Status

- [ ] Code base linted
- [ ] Ready to go


qcfractal/interface/collections/dataset.py
```diff
@@ -700,7 +700,7 @@ def _get_native_values(
         DataFrame
             A DataFrame of the queried parameters
         """
-        au_units = {"energy": "hartree", "gradient": "hartree/bohr", "hessian": "hartree/bohr**2"}
+        au_units = {"energy": "hartree", "gradient": "hartree/bohr", "hessian": "hartree/bohr**2","properties","hartree"}
```
dgasmith: Needs to be `"properties": "hartree"`

ahurta92: Yeah sure!  I'm guessing you want me to add the test to test_dataset.py

dgasmith:
Yes, please!

bennybp:
Datasets have radically changed in v0.50, and most defaults are no longer settable in a singlepoint dataset. I would be happy to talk and go over the new interface if you would like!

PR #556, opened 2020-02-04
Storage: canonicalize molecular formula before query

mattwelborn:
## Description
This PR sorts chemical formula to match qcel's order before querying the database. Now, e.g. searching for "NH3" will work.

I implemented the canonicalization code inside storage_socket, which is definitely the wrong place. Unfortunately, I think this should go in qcel?

## Changelog description
Molecule queries filtered on molecular formula no longer depend on the order of elements.

## Status

- [x] Code base linted
- [ ] Ready to go

### DIFF
--- a/qcfractal/storage_sockets/sqlalchemy_socket.py
+++ b/qcfractal/storage_sockets/sqlalchemy_socket.py
@@ -723,33 +723,14 @@ class SQLAlchemySocket:
         return ret
 
     def get_molecules(self, id=None, molecule_hash=None, molecular_formula=None, limit: int = None, skip: int = 0):
-        def canonicalize_formula(string: str) -> str:
-            """Converts a chemical formula to alphabetical order, matching behavior in qcel"""
-            import re
-            from collections import defaultdict
-
-            matches = re.findall("[A-Z][^A-Z]*", string)
-            count = defaultdict(int)
-            for match in matches:
-                match_n = re.match("(\D+)(\d*)", match)
-                assert match_n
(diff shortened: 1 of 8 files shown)


qcfractal/storage_sockets/sqlalchemy_socket.py
```diff
@@ -722,6 +722,33 @@ def add_molecules(self, molecules: List[Molecule]):
         return ret
 
     def get_molecules(self, id=None, molecule_hash=None, molecular_formula=None, limit: int = None, skip: int = 0):
+        def canonicalize_formula(string: str) -> str:
+            """Converts a chemical formula to alphabetical order, matching behavior in qcel"""
+            import re
+            from collections import defaultdict
+
+            matches = re.findall("[A-Z][^A-Z]*", string)
```
dgasmith: @loriab Can you verify our regex foo?

qcfractal/tests/test_storage.py
```diff
@@ -1193,18 +1193,56 @@ def test_mol_pagination(storage_socket):
...
+        "water_dimer_minima.psimol",
+    ]
+    total = len(mol_names)
+    molecules = []
+    for mol_name in mol_names:
+        mol = ptl.data.get_molecule(mol_name)
+        molecules.append(mol)
+
+    inserted = storage_socket.add_molecules(molecules)
+    try:
```
dgasmith: Good point on the try/finally to ensure cleanup.

PR #557, opened 2020-02-04
[DNM] CI debug

mattwelborn:

### DIFF
--- a/qcfractal/interface/models/tests/test_hashes.py
+++ b/qcfractal/interface/models/tests/test_hashes.py
@@ -50,6 +50,7 @@ def test_molecule_geometry_canary_hash(geom):
 
 ## Keyword Set hash
 
+# test comment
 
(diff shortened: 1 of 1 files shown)


dgasmith:
Looks like the water hash changed after spare QCElemental Molecule.

mattwelborn:
> Looks like the water hash changed after spare QCElemental Molecule.

Does this mean we need to data-migrate all the hashes in QCA? Are we using these new hashes in production now?

dgasmith:
This isn't an issue with the hash. Its an order of operations issue where the test performs a different rounding + orientation scheme. Of course no hashing is going to be idempotent with respect to repeated canonical orientation and geometry rounding. This isn't a huge deal as it doesn't invalidate our current hashes, but is somewhat concerning for trying to build collisions with respect to old hashes from the same data.

dgasmith:
Handled in #561.

PR #564, opened 2020-02-08
Services: ensure protocols from are passed to procedures, fixes #563

### LINKED ISSUE #563: Torsion/GridOptimization Protocol Tracking
**Describe what you are trying to achieve**
We need to ensure optimization and result protocols are correctly tracked through services. A quick glance that results passes through correctly while geometry optimization protocols are not.

dgasmith:
## Changelog description
Fixes a bug where optimization protocols were not respected during torsiondrives and grid optimizations.

## Status

- [x] Code base linted
- [x] Ready to go

### DIFF
--- a/devtools/conda-envs/adapters.yaml
+++ b/devtools/conda-envs/adapters.yaml
@@ -48,7 +48,7 @@ dependencies:
 
 #   QCArchive includes
   - qcengine >=0.11.0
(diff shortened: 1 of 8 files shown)


dgasmith:
I did that as well. Seems to be test instabilities? Let’s see if this passes.

Issue #562, opened 2020-02-07
Restraint indices out of bounds in Trivalent Nitrogen Set 3


btjanaka:
**Describe the bug**

I am trying to access the atoms involved in the optimizations in the "OpenFF Trivalent Nitrogen Set 3" dataset. I am able to access the indices, and I am able to access the molecule, but I find that the indices are out of bounds; i.e. the molecule seems to be lacking atoms. For instance, one of the molecules might have restraint indices [0, 13, 2, 1] but the molecule would only have 9 atoms, whereas it would need at least 14 here (due to zero-based indexing).

I have tried converting the SMILES for the molecule to Openey's oechem.OEMol and accessing the atoms in the resulting molecule, but the indices do not seem to match up with those in the qcelemental molecule. I think the atom at the first restraint index should be the trivalent nitrogen, but this is not always the case when using the OEMol.

**To Reproduce**

The following code prints out the molecules that I have been having issues with.

```
import numpy as np
import qcportal as ptl
from openeye import oechem, oedepict

# Loading Data
client = ptl.FractalClient()
ds = client.get_collection('GridOptimizationDataset',
                           "OpenFF Trivalent Nitrogen Set 3")
molecule_ids = [ds.get_entry(index).initial_molecule for index in ds.df.index]
molecules = client.query_molecules(molecule_ids)
smiles_strings = [
    ds.get_entry(index).attributes["canonical_explicit_hydrogen_smiles"]
    for index in ds.df.index
]
grid_opt_record_ids = [
    ds.get_entry(index).object_map['default'] for index in ds.df.index
]
grid_opt_records = client.query_procedures(grid_opt_record_ids)

# Looking at the restraint indices in each record, and seeing whether they are
# available in the molecule
for record, molecule, smiles in zip(grid_opt_records, molecules,
                                    smiles_strings):
    restraint_indices = np.array(record.keywords.scans[0].__dict__['indices'])
    num_atoms = len(molecule.symbols)

    # If any restraint index is out of bounds
    if np.any(restraint_indices >= num_atoms):
        print("--------------------------------------")
        print("GridOptRecord id :", record.id)
        print("SMILES           :", smiles)
        print("Restraint indices:", restraint_indices)
        print("Number of atoms  :", num_atoms)

        mol = oechem.OEMol()
        oechem.OESmilesToMol(mol, smiles)
        idx_to_symbol = dict(
[...]

PR #565, opened 2020-02-11
Return collected records in `query`

btjanaka:
Returning the records is more useful than just returning the specification name.

Addresses #562 

## Description

Previously, the `query` method for dataset objects merely returned the specification name. This was not very useful, as one would expect to receive data they can use after calling `query`. This PR returns the data collected from the server after calling `query`.

## Changelog description

`query` method for datasets now returns collected records

## Status

- [x] Code base linted
- [x] Ready to go


mattwelborn:
Thanks @btjanaka!

PR #575, opened 2020-02-28
add_results and update_results performance imporvement + benchmark files

codtiger:
## Description
Changed the ```add_results``` and ```update_results``` function in ```sqlalchemy_socket.py``` to do only one commit in case of non-repeated inserts. Benchmark file also provided for comparison. The performance comparisons are also attached.
[runs.txt](https://github.com/MolSSI/QCFractal/files/4366282/runs.txt)

Performance(Best Run) Comparison of **source repo**:
Total time / average time

1-**No.Items:    1**
  **Add   :**     15.559/ 15.559
   **Update:**       4.480/  4.480

2- **No.Items:    5**
  **Add   :**          35.250/  7.050
   **Update:**         17.765/  3.553

3- **No.Items:    10**
**Add   :**         65.656/  6.566
**Update:**        39.594/  3.959

4-**No.Items:    25**
**Add   :**        183.625/  7.345
**Update:**         95.575/  3.823

5-**No.Items:    50**
**Add   :**        322.899/  6.458
**Update:**        197.374/  3.947

6-**No.Items:    100**
**Add   :**       607.728/  6.077
**Update:**   380.709/  3.807

7-**No.Items:    500**
**Add   :**  3076.813/  6.154
**Update:**  1977.275/  3.955

8-**No.Items:    1000**
**Add   :**  5956.370/  5.956
**Update:**  3844.968/  3.845

Performance(Best Run) Comparison of **codtiger repo**:
Total time / average time

1-**No.Items:    1**
**Add   :**     6.988/  6.988
**Update:**          5.532/  5.532

2-**No.Items:    5**
**Add   :**    21.154/  4.231
**Update:**     9.198/  1.840

3-**No.Items:    10**
**Add   :**     10    36.742/  3.674
**Update:**     10     9.910/  0.991

4-**No.Items:    25**
**Add   :**     25    82.651/  3.306
**Update:**     25    20.980/  0.839

5-**No.Items:    50**
**Add   :**       165.057/  3.301
**Update:**        34.837/  0.697

6-**No.Items:    100**
**Add   :**       327.490/  3.275
**Update:**        63.879/  0.639

7-**No.Items:    500**
**Add   :**     1635.619/  3.271
**Update:**     288.668/  0.577

8-**No.Items:    1000**
**Add   :**  3242.667/  3.243
**Update:**    625.845/  0.626 

## Changelog description
add_results , update_results performance improvement

## Status

- [ ] Code base linted
- [ ] Ready to go

### DIFF
--- a/benchmarks/bench_result_add.py
+++ b/benchmarks/bench_result_add.py
@@ -67,8 +67,6 @@ for trial in mol_trials:
 
     t = time.time()
     ret = storage.add_results(results)["data"]
-    meta = storage.add_results(results)["meta"]
-    print (meta["n_inserted"])
 
     ttime = (time.time() - t) * 1000
     time_per_mol = ttime / trial
--- a/benchmarks/bench_result_add.py
+++ b/benchmarks/bench_result_add.py
@@ -1,0 +1,72 @@
+import time
+from qcfractal.interface.models.records import ResultRecord
+import qcfractal
+import numpy as np
+import qcelemental as qcel
+
+print("Building and clearing the database...\n")
+db_name = "molecule_tests"
+storage = qcfractal.storage_socket_factory(f"postgresql://localhost:5432/{db_name}")
+storage._delete_DB_data(db_name)
+
+run_tests = False
+mol_trials = [1, 5, 10, 25, 50, 100, 500, 1000]
+# _size = 20
+
+COUNTER_MOL = 0
+
+
+def build_unique_mol():
+    global COUNTER_MOL
+    mol = qcel.models.Molecule(symbols=["He", "He"], geometry=np.random.rand(2, 3) + COUNTER_MOL, validated=True)
+    COUNTER_MOL += 1
+    return mol
+
+
+def create_unique_result():
+    mol = build_unique_mol()
+    ret = storage.add_molecules([mol])["data"]
+    result = ResultRecord(version='1', driver='energy', program='games', molecule=ret[0],
+                          method='test', basis='6-31g')
+
+    return result
+
+
+if run_tests:
+    print("Running tests...\n")
+    # Tests
+    test_res1 = create_unique_result()
+    test_res2 = create_unique_result()
+    test_res3 = create_unique_result()
+
+    # Sequential
+    ret = storage.add_molecules([test_res1, test_res2, test_res3])["data"]
+    assert ret[0] != ret[1]
+    assert ret[1] != ret[2]
+
+    # Duplicates
+    test_mol1 = create_unique_result()
+    ret = storage.add_results([test_res1, test_res2, test_res3])["data"]
+    assert len(ret) == 3
+    assert ret[0] == ret[1]
+    assert ret[1] == ret[2]
+
+    # Duplicates
+    test_res1 = create_unique_result()
+    test_res2 = create_unique_result()
+    ret = storage.add_results([test_res1, test_res2, test_res1])["data"]
+    assert len(ret) == 3
+    assert ret[0] != ret[1]
+    assert ret[0] == ret[2]
+
+
+print("Running timings...\n")
+for trial in mol_trials:
+    results = [create_unique_result() for x in range(trial)]
+
+    t = time.time()
+    ret = storage.add_results(results)["data"]
+    ttime = (time.time() - t) * 1000
+    time_per_mol = ttime / trial
+
+    print(f"{trial:6d} {ttime:9.3f} {time_per_mol:6.3f}")
--- a/benchmarks/bench_result_add.py
+++ b/benchmarks/bench_result_add.py
@@ -40,7 +40,7 @@ if run_tests:
     test_res3 = create_unique_result()
 
     # Sequential
-    ret = storage.add_molecules([test_res1, test_res2, test_res3])["data"]
+    ret = storage.add_results([test_res1, test_res2, test_res3])["data"]
     assert ret[0] != ret[1]
     assert ret[1] != ret[2]
 
@@ -83,5 +83,4 @@ for trial in mol_trials:
     trial = len(results)
 
     print(f"update: {trial:6d} {ttime:9.3f} {time_per_mol:6.3f}")
-    print()
 
--- a/benchmarks/bench_result_add.py
+++ b/benchmarks/bench_result_add.py
@@ -67,12 +67,14 @@ for trial in mol_trials:
 
     t = time.time()
     ret = storage.add_results(results)["data"]
+    meta = storage.add_results(results)["meta"]
+    print (meta["n_inserted"])
 
     ttime = (time.time() - t) * 1000
     time_per_mol = ttime / trial
 
     print(f"add   : {trial:6d} {ttime:9.3f} {time_per_mol:6.3f}")
-
+    
     for r, rid in zip(results, ret):
         r.__dict__["id"] = rid
 
@@ -83,4 +85,5 @@ for trial in mol_trials:
     trial = len(results)
 
     print(f"update: {trial:6d} {ttime:9.3f} {time_per_mol:6.3f}")
+    print()
 
--- a/benchmarks/bench_result_add.py
+++ b/benchmarks/bench_result_add.py
@@ -33,6 +33,13 @@ def create_unique_result():
 
     return result
 
+mol1 = build_unique_mol()
+ret_mol = storage.add_molecules([mol1])["data"]
+result1 = ResultRecord(version='1', driver='energy',program='games',molecule=ret_mol[0], method='test', basis="")
+result2 = ResultRecord(version='1', driver='energy',program='games',molecule=ret_mol[0], method='test', basis=None)
+ret = storage.add_results([result1])["meta"]
+ret = storage.add_results([result2])["meta"]
+print ("number of inserted {}".format(ret["n_inserted"]))
 
 if run_tests:
     print("Running tests...\n")
@@ -60,6 +67,12 @@ if run_tests:
     assert len(ret) == 3
     assert ret[0] != ret[1]
     assert ret[0] == ret[2]
+
+    # Mutliple Adds of same results
+    test_res3 = create_unique_result()
+    ret1 = storage.add_results([test_res1, test_res2, test_res3])["data"]
+    ret2 = storage.add_results([test_res1, test_res2, test_res3])["data"]
+    assert ret1[0] == ret2[0]
     
 print("Running timings for add and update...\n")
 for trial in mol_trials:
--- a/benchmarks/bench_result_add.py
+++ b/benchmarks/bench_result_add.py
@@ -11,7 +11,7 @@ db_name = "molecule_tests"
 storage = qcfractal.storage_socket_factory(f"postgresql://localhost:5432/{db_name}")
 storage._delete_DB_data(db_name)
 
-run_tests = False
+run_tests = True
 mol_trials = [1, 5, 10, 25, 50, 100, 500, 1000]
 # _size = 20
 
@@ -47,8 +47,8 @@ if run_tests:
     assert ret[1] != ret[2]
 
     # Duplicates
-    test_mol1 = create_unique_result()
-    ret = storage.add_results([test_res1, test_res2, test_res3])["data"]
+    test_res1 = create_unique_result()
+    ret = storage.add_results([test_res1, test_res1, test_res1])["data"]
     assert len(ret) == 3
     assert ret[0] == ret[1]
     assert ret[1] == ret[2]
--- a/benchmarks/bench_result_add.py
+++ b/benchmarks/bench_result_add.py
@@ -26,6 +26,7 @@ def build_unique_mol():
 
 
 def create_unique_result():
+
     mol = build_unique_mol()
     ret = storage.add_molecules([mol])["data"]
     result = ResultRecord(version='1', driver='energy', program='games', molecule=ret[0],
--- a/benchmarks/bench_result_add.py
+++ b/benchmarks/bench_result_add.py
@@ -1,6 +1,8 @@
 import time
 from qcfractal.interface.models.records import ResultRecord
+from qcfractal.interface.models import KeywordSet
 import qcfractal
+import qcfractal.interface as ptl
 import numpy as np
 import qcelemental as qcel
 
@@ -58,8 +60,7 @@ if run_tests:
     assert len(ret) == 3
     assert ret[0] != ret[1]
     assert ret[0] == ret[2]
-
-
+    
 print("Running timings for add and update...\n")
 for trial in mol_trials:
     results = [create_unique_result() for x in range(trial)]
--- a/benchmarks/bench_result_add.py
+++ b/benchmarks/bench_result_add.py
@@ -33,13 +33,6 @@ def create_unique_result():
 
     return result
 
-mol1 = build_unique_mol()
-ret_mol = storage.add_molecules([mol1])["data"]
-result1 = ResultRecord(version='1', driver='energy',program='games',molecule=ret_mol[0], method='test', basis="")
-result2 = ResultRecord(version='1', driver='energy',program='games',molecule=ret_mol[0], method='test', basis=None)
-ret = storage.add_results([result1])["meta"]
-ret = storage.add_results([result2])["meta"]
-print ("number of inserted {}".format(ret["n_inserted"]))
 
 if run_tests:
     print("Running tests...\n")
@@ -73,6 +66,8 @@ if run_tests:
     ret1 = storage.add_results([test_res1, test_res2, test_res3])["data"]
     ret2 = storage.add_results([test_res1, test_res2, test_res3])["data"]
     assert ret1[0] == ret2[0]
+    assert ret1[1] == ret2[1]
+    assert ret1[2] == ret2[2]
     
 print("Running timings for add and update...\n")
 for trial in mol_trials:
--- a/benchmarks/bench_result_add.py
+++ b/benchmarks/bench_result_add.py
@@ -60,13 +60,28 @@ if run_tests:
     assert ret[0] == ret[2]
 
 
-print("Running timings...\n")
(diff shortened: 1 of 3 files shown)


qcfractal/storage_sockets/sqlalchemy_socket.py
```diff
@@ -1138,9 +1138,12 @@ def add_results(self, record_list: List[ResultRecord]):
...
-        result_ids = []
+        result_ids = ["placeholder"] * len(record_list)
+
+        results_list = []
+        results_idx = []
         with self.session_scope() as session:
-            for result in record_list:
+            for i, result in enumerate(record_list):
 
                 doc = session.query(ResultORM).filter_by(
```
dgasmith: This doesn't work for duplicates added in the query since it isn't in the database yet.

qcfractal/storage_sockets/sqlalchemy_socket.py
```diff
@@ -1184,25 +1196,36 @@ def update_results(self, record_list: List[ResultRecord]):
         -------
             number of records updated
         """
+        query_ids = [res.id for res in record_list]
+        duplicates = len (query_ids) != len (set(query_ids))
```
dgasmith: Please use the black formatter, `make format` will automatically apply this.

qcfractal/storage_sockets/sqlalchemy_socket.py
```diff
@@ -1184,25 +1196,36 @@ def update_results(self, record_list: List[ResultRecord]):
...
 
-            if result.id is None:
-                self.logger.error("Attempted update without ID, skipping")
-                continue
-            with self.session_scope() as session:
+            # try:
+            updated_count = 0
+            for result in record_list:
+
+                if result.id is None:
```
dgasmith: This should not be checked above when you find all of the query ids.

qcfractal/storage_sockets/sqlalchemy_socket.py
```diff
@@ -1153,15 +1156,24 @@ def add_results(self, record_list: List[ResultRecord]):
...
                 else:
                     id = str(doc.first().id)
                     meta["duplicates"].append(id)  # TODO
                     # If new or duplicate, add the id to the return list
-                    result_ids.append(id)
+                    result_ids[i] = id
+
+            session.add_all(results_list)
+            session.commit()
+            for i, res in enumerate(results_list):
```
dgasmith: The changes to `add_results` are not faster since we will need to query every loop which is the expensive part.

dgasmith:
As discussed please post the speed ups for master and your current branch.

benchmarks/bench_result_add.py
```diff
@@ -0,0 +1,87 @@
...
+
+if run_tests:
+    print("Running tests...\n")
+    # Tests
+    test_res1 = create_unique_result()
+    test_res2 = create_unique_result()
+    test_res3 = create_unique_result()
+
+    # Sequential
+    ret = storage.add_molecules([test_res1, test_res2, test_res3])["data"]
```
dgasmith: Your tests are adding molecules and not results. Please update *and run them* before saying the PR is ready to go.

qcfractal/storage_sockets/sqlalchemy_socket.py
dgasmith:
```python
cond1 = [Mymodel.basis=='A', Mymodel.driver=='B']
cond1 = [Mymodel.basis=='C', Mymodel.driver=='D']
Mymodel.query.filter(or_(and_(*cond1), and_(*cond2)))
```

This may work to query all multi-indices at the same time. Thanks @doaa-altarawy!

qcfractal/storage_sockets/sqlalchemy_socket.py
codtiger: Formatted the code.

qcfractal/storage_sockets/sqlalchemy_socket.py
codtiger: Not sure on this one, since it was in the original code, and we are still iterating over all the results. Shall I use the dictionary of found items instead?

qcfractal/storage_sockets/sqlalchemy_socket.py
codtiger: Thanks for your input @doaa-altarawy ! I tried using `tuple_` and `in_` first which seemed more convenient, but had to fall back to this solution since `in_` does not translate `None` to `IS NULL`. The change is brought in the new code, along with new comparison.

qcfractal/storage_sockets/sqlalchemy_socket.py
codtiger: Changed the code to address the results that were added before.

benchmarks/bench_result_add.py
codtiger: Corrected in the new commit.

benchmarks/bench_result_add.py
```diff
@@ -0,0 +1,88 @@
...
+    test_res2 = create_unique_result()
+    test_res3 = create_unique_result()
+
+    # Sequential
+    ret = storage.add_results([test_res1, test_res2, test_res3])["data"]
+    assert ret[0] != ret[1]
+    assert ret[1] != ret[2]
+
+    # Duplicates
+    test_mol1 = create_unique_result()
```
dgasmith: This isn't correctly testing anything. Please update this test to have minimal tests and conflicts.

qcfractal/storage_sockets/sqlalchemy_socket.py
dgasmith: Yea the choice for keywords to be a true `NULL` may need to be revisited in the future. We may want a "blank" Keywords entry that effectively returns `NULL`

qcfractal/storage_sockets/sqlalchemy_socket.py
```diff
@@ -1135,33 +1135,66 @@ def add_results(self, record_list: List[ResultRecord]):
...
+                    meta["duplicates"].append(id_)  # TODO
                     # If new or duplicate, add the id to the return list
-                    result_ids.append(id)
+                    result_ids[i] = id_
+
+            session.add_all(results_list)
+            session.commit()
+            for i, res in enumerate(results_list):
+                # print (type(new_records[i]))
+                result_ids[new_record_idx[i]] = str(res.id)
```
dgasmith: Ok, this is starting to look good. Can you clean up the print statements?

qcfractal/storage_sockets/sqlalchemy_socket.py
dgasmith: I mean I would check and filter these before you open a session.

dgasmith:
From looking at your tests there is a chance that your add with duplicates does not work correctly. Please do provide reasonable short tests in the `test_benchmark` and show speedups.

benchmarks/bench_result_add.py
```diff
@@ -0,0 +1,88 @@
...
+    print("Running tests...\n")
+    # Tests
+    test_res1 = create_unique_result()
+    test_res2 = create_unique_result()
+    test_res3 = create_unique_result()
+
+    # Sequential
+    ret = storage.add_results([test_res1, test_res2, test_res3])["data"]
+    assert ret[0] != ret[1]
+    assert ret[1] != ret[2]
```
dgasmith: Lets go back to the requirements here:
1) Unique objects require unique ID's (test on 45)
2) Objects can be duplicated in the add (test on 51/59).
3) Repeated objects have the same ID in multiple adds (no test).

So please add the following:

```python
    ret1 = storage.add_results([test_res1, test_res2, test_res3])["data"]
    ret2 = storage.add_results([test_res1, test_res2, test_res3])["data"]
    assert ret1[0] == ret2[0]
    ...
```

qcfractal/storage_sockets/sqlalchemy_socket.py
```diff
@@ -1176,24 +1176,33 @@ def add_results(self, record_list: List[ResultRecord]):
...
+                    result_ids[i] = str(doc.id)
 
             session.add_all(results_list)
             session.commit()
-            for i, res in enumerate(results_list):
-                # print (type(new_records[i]))
-                result_ids[new_record_idx[i]] = str(res.id)
+            meta["duplicates"] = [str(doc.id) for doc in meta["duplicates"]] 
+            j, k = 0, 0
+            for i, _ in enumerate(result_ids):
```
dgasmith: The previous method seemed much cleaner. Can you elaborate what you are trying to accomplish here?

benchmarks/bench_result_add.py
codtiger: I will add the following test today. Also one thing I noticed today by going through the tests failing on Travis CI server, ```""``` and ```None``` for basis should be equal when added to the database right? It wasn't handled in my code and I had to change it.

qcfractal/storage_sockets/sqlalchemy_socket.py
codtiger: yeah unfortunately had to do some drastic changes, since when you add a new record, that new record should also be added to ```existing_res```, so a repetition of items in ```record_list``` be handled. Since all the new added items and the repeated ones in ```record_list``` do not have any id until ```session.add_all()```, we have to set their ```result_ids``` only after adding all, hence this block would exist.

benchmarks/bench_result_add.py
dgasmith: Whenever you create `ResultRecord` object this should be automatically coerced to the proper shape.

qcfractal/storage_sockets/sqlalchemy_socket.py
dgasmith: I mean the previous dictionary approach was much cleaner. Here you have a lot of offsets that could potentially go wrong depending on edge cases. Storing a list of `indices` and a dictionary of `indices : ids` will allow you to rebuild the `result_ids` in one pass without complex shifts. A list of duplicate `indices` could deliver similar ideas.

codtiger: I mean I thought of your approach and it is actually pretty reasonable and actually adds more to the code readability, but it requires you to use another for loop through the duplicate ```indices``` which might be degrading to the performance even though it should not be much. But then again, I guess you are right that the readability is worth the really small performance gain.

dgasmith: Well, looping through a list of dict will take on the order of ~1 us while each query/check will take on the order of 1ms as shown by your timing test. Anything that doesn't hit the database is completely negligible in cost.

benchmarks/bench_result_add.py
```diff
@@ -0,0 +1,101 @@
...
+
+def create_unique_result():
+    mol = build_unique_mol()
+    ret = storage.add_molecules([mol])["data"]
+    result = ResultRecord(version='1', driver='energy', program='games', molecule=ret[0],
+                          method='test', basis='6-31g')
+
+    return result
+
+mol1 = build_unique_mol()
```
dgasmith: Move this inside the test please and run the formatter over it.

benchmarks/bench_result_add.py
```diff
@@ -0,0 +1,101 @@
...
+ret_mol = storage.add_molecules([mol1])["data"]
+result1 = ResultRecord(version='1', driver='energy',program='games',molecule=ret_mol[0], method='test', basis="")
+result2 = ResultRecord(version='1', driver='energy',program='games',molecule=ret_mol[0], method='test', basis=None)
+ret = storage.add_results([result1])["meta"]
+ret = storage.add_results([result2])["meta"]
```
dgasmith: What are you hoping to test here? Please add an assert statement.

benchmarks/bench_result_add.py
```diff
@@ -0,0 +1,101 @@
...
+    ret = storage.add_results([test_res1, test_res2, test_res1])["data"]
+    assert len(ret) == 3
+    assert ret[0] != ret[1]
+    assert ret[0] == ret[2]
+
+    # Mutliple Adds of same results
+    test_res3 = create_unique_result()
+    ret1 = storage.add_results([test_res1, test_res2, test_res3])["data"]
+    ret2 = storage.add_results([test_res1, test_res2, test_res3])["data"]
+    assert ret1[0] == ret2[0]
```
dgasmith: Test each element, not just the first.

qcfractal/storage_sockets/sqlalchemy_socket.py
```diff
@@ -1135,33 +1135,80 @@ def add_results(self, record_list: List[ResultRecord]):
...
-                    driver=result.driver,
-                    method=result.method,
-                    basis=result.basis,
-                    keywords=result.keywords,
-                    molecule=result.molecule,
+        with self.session_scope() as session:
+            docs = session.query(ResultORM).filter(or_(*conds)).all()
+
+            existing_res = {
+                (doc.program, doc.driver, doc.method, doc.basis, str(doc.keywords), str(doc.molecule)): doc
```
dgasmith: I am not sure why you are casting `keywords` here, None is perfectly valid in dictionary tuple keys.

qcfractal/storage_sockets/sqlalchemy_socket.py
```diff
@@ -1135,33 +1135,80 @@ def add_results(self, record_list: List[ResultRecord]):
...
 
-                doc = session.query(ResultORM).filter_by(
-                    program=result.program,
-                    driver=result.driver,
-                    method=result.method,
-                    basis=result.basis,
-                    keywords=result.keywords,
-                    molecule=result.molecule,
+        with self.session_scope() as session:
+            docs = session.query(ResultORM).filter(or_(*conds)).all()
```
dgasmith: Please change this to only return the fields you are interested in to reduce the cost of the query (the six tuple of interest + the id).

qcfractal/storage_sockets/sqlalchemy_socket.py
```diff
@@ -1135,33 +1135,80 @@ def add_results(self, record_list: List[ResultRecord]):
...
+
+            existing_res = {
+                (doc.program, doc.driver, doc.method, doc.basis, str(doc.keywords), str(doc.molecule)): doc
+                for doc in docs
+            }
+            for i, result in enumerate(record_list):
+                # necessary since result will be used to search through found items
+                if result.basis == "":
```
dgasmith: This should not be need as the `basis` is correctly automatically coerced:

```
>>> ret1 = ResultRecord(driver='energy', program='a', molecule=5, method='test', basis='')
>>> type(ret1.basis)
<class 'NoneType'>
```

benchmarks/bench_result_add.py
codtiger: Deleted this, it was just a sanity test for myself.

benchmarks/bench_result_add.py
codtiger: Done in the new commit.

qcfractal/storage_sockets/sqlalchemy_socket.py
codtiger: as we're looking through a found_items dictionary, the keyword item may return a valid object(instead of ```None```), which is of type ```KeywordSet```, so instead of instantiating a ```KeywordSet``` object, I represent it with the given input string and compare with that.

qcfractal/storage_sockets/sqlalchemy_socket.py
codtiger: Done in the new commit.

qcfractal/storage_sockets/sqlalchemy_socket.py
codtiger: Again, instantiating a ```ResultRecord``` object with the ```result.basis``` that was provided can be the way to go for comparison on the found items dictionary, but this one seems more efficient.

qcfractal/storage_sockets/sqlalchemy_socket.py
dgasmith: This should not be the case snipping from the ORM:
```
    keywords = Column(Integer, ForeignKey("keywords.id"))
    keywords_obj = relationship(KeywordsORM, lazy="select")
```

Plus if the keywords field is a `KeywordSet` casting to a string doesn't give you want you want. The string representation is not guaranteed to be unique.

qcfractal/storage_sockets/sqlalchemy_socket.py
dgasmith: The code requires a list of `ResultRecord` object which will have already correctly coerced the `result.basis`. This seems like an unnecessary `if` statement which is double true since the coercion logic is more complex than shown.

codtiger: moved to if the ```idx``` assignment part.

dgasmith:
@doaa-altarawy Can you please review this when you have time?

qcfractal/storage_sockets/sqlalchemy_socket.py
```diff
@@ -1135,33 +1135,90 @@ def add_results(self, record_list: List[ResultRecord]):
...
-            for result in record_list:
+        result_ids = ["placeholder"] * len(record_list)
+
+        results_list = []
+        existing_res = {}
+        new_record_idx, duplicates_idx = [], []
+        conds = []
+
+        conds = [
+            and_(
```
doaa-altarawy: Why do you need to use `_or` and `_and` ?
The default for `filter` is `and` of all conditions

codtiger: You are totally right, removed this in the new commit.

qcfractal/storage_sockets/sqlalchemy_socket.py
doaa-altarawy: @codtiger

codtiger: Should I pop the results with ```None``` ids out of the list before opening the session? So I only iterate over valid ones?

PR #576, opened 2020-03-02
Pin pint !=0.11

mattwelborn:
## Description
This PR pins pint to avoid v0.11 so that CI will pass. This is intended to be reverted once the problem is either fixed upstream or mitigated in qcel.

## Changelog description
N/A

## Status

- [x] Code base linted
- [x] Ready to go


dgasmith:
Wow... synchronous.

PR #597, opened 2020-05-21
Add ability to query managers from portal

bennybp:
## Description
Adds the ability to query for compute manager information via qcportal.
This is done by adding the appropriate REST models and handlers, then adding
the appropriate function to the client.

Currently, admin access is required, although this may be relaxed in the future. Also, the passwords used by the managers to connect to the fractal instance are scrubbed.

This is required for the dashboard.

Also, I fixed a few of the comments in the REST handlers.

## Changelog description
Adds ability to query manager information from the client

## Status

- [x] Tests
- [x] Code base linted
- [x] Ready to go

### DIFF
--- a/qcfractal/server.py
+++ b/qcfractal/server.py
@@ -18,7 +18,7 @@ import tornado.web
 
 from .extras import get_information
 from .interface import FractalClient
-from .queue import QueueManager, QueueManagerHandler, ServiceQueueHandler, TaskQueueHandler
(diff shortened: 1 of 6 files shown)


qcfractal/server.py
```diff
@@ -291,6 +291,7 @@ def __init__(
             (r"/task_queue", TaskQueueHandler, self.objects),
             (r"/service_queue", ServiceQueueHandler, self.objects),
             (r"/queue_manager", QueueManagerHandler, self.objects),
+            (r"/manager_info", ComputeManagerHandler, self.objects),
```
doaa-altarawy, line 294: without info

PR #606, opened 2020-07-20
Torsioninitmol mapping class replacing table representation

codtiger:
## Description
```torsion_init_mol``` table now has a mapping ORM class assigned in the codebase, rather than a simple table representation.

## Changelog description
```torsion_init_mol``` was represented as a simple table which would be populated every time an insertion or update command was issued to ```torsiondrive_procedure```. Now it has its own mapping class which would make it simpler for manipulation of rows from objects in code. Since the current production database has rows where the ```molecule_id``` and ```torsion_id``` are not unique- hence the existence of duplicate rows - an id for each row as primary key is also added to the table(through alembic).

## Status

- [x] Code base linted
- [x] Ready to go

### DIFF
--- a/qcfractal/alembic/versions/1604623c481a_id_pirmary_key_for_torsion_init_mol.py
+++ b/qcfractal/alembic/versions/1604623c481a_id_pirmary_key_for_torsion_init_mol.py
@@ -17,8 +17,19 @@ depends_on = None
 
 
 def upgrade():
-    op.execute("alter table torsion_init_mol_association add column id serial primary key;")
+
+    op.execute(
+        "DELETE FROM torsion_init_mol_association a USING \
+    (SELECT MIN(ctid) as ctid, torsion_id, molecule_id \
+        FROM torsion_init_mol_association \
+        GROUP BY torsion_id, molecule_id HAVING COUNT(*) > 1 \
+      ) b \
+      WHERE a.torsion_id = b.torsion_id and a.molecule_id = b.molecule_id \
+      AND a.ctid <> b.ctid"
+    )
+
+    op.execute("alter table add primary key (torsion_id, molecule_id)")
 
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by bennybp** [comment]
I am testing out if duplicate rows in that table are errors. I believe they are and can be harmlessly removed, but am testing with a copy of the database to be sure. I will let you soon

File: qcfractal/alembic/versions/1604623c481a_id_pirmary_key_for_torsion_init_mol.py [hunk 1]
```diff
@@ -0,0 +1,35 @@
...
+        "DELETE FROM torsion_init_mol_association a USING \
+    (SELECT MIN(ctid) as ctid, torsion_id, molecule_id \
+        FROM torsion_init_mol_association \
+        GROUP BY torsion_id, molecule_id HAVING COUNT(*) > 1 \
+      ) b \
+      WHERE a.torsion_id = b.torsion_id and a.molecule_id = b.molecule_id \
+      AND a.ctid <> b.ctid"
+    )
+
+    op.execute("alter table add primary key (torsion_id, molecule_id)")
```
bennybp, line 31: Are you missing the table name here? I get a syntax error when running

(also possibly missing from the downgrade)

codtiger: I just saw this...yeah it's missing, I'll pull and reformat the code.

bennybp: This has already been fixed and merged. See the commit below

https://github.com/MolSSI/QCFractal/commit/842b128dd4fb3b93492578008de0969e85e3039a

PR #601, opened 2020-06-18
QCSpec table migration on procedure tables(optimization_procedure, grid_optimization_procedure, torsiondrive_procedure)

codtiger:
## Description
The biggest change comes in the alembic migration file while ```sqlalchemy_socket.py``` has also seen some small changes due to the new migration. A new table ```qc_spec``` and its corresponding ORM class ```QCSpecORM``` have been created and the ```qc_spec``` json column in the procedure tables have been switched to an integer column pointing to the specification in the qc_spec table. A new private function ```_add_spec``` is added to as well to help the addition of specifications. The test files ```test_sqlalchemy.py``` also had to receive minor changes since the construction of procedure ORM class requires construction of a ```QCSpecORM``` object - since its id will be used- beforehand. 

<b>Warning</b>: Two of the changes on ```test_collections.py``` are not related to the changes done in this PR, and are only done to reflect the master changes.

<b>Important</b>: Since ```get_procedures``` had to be changed to accompany the ease of getting and updating a procedure in two lines, some performance testing might be good to check if we can observe heavy query costs or not.
## Changelog description
This is a large change due to the volume of the migrations ready to be exectued. Briefly, a new table ```qc_spec``` has been created, the ```qc_spec``` column of the procedure tables have changed to reference columns to point to a specific row in ```qc_spec``` and also some inevitable changes to the test had to be performed accordingly.

### Benchmarks
| add/get procedure | Before(total / avg) | After (total / avg) | Num trials
| -----| ------- | ------|-------
| **add** | 17.689 / 17.689 | 25.122 / 25.122 | 1
| **get** | 4.936 / 4.936 | 5.984 / 5.984 | 1
| **add**| 103.831/ 10.383 | 125.761/ 12.576 | 10
| **get**| 7.038 / 0.704 | 8.986 /  0.899 | 10
| **add** | 1057.561/ 10.576 | 1162.054/ 11.621 | 100
| **get** | 30.550/  0.306 | 35.585 / 0.356 | 100
| **add** |5483.121/ 10.966 | 5938.493 / 11.877 | 500
|**get** | 148.964 / 0.298 | 164.594/  0.329 | 500
|**add**| 11144.793/ 11.145 | 11877.745/ 11.878 | 1000
|**get**  | 239.103/  0.239 | 296.329/  0.296 | 1000

All the numbers are in milliseconds.
## Status

- [X] Code base linted
- [x] Tests for QCSpecORM class(esp. ```_add_spec```)
[...]

PR #607, opened 2020-07-30
Result table + QCSpec migration + query projection optional join

codtiger:
## Description
Following [this](https://github.com/MolSSI/QCFractal/pull/601), ```results``` table and accordingly ```ResultORM``` has been changed to follow the normalization process of preserving the five fields of ```program, basis, driver, method, keywords``` at one place. Along this change, ```get_query_projection``` is also tweaked with two optional arguments supporting joining on a secondary table.

## Changelog description
The database migration consisting of schema and data migration for pushing all the five tuples to a seperate table and keeping a ```qc_spec``` reference is done in a single alembic update file.(ending in ```result_to_qc_spec_normalization.py```) . ```ResultORM``` class is changed accordingly, with a new column called ```qc_spec``` and the relationship accompanying it. With these changes, necessary changes was brought to ```add_results``` and ```update_results```. Despite the two function changes being similar to their procedure function counterparts, ```get_results``` went down a different route which required a mandatory join on two tables ```results``` and ```qc_spec```. This change is done with two optional keywords added to the ```get_query_projection``` which can be used with a new join table and secondary query conditions(on the join table).

Also, ```db_queries``` consisted of some hand-written queries on ```results``` which used the previously available five fields to construct ```ResultRecord``` objects. These two hand-written queries were also altered to retrieve the correct specification from ```qc_spec``` table.

## Status

- [X] Commenting and providing details
- [X] Code base linted
- [ ] Ready to go


bennybp:
This is being a bit subsumed by a big refactor  happening in `next`. I have a local copy of this branch so will be rebasing/cherry picking these changes into there at some point. So closing this PR.

PR #608, opened 2020-08-03
Dataset protocols

jthorton:
## Description

This will add the ability to pass protocols to the basic dataset compute function instead of using the `client.add_compute`.

Should the protocol also be included in the history along with the program, basis and method? 

## Changelog description

Added the ability to pass protocols to basic dataset compute function.

## Status

- [X] Code base linted
- [ ] Ready to go

### DIFF
--- a/qcfractal/interface/collections/dataset.py
+++ b/qcfractal/interface/collections/dataset.py
@@ -1220,7 +1220,9 @@ class Dataset(Collection):
 
             for i in range(0, len(umols), self.client.query_limit):
                 chunk_mols = umols[i : i + self.client.query_limit]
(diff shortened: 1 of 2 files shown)


qcfractal/interface/collections/dataset.py
```diff
@@ -1644,6 +1648,9 @@ def compute(
             The queue tag to use when submitting compute requests.
         priority : Optional[str], optional
             The priority of the jobs low, medium, or high.
+        protocols: Optional[Dict[str, Any]], optional
+            Protocols for store more or less data per field. Current valid
```
dotsdl, line 1652:
```suggestion
            Protocols for storing more or less data per field. Current valid
```

dotsdl:
I'm not currently a maintainer on this repository, so I can't do anything further. @bennybp is this something you are willing to merge?

PR #610, opened 2020-08-28
Fix task queue index migration

bennybp:
## Description
Fixes a possible issue with a previous migration. Migration fb5bd88ae2f3 was maybe incomplete, causing all future autogenerated migrations to include a fix.

This PR is just an autogenerated alembic migration.

## Status

- [X] Code base linted
- [X] Ready to go


bennybp:
Ok my initial commit was completely wrong and based off an old branch. Should be good now.

bennybp:
Nevermind this migration. Alembic is always detecting that this migration must be done, even after this migration is applied. Something more is going on (although seems to be an alembic issue)

PR #615, opened 2020-09-10
Fix bug in add_results

bennybp:
## Description
Fixes a serious bug in add_results

Before, there would have been a query that returned the entire results table. This was confirmed on a small test server. Now, the code works as expected, returning only duplicate entries from the results table.

The bug is a result of creating the conditions for the query based on an empty list (rather than the input list to the function)

This bug was added in https://github.com/MolSSI/QCFractal/commit/b88f874d43e5075288b3d647570838e76a6efc0a as part of PR https://github.com/MolSSI/QCFractal/pull/575

In addition, some general cleanup on the function was done.

## Status

- [ ] Code base linted
- [x] Ready to go

### DIFF
--- a/qcfractal/storage_sockets/sqlalchemy_socket.py
+++ b/qcfractal/storage_sockets/sqlalchemy_socket.py
@@ -3,7 +3,7 @@ SQLAlchemy Database class to handle access to Pstgres through ORM
 """
 
 try:
-    from sqlalchemy import create_engine, or_, case, func
+    from sqlalchemy import create_engine, and_, or_, case, func
     from sqlalchemy.exc import IntegrityError
     from sqlalchemy.orm import sessionmaker, with_polymorphic
     from sqlalchemy.sql.expression import desc
@@ -1138,19 +1138,21 @@ class SQLAlchemySocket:
         -------
         Dict[str, Any]
             Dict with keys: data, meta
-            Data is the ids of the inserted/updated/existing docs
+            Data is the ids of the inserted/updated/existing docs, in the same order as the
+            input record_list
         """
+
         meta = add_metadata_template()
-        # putting the placeholder for all the ids, since some can be duplicate, and some will only have ids after add operation.
-        result_ids = ["placeholder"] * len(record_list)
 
         results_list = []
-        existing_res = {}
+
+        # Stores indices referring to elements in record_list
         new_record_idx, duplicates_idx = [], []
-        conds = []
+
         # creating condition for a multi-value select
+        # This can be used to query for multiple results in a single query
         conds = [
-            (
+            and_(
                 ResultORM.program == res.program,
                 ResultORM.driver == res.driver,
                 ResultORM.method == res.method,
@@ -1158,7 +1160,7 @@ class SQLAlchemySocket:
                 ResultORM.keywords == res.keywords,
                 ResultORM.molecule == res.molecule,
             )
-            for res in results_list
+            for res in record_list
         ]
 
         with self.session_scope() as session:
@@ -1175,12 +1177,15 @@ class SQLAlchemySocket:
                 .filter(or_(*conds))
                 .all()
             )
+
             # adding all the found items to a dictionary
-            existing_res = {
+            existing_results = {
                 (doc.program, doc.driver, doc.method, doc.basis, doc.keywords, str(doc.molecule)): doc for doc in docs
             }
+
+            # Loop over all (input) records, keeping track each record's index in the list
             for i, result in enumerate(record_list):
-                # constructing an index from record_list to compare against found items(existing_res)
+                # constructing an index from the record compare against items existing_results
                 idx = (
                     result.program,
                     result.driver.value,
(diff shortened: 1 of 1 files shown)


bennybp:
@codtiger Could you take a quick look at this and see if it makes sense? Mostly just the part where the `conds` list is set.

codtiger:
> @codtiger Could you take a quick look at this and see if it makes sense? Mostly just the part where the `conds` list is set.

Almost forgot this bug existed.
I actually found this bug way after the PR was accepted and fixed it just the same way you did, but forgot to report it between other works. So the wrong `conds` construction causes the query on line 1175 to return all items in the table, which is just against the logic of the code.
But there is an issue with the and_ approach and it is related to the query execution speed. Given lots of results, this query would be pretty long and if I remember correctly would damage the speedup reported in PR #575. I believe there should be a middle ground for when all records should be retrieved and when the `and_` conditions be used. Still, the PR you made is the logical error-proof code, though the benchmarks may have to be remeasured.

doaa-altarawy:
@codtiger  That's a good point, what effect do you expect on the benchmark results?

codtiger:
> @codtiger That's a good point, what effect do you expect on the benchmark results?

I pulled Ben's code on the current master repo and ran the benchmark for the current master and the pull request. Here are the results(Numbers in parentheses are average time per item)
No. Add| Master Repo(avg)  | Fix PR(avg) |
-----| ------------- | ------------- |
| 1 | 7.020 (7.020 )|7.001 (7.001)
| 5| 21.503(4.301)  |23.368 (4.674)
| 10| 35.357(3.536)  |39.815 (3.982)
| 25| 83.169(3.327)  |94.793 (3.792)
| 50| 159.902(3.198) |185.988 (3.720)
| 100| 327.652 (3.277 )|375.033 (3.750)
| 500| 1543.917 (3.088 )|1753.992 (3.508)
| 1000| 3165.379 (3.165 )|3571.552 (3.572)

As you can see the fix is slower in terms of add time, specifically over larger inserts. Though, you should note that the benchmark only adds to an already empty database. So overall, after reaching the 1000 item insertion benchmark, there is already ~700 items in the table.
What this tells me, which I remembered from my experiment back when I found out about the bug, is that the fixed code is the way to go. But if the content of `results` table does not take much memory, it is probably better to take all records from the table, rather than querying with multitudes of `and_` clauses wrapped in a `or_`. This probably can be done using a simple if statement that checks the current number of records on database and then make a decision about `conds`. The threshold for such a decision should probably be based on real production environment statistics though.

bennybp:
Oh the results table is way too big. Currently it has about 23.5 million records. A middle ground might be to batch queries together in groups of 10 or so. That would still be an improvement over one-by-one.

I feel there's got to be a better way, though.

codtiger:
> Oh the results table is way too big. Currently it has about 23.5 million records. A middle ground might be to batch queries together in groups of 10 or so. That would still be an improvement over one-by-one.
> 
> I feel there's got to be a better way, though.

This might be a reasonable approach, though I have to point out that I checked the table and it only has an index on the primary key. Probably adding some indexes in prefix order would ease the query + batching.
But if PR #607 was to be merged in the near future, any change to the database schema wouldn't be effective since it will most likely get overwritten by that ```result``` table structure.
But again, I believe this code should be merged. The middle ground can be discussed in a future PR in my opinion.

bennybp:
Ok sounds good. Right now I don't think add_results is called with too many inputs, but that will change with some modifications I am making elsewhere in the code. But I will look to fix that in another PR

bennybp:
I ran some (possibly contrived) benchmarks on a copy of the production database with 23 million rows, and it seems that iterating and querying one-by-one seems to work best for > 200 input queries. Below that, using `IN` seems to be significantly faster, although we are still looking at millisecond difference. So it may not be worth splitting out functionality for <200 and >200 records.

codtiger:
> I ran some (possibly contrived) benchmarks on a copy of the production database with 23 million rows, and it seems that iterating and querying one-by-one seems to work best for > 200 input queries. Below that, using `IN` seems to be significantly faster, although we are still looking at millisecond difference. So it may not be worth splitting out functionality for <200 and >200 records.
> 
> Times below are total times (not per row)
> 
> ```
> Got 1 random rows
> Searching individually: Time = 0.002459287643432617 seconds
> Searching all with AND/OR: Time = 0.002338886260986328 seconds
> Searching all with IN: Time = 0.0010671615600585938 seconds
> 
> Got 10 random rows
> Searching individually: Time = 0.01614665985107422 seconds
> Searching all with AND/OR: Time = 0.0028574466705322266 seconds
> Searching all with IN: Time = 0.0038411617279052734 seconds
> 
> Got 50 random rows
> Searching individually: Time = 0.023630142211914062 seconds
> Searching all with AND/OR: Time = 0.011995553970336914 seconds
> Searching all with IN: Time = 0.004986763000488281 seconds
> 
> Got 100 random rows
> Searching individually: Time = 0.031865596771240234 seconds
> Searching all with AND/OR: Time = 0.01993703842163086 seconds
> Searching all with IN: Time = 0.018311262130737305 seconds
> 
> Got 200 random rows
> Searching individually: Time = 0.04730701446533203 seconds
> Searching all with AND/OR: Time = 0.0415346622467041 seconds
> Searching all with IN: Time = 0.042656660079956055 seconds
> 
> Got 300 random rows
> Searching individually: Time = 0.07261276245117188 seconds
> Searching all with AND/OR: Time = 0.06915664672851562 seconds
> Searching all with IN: Time = 0.07406139373779297 seconds
> 
> Got 400 random rows
> Searching individually: Time = 0.11962199211120605 seconds
> Searching all with AND/OR: Time = 0.1134040355682373 seconds
> Searching all with IN: Time = 0.1081845760345459 seconds
> 
> Got 500 random rows
> Searching individually: Time = 0.39232420921325684 seconds
> Searching all with AND/OR: Time = 0.15880489349365234 seconds
> Searching all with IN: Time = 0.15682029724121094 seconds
> 
> Got 1000 random rows
> Searching individually: Time = 0.5176310539245605 seconds
> Searching all with AND/OR: Time = 0.5051937103271484 seconds
> Searching all with IN: Time = 0.5019245147705078 seconds
> ```

This is pretty good, and also gives me the impression that it is probably better to change the benchmark accordingly to prepare a database at least accurate to a fraction( x10 smaller, for example) before carrying out the tests. I can probably do that in a separate PR.

bennybp:
I think in general we can leave the benchmark for now. It is a bit difficult to set up even a 10% version of the database, as that would still be ~80GB. But it's something we can keep in mind.

I am basically ready to merge this, but I want to wait until travis passes. But we are hitting https://github.com/conda-forge/pint-feedstock/issues/41

Issue #611, opened 2020-08-30
Inserting Tasks is Slow for Large Numbers of Tasks


WardLT:
**Describe the bug**
Very long times to insert large numbers of tasks. 

```
[I 200830 21:30:26 handlers:44] POST: TaskQueue -  Added 1000 tasks.
[I 200830 21:30:26 web:2250] 200 POST /task_queue (::1) **1208480.53ms <---**
[I 200830 21:30:27 web:2250] 200 PUT /queue_manager (::1) 107.64ms
[I 200830 21:30:27 web:2250] 200 PUT /queue_manager (::1) 27.68ms
[I 200830 21:40:26 handlers:44] POST: TaskQueue -  Added 498 tasks.
[I 200830 21:40:26 web:2250] 200 POST /task_queue (::1) 599074.87ms
[I 200830 21:40:29 web_handlers:396] POST: Collections - 1 inserted.
```
**To Reproduce**

The above bug is on a long-running server and I'm not sure what's particular about it's configuration that makes these queries take so long:

- It is sitting on an HPC login node and the database is on a global FS (high latency IO)
- I have ~20k results and 100k molecules

**Expected behavior**

I'll not sure how quick these operations should take, but 1 s/result seems a bit slow.

**Additional context**

Should we revive PR #571 ?

bennybp:
This may have been at least partially improved in PR https://github.com/MolSSI/QCFractal/pull/584
That PR should be available in the next release (scheduled for the end of September).
I have my eye on refactoring that code soon, with hopes of that also making it into that release.

The DB being on a global FS is probably exacerbating the issue, but there may be something else going on as well.

bennybp:
I ran a copy of our production database (with the current release, v0.13.1) under a profiler. I see something similar, although still quite a bit faster than your timings.

The bulk of the time is spent in the psycopg2 `commit` function, as somewhat expected. For inserting 1000 tasks, this is called ~3000 times. This can be improved by batching things together like you did in the PR, so this gives me a baseline for what the latest release does. I am actively working on refactoring that piece as we speak.

The below is for inserting 1000 tasks. I think each task will do 3 separate inserts - molecule, result, task. Each is done separately at the moment, which is not ideal. This database in on a spinning disk.

```
         9123931 function calls (8834162 primitive calls) in 47.867 seconds

   Ordered by: internal time 

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
[...]

PR #620, opened 2020-09-24
Refactoring of procedures

bennybp:
## Description
This a refactoring of the code in the procedures subdirectory. There are a few reasons behind this:

* Reducing the possibly of some database inconsistencies
* Separating the code for creating result objects and task entries
* General documentation and cleanup
* Issue #611 

The main push for this was the realization that there are some inconsistent states in the database - there are entries in the `base_result` table that are incomplete or errored, but with no corresponding entry in the task_queue. The cause of this is still unknown, but this PR adds code for regenerating tasks for existing result records in the database.

The procedures contained in procedures.py were moved to separate files, as they became quite a bit larger. There is still lots of room for improvement in these files, including deduplication, but this is good enough for a first pass.

This PR also adds the ability to modify the tag and priority of existing tasks (if they are in the waiting or error state).

I have not tested new task insertion time, but will before final release of v0.14

## Changelog description
* Adds ability to regenerate missing tasks
* Adds ability to modify tag & priority of existing tasks
* Make task insertion faster

## Status

- [x] Add tests for new functionality
- [x] Add additional logging
- [x] Code base linted
- [x] Ready to go


dotsdl:
Reviewing!

qcfractal/queue/handlers.py
```diff
@@ -65,7 +68,46 @@ def put(self):
...
+            new_tag = body.data.new_tag
+            if body.data.new_priority is None:
+                new_priority = PriorityEnum.NORMAL
+            else:
+                new_priority = PriorityEnum(int(body.data.new_priority))
+
+            for r in result_data:
+                model = build_procedure(r)
+
+                # Only regenerate the task if the base record is not complete
```
doaa-altarawy, line 89: Too much application logic in this PUT method, the REST should not handle all that knowledge about the Queue and it's status. At most it can use one variable to redirect to different socket functions. I suggest moving that logic to other layers

doaa-altarawy (changes requested):

qcfractal/queue/handlers.py
bennybp: Completely agree. However, was trying to minimize changes to that file due to your REST rewriting

doaa-altarawy: This file is fully refactored already in the REST refactor too :)

doaa-altarawy (approved):

qcfractal/interface/client.py
```diff
@@ -1026,6 +1029,8 @@ def modify_tasks(
         operation : str
             The operation to perform on the selected tasks. Valid operations are:
              - `restart` - Restarts a task by moving its status from 'ERROR' to 'WAITING'
+             - `regenerate` - Regenerates a missing task
+             - `modify` - Modify a tasks tag or priority
```
dotsdl, line 1033: Oooh! This is nice! Thanks for adding both of these @bennybp!

qcfractal/tests/test_compute.py
```diff
@@ -112,6 +112,94 @@ def test_task_client_restart(fractal_compute_server):
...
+    res = client.query_procedures(base_ids)
+    assert all(x.status == "INCOMPLETE" for x in res)
+
+
+def test_task_modify(test_server):
+    client = ptl.FractalClient(test_server)
+
+    # Add a single computation
+    mol = ptl.models.Molecule(**{"geometry": [0, 0, 1], "symbols": ["He"]})
+    ret1 = client.add_compute("rdkit", "cookiemonster", "", "energy", None, [mol], tag="test_tag_1", priority=1)
```
dotsdl, line 187: I thought priorities could be one of `{"HIGH", "MEDIUM", "LOW"}`. Has this changed?

dotsdl (approved):
Looks great @bennybp! Thank you for this big advancement; I'm really excited about the ability to modify tasks and regenerate ones that are missing. Just a couple notes here, but looks good!

May want to add changelog entries to `docs/qcfractal/source/changelog.rst` before merge!

qcfractal/tests/test_compute.py
bennybp: https://github.com/MolSSI/QCFractal/blob/e6c108f17beb3c248d6f8302dd9a826f528d708f/qcfractal/interface/models/task_models.py#L40

It's an integer, but there is some conversions happening. Thanks for pointing this out, though, I have been making an effort to consistently use the enums themselves.

bennybp:
> Looks great @bennybp! Thank you for this big advancement; I'm really excited about the ability to modify tasks and regenerate ones that are missing. Just a couple notes here, but looks good!
> 
> May want to add changelog entries to `docs/qcfractal/source/changelog.rst` before merge!

I will do that in the final commit before release (since I've neglected to do it for the other PRs)

qcfractal/queue/handlers.py
bennybp: Not the socket, which would be too low level. There isn't a very good place for it yes (except maybe procedures?) but we are definitely planning on revisiting this.
handlers.py is being modified in another refactor at the moment so it might be best to just leave it for the time being.

dotsdl: Absolutely agree. It has to live somewhere, and I think it's fine where it rests for now.

PR #623, opened 2020-10-07
[WIP] Update versions of python

bennybp:
## Description
Update/Test newer versions of python, now that python 3.9 is released

## Status

- [ ] Code base linted
- [ ] Ready to go


dotsdl:
Thanks @bennybp, I'll take this on.

dotsdl:
It appears that `parsl` conda packages conflict with `dask`. Commenting out `parsl` and the `ipy*` packages (that appear to be recommended for it in the `generate_envs.py` script) yields a conda env that resolves fine.

As for the openff env, [there are no `openforcefield` Python 3.8 packages on the `omnia` channel](https://anaconda.org/omnia/openforcefield/files), so this fails to work when trying to create a Python 3.8 environment. Recommending we stick to Python 3.7 for this environment until things change on the OpenFF end.

As for parsl vs. dask, it seems best to split these out into separate envs so they don't conflict. Thoughts @bennybp? This may also be a transient problem, since we are relying on how dependencies are resolved from the specified constraints of these complex distributed computing frameworks.

bennybp:
On my computer it seems python 3.7 works with the adapter env. I guess we would generally need to stop at 3.7.

I wonder what is causing the parsl/dask conflict that only shows up when python 3.8 is used, but I guess it's a moot point for now.

loriab:
fwiw, `conda create -n asdf parsl dask python=3.8 -c conda-forge` works for me

dotsdl:
> fwiw, `conda create -n asdf parsl dask python=3.8 -c conda-forge` works for me

Hmmm, I suppose it's unlikely to be a pairwise interaction that's creating problems here then. I *think* we're running into issues with the distributed computing frameworks, but it's frankly hard to be sure. A lot of the dependency resolution logging I was seeing with `conda env create -vv -f adapters.yaml` (setting `python =3.8`) suggested to me the conflicts were concentrated around dask and ipyparallel. This is only an impression though.

I still think it may be wise to separate out the adapters into their own environments, since adapters are unlikely to be deployed in the same environment in production, yes?

mattwthompson:
> As for the openff env, there are no openforcefield Python 3.8 packages on the omnia channel, so this fails to work when trying to create a Python 3.8 environment. Recommending we stick to Python 3.7 for this environment until things change on the OpenFF end.

Once OpenMM 7.5 is released, it will support Python 3.8 the rest of the OpenFF stack should be able to follow suit relatively quickly. Not sure a timetable, it may possibly not happen until the `conda-forge` migration is complete.

loriab:
> I still think it may be wise to separate out the adapters into their own environments, since adapters are unlikely to be deployed in the same environment in production, yes?

agreed, the lanes will run faster, and problems easier to isolate.

dotsdl:
@bennybp taking this on; will create a separate PR based on this branch that uses GH Actions instead. As discussed w/ @bennybp, this will simplify the CI approach by cutting out Travis as a separate service.

dotsdl:
We're getting all jobs passing now, but to confirm, we instead want to have this build matrix:
- base: py3.7
- base: py3.8
- adapter_parsl: py3.8
- adapter_dask: py3.8
- adapter_fireworks: py3.8
- openff: py3.7
- dev_head: py3.8

Correct?

bennybp:
I was thinking more of a product between python versions and those conda environments. So one axis is python `3.7`, `3.8` (and eventually `3.9`, and the other axis is `base`, `adapter_parsl`, `adapter_dask`, `adapter_fireworks`, `openff`, and `dev_head`. So there would be 6*2 = 12 tests.

It just seemed awkward to do this in travis, since you can't have both axes be `env`

bennybp:
Looks like parsl doesn't work with python 3.8 (or at least the conda package doesn't). The openff test failed with the occasionally-failing test that I am looking into

bennybp:
Superceded by #628

PR #628, opened 2020-10-21
CI via Github Actions

dotsdl:
## Description

Adds CI via Github Actions.

We are considering switching CI execution from Travis CI to Github Actions. This PR tests this approach. One issue we may need to solve first is how we want to allow some builds to fail if they are known to. [There does not appear to be a great solution to this yet](https://github.com/actions/toolkit/issues/399).

## Changelog description

Added Github Actions continuous integration workflow to repository.

## Status

- [ ] Code base linted
- [ ] Ready to go


dotsdl:
I should note that this branch is based off of #623, as we did a good deal of work updating the conda build environments. We anticipate that will be merged fairly soon into `master`.

dotsdl:
This may work for allowed failures, though from comments it appears it may not work as expected: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error

bennybp:
Removing `- python` from the yaml files on my local machine seems to have made it keep python 3.8 for reasons I do not understand

dotsdl:
@bennybp oh my, well in that case, perhaps that's the best path forward. I don't understand it either.

dotsdl:
I'm able to reproduce @bennybp. The [adapter_parsl run](https://github.com/MolSSI/QCFractal/pull/628/checks?check_run_id=1289872182) with Python 3.8 fails because `conda` cannot resolve a solution for the dependencies. We can add this to the exclusion list.

dotsdl:
If this passes, we can merge this PR as an alternative to #623, since this includes all of those changes as well.

bennybp:
Excellent work! Unless you have objections, I will push a commit to remove the travis.yml and squash all these commits

dotsdl:
No objections! Don't mind one bit. We will need to see how the **codecov/project** integration works for additional PRs; I anticipate it will work fine from what codecov receives from each CI coverage report, but will need to see it in action.

PR #639, opened 2020-11-03
Remove _qcfractal_tags from optimizations extras

bennybp:
## Description

Removes `_qcfractal_tag` from optimizations. The record has all the information it needs, and so no need to pass this around.

Existing tasks with the `_qcfractal_tag` do not need to be migrated. It will be ignored.

## Changelog description

Remove unneeded `_qcfractal_tag` from optimizations

## Status

- [X] Code base linted
- [x] Ready to go


bennybp:
Just tested with a database with older tasks and seems to tolerate the existence of the _qcfractal_tags just fine

Issue #591, opened 2020-05-04
Can't serialize to JSON


mattwthompson:
**Describe the bug**
I wanted to save a collection to disk in order to avoid needing to downloads a large dataset every time I ran a test or re-started a notebook.

**To Reproduce**
```python3
import json

import qcportal


client = ptl.FractalClient()
ds = client.get_collection('OptimizationDataset', 'OpenFF Optimization Set 1')
ds.to_json(filename='data.json')
```
raises `TypeError: Object of type set is not JSON serializable`

**Expected behavior**
I expected to be able to save this out to JSON

**Additional context**
There is data in the collection object:

![image](https://user-images.githubusercontent.com/7935382/81007313-9c3eaf80-8e16-11ea-8926-823975422cfb.png)

 @loriab suggested on Slack that a `set` may have snuck in somewhere. This is probably a terrible collection to debug on since it includes something like 20,000 records.

bennybp:
Ok I see the problem. The `history` key is a set

dotsdl:
Diving into this one now.

dotsdl:
@bennybp I see you made a commit that addressed `history` being a `set` in a branch on your fork. Are you planning to merge this? Otherwise I can make the change in a PR here.

bennybp:
I started to fix it in that PR but abandoned it. It is a little more involved than just changing it to a list (there are some places that use the set functionality that have to also be modified.

Looking at the database, the database stores this info as JSON. Not entirely sure where this gets converted from a set to a list on the backend...

dotsdl:
Ah cool, thank you for that clarification. Working on a solution that doesn't fail in other places.

mattwthompson:
I ran into this again today. For my provenance, the quickest solution is to just pop the history. There's probably a way to map it onto a `list` but I don't think I need it for my use use case and just `data['history']= list(data['history'])` did not completely  work - it was happy to write to disk but could not be read back. I didn't look further into why.

```python3
import json

import qcportal

client = qcportal.FractalClient(verify=False)

dataset = client.get_collection(
    "OptimizationDataset",
    "OpenFF Iodine Chemistry Optimization Dataset v1.0",
)

with open("dataset.json", "w") as file:

    data = dataset.to_json()
    data.pop("history")

    json.dump(data, file)

with open("dataset.json", "r") as file:
    data = json.load(file)

bennybp:
Superseded by #749 for v0.50

PR #643, opened 2020-11-04
[WIP] Conceptual split of *_json into *_json, *_dict methods on Collections

dotsdl:
## Description

Addresses #591

Pydantic models support direct JSON serialization; we take advantage of this here to support use of e.g. sets in a model.

The `to_json` and `from_json` methods also technically dealt with dicts, not JSON strings; this is a distinction that matters since not all dicts are valid JSON constructs. We resolve that distinction here.

To do:
1. Need explicit roundtripping tests for `*_json` and `*_dict` methods yet.

## Changelog description

Collections feature `to_dict`,`from_dict` and `to_json`,`from_json` methods that are conceptually clearer, roundtrip with model fidelity.

## Status

- [ ] Code base linted
- [ ] Ready to go

### DIFF
--- a/qcfractal/interface/collections/collection.py
+++ b/qcfractal/interface/collections/collection.py
@@ -144,23 +144,25 @@ class Collection(abc.ABC):
         if tmp_data.meta.n_found == 0:
             raise KeyError("Warning! `{}: {}` not found.".format(class_name, name))
 
-        return cls.from_json(tmp_data.data[0], client=client)
+        return cls.from_dict(tmp_data.data[0], client=client)
 
     @classmethod
-    def from_json(cls, data: Dict[str, Any], client: "FractalClient" = None) -> "Collection":
-        """Creates a new class from a JSON blob
+    def from_dict(cls, data: Dict[str, Any], client: "FractalClient" = None) -> "Collection":
+        """Creates a new Collection instance from a dict representation.
+
+        Allows roundtrips from `Collection.to_dict`.
 
         Parameters
         ----------
         data : Dict[str, Any]
-            The JSON blob to create a new class from.
+            A dict to create a new Collection instance from.
         client : FractalClient, optional
-            A FractalClient connected to a server
+            A FractalClient connected to a server.
 
         Returns
         -------
         Collection
-            A constructed collection.
+            A Collection instance.
 
         """
         # Check we are building the correct object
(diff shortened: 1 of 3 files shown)


qcfractal/interface/collections/collection.py
```diff
@@ -180,26 +182,76 @@ def from_json(cls, data: Dict[str, Any], client: "FractalClient" = None) -> "Col
         ret = cls(name, client=client, **data)
         return ret
 
+    @classmethod
+    def from_json(cls, jsondata: Optional[str] = None, filename: Optional[str] = None, client: "FractalClient" = None) -> "Collection":
```
dgasmith, line 186: Recommend explicit `from_json(cls, *, jsondata...)` here.

qcfractal/interface/collections/collection.py
```diff
@@ -180,26 +182,76 @@ def from_json(cls, data: Dict[str, Any], client: "FractalClient" = None) -> "Col
...
+            A Collection instance.
+
+        """
+        if (jsondata is not None) and (filename is not None):
+            raise ValueError("One of `jsondata` or `filename` must be specified, not both")
+
+        if jsondata is not None:
+            data = json.loads(jsondata)
+        elif filename is not None:
+            with open(filename, 'r') as jsonfile:
```
dgasmith, line 214: Recommend path lib these days over raw `str`.

dgasmith (commented):
Perhaps its better to mirror the raw pydantic `.json()` and `parse_file/raw` methods here? Did OpenFF settle on a specific pattern?

qcfractal/interface/collections/collection.py
dotsdl: How do you mean? `pathlib.Path` objects should work with `open` in this way as well. 

I can change the docstring to indicate that a `path-like` is acceptable here; is there a preferred way to do this in the type hints?

qcfractal/interface/collections/collection.py
dotsdl: I don't quite understand the motivation; can you explain?

dotsdl:
> Perhaps its better to mirror the raw pydantic `.json()` and `parse_file/raw` methods here? Did OpenFF settle on a specific pattern?

Thank you for the review @dgasmith! 

I would favor what I consider the clearer `to_*` and `from_*` semantics that are used in e.g. `pandas`, `openforcefield`, `xarray`, etc. That appears to be a pattern pretty widely used in the scientific Python stack where data formats abound in variety, whereas `pydantic` appears more web-services oriented in which most things are represented in `json` or `json`-compatible serialization formats (my impression).

dgasmith:
Good to have confirmation, this is my base stance as well. I often cut out pydantic functionality these days from being easily discoverable:
```python
    def __dir__(self) -> List[str]:
        # Hide many Pydantic base functions to skip confusion in notebooks
        hidden_funcs = {
            "json",
            "parse_obj",
            "parse_raw",
            "parse_file",
            "from_orm",
            "construct",
            "schema_json",
            "schema",
            "validate",
            "update_forward_refs",
            "to_string",
        }
        return sorted(set(dir(super())) - hidden_funcs)
```

Note there is also movement to have pydantic model methods moved to an [internal namespace](https://github.com/samuelcolvin/pydantic/issues/1001) and since these objects are metaclasses, the ability to simply turn off the base functionality entirely. This gives a more composable surfaces and allows you to ditch the more webserver-esq abstractions.

qcfractal/interface/collections/collection.py
dgasmith: Sure, since `filename` and `jsondata` are ordered, the above natively gives precedences to jsondata while in reality they are mutually exclusive.

Consider:
```python
path = Path("collection.json")
from_json(path) # Fails
from_json(path.read_text()) # Works
from_json(filename=path)) # Works
```

The `*` enforces "good" behavior `from_json(jsondata=path.read_text())`. Fundamentally diligence doesn't scale, allowing poor/lazy behaviors to leak through replicates and grows within code bases over time.

qcfractal/interface/collections/collection.py
dgasmith: In light of static/gradual typing I try not to allow ambiguous behavior through which generally forces known types to be those types in stdlib. As in the only allowed option here is a `Path` object, not a string. You could also `Union[str, Path]`, but raw `str` is a bit restrictive when running through MyPy.

qcfractal/interface/collections/collection.py
dotsdl: Awesome, thanks! We're in agreement.

bennybp:
I think this was largely addressed in the previous version, and hopefully addressed in v0.50.

Likely to become even more moot with pydantic v2 (for example, PR #736)

PR #652, opened 2020-12-03
Generalized upsert functionality in socket

bennybp:
## Description

There are many functions in the socket that manually perform upserts (update or insert). This pulls that functionality
into a general function, hopefully cleaning up a lot of existing code and making it less error prone.

This will possibly enable a future optimization, where we use postgres upsert functionality (`ON CONFLICT DO ...`), increasing the performance. For now, this function is doing the slower, but more explicit way.

This PR adds two functions: One for upsert, and one for a pure insert that checks for existing records based on search keys.
These new functions are used for `kv_store` (upsert), and the keywords table (insert).

This PR also adds an improved way for returning metadata from the socket. This is implemented via dataclasses rather than dictionaries, and via multiple return. This is done for the kvstore functions. The hope is to move to a more standardized way of reporting this information outside the socket.

Because the add_keywords function is more user-facing, the return is left as is, which a shim converting to the old type of metadata. This has the benefit of not requiring changes anywhere else in the code including tests (for now).

## Changelog description

Start using common upsert/insert functions in the storage socket

## Status

- [x] Write upsert function
  - [X] General functionality
  - [x] Proper autoincrement primary key handling (error)
- [x] Code base linted
- [x] Ready to go

### DIFF
--- a/qcfractal/procedures/base.py
+++ b/qcfractal/procedures/base.py
@@ -73,12 +73,16 @@ class BaseTasks(abc.ABC):
             self.logger.warning(f"Found uncompressed error for result id {rdata['id']}")
             error = KVStore(data=rdata["error"])
 
-        # Now add to the database and set the ids in the diction
-        outputs = [stdout, stderr, error]
-        stdout_id, stderr_id, error_id = self.storage.add_kvstore(outputs)["data"]
-        rdata["stdout"] = stdout_id
-        rdata["stderr"] = stderr_id
-        rdata["error"] = error_id
+        # Now add to the database and set the ids in the dictionary
+        if stdout is not None:
+            _, stdout_add = self.storage.add_kvstore([stdout])
+            rdata["stdout"] = stdout_add[0].id
+        if stderr is not None:
+            _, stderr_add = self.storage.add_kvstore([stderr])
+            rdata["stderr"] = stderr_add[0].id
+        if error is not None:
+            _, error_add = self.storage.add_kvstore([error])
+            rdata["error"] = error_add[0].id
 
     @abc.abstractmethod
     def verify_input(self, data):
--- a/qcfractal/storage_sockets/models/sql_models.py
+++ b/qcfractal/storage_sockets/models/sql_models.py
@@ -28,6 +28,7 @@ from sqlalchemy.sql import text
 
 from qcfractal.interface.models.task_models import ManagerStatusEnum, PriorityEnum, TaskStatusEnum
 from qcfractal.interface.models.common_models import CompressionEnum
+from qcfractal.interface.models import KVStore
 from qcfractal.storage_sockets.models.sql_base import Base, MsgpackExt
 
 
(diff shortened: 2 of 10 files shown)


qcfractal/storage_sockets/models/sql_models.py
```diff
@@ -196,6 +219,13 @@ class KeywordsORM(Base):
     __table_args__ = (Index("ix_keywords_hash_index", "hash_index", unique=True),)
     # meta = {'indexes': [{'fields': ('hash_index', ), 'unique': True}]}
 
+    def to_pydantic_model(self):
```
bennybp, line 222: Should ORM classes have this function, even if trivial (like this one for keywords?)

bennybp:
Ready for review. I prefer to keep this small-ish to make review easier :)

The next PR will use this functionality to fix a few small bugs in services related to duplicating kvstore rows.

qcfractal/procedures/base.py
```diff
@@ -73,12 +73,16 @@ def retrieve_outputs(self, rdata):
             self.logger.warning(f"Found uncompressed error for result id {rdata['id']}")
             error = KVStore(data=rdata["error"])
 
-        # Now add to the database and set the ids in the diction
-        outputs = [stdout, stderr, error]
-        stdout_id, stderr_id, error_id = self.storage.add_kvstore(outputs)["data"]
-        rdata["stdout"] = stdout_id
-        rdata["stderr"] = stderr_id
-        rdata["error"] = error_id
+        # Now add to the database and set the ids in the dictionary
```
dgasmith, line 76: Originally performed this way for bulk insert performance, I'm unsure if you will still hit these bottlenecks.

qcfractal/storage_sockets/models/sql_models.py
dgasmith: Could help with consistency, since the ORM and model diverge slightly it would prevent replication of the alterations throughout.

qcfractal/storage_sockets/sqlalchemy_common.py
```diff
@@ -0,0 +1,192 @@
...
+
+    # This is the type of the ORM. It should be the same for all items in the list
+    orm_type = type(data[0])
+    assert all(type(x) == orm_type for x in data)
+
+    for idx, rec in enumerate(data):
+        # Build up the filter dictionary
+        query = {}
+        for c in search_cols:
+            query[c] = getattr(rec, c)
```
dgasmith, line 164: Since you have the keys already you can likely generalize the bulk add by building a key tuple (assuming `search_col` order is static) that you can map back and forth between objects and indices. A single query can then be performed to check of existing and then a bulk add performed which will be ~100-1000x faster than current.

I'm unsure of your current performance constraints, but worth thinking about if needed.

dgasmith (commented):
Cool!

bennybp:
Closing since this is outdated. Going a slightly different direction in the `next` branch

PR #664, opened 2021-03-08
Adds indices to base_result and access_log

bennybp:
Adds a unique index to `base_result.hash_index`. This field is used in de-duplicating procedures, so may have been a source of slowdown. This is only temporary, as we are working towards removing indices for that use case.

This also fixes/adds indices on access_log (adds one for `access_type`. The old index named access_type was actually the date).

## Description

Adds database indices for access_log and base_result.hash_index

## Status

- [X] Code base linted
- [ ] Ready to go


qcfractal/alembic/versions/cd3d3e2915ce_add_indices_to_access_and_base_result.py
```diff
@@ -0,0 +1,34 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_index("access_type", table_name="access_log")
+    op.create_index(op.f("ix_access_log_access_date"), "access_log", ["access_date"], unique=False)
+    op.create_index(op.f("ix_access_log_access_type"), "access_log", ["access_type"], unique=False)
+    op.create_index("ix_base_result_hash_index", "base_result", ["hash_index"], unique=True)
```
doaa-altarawy, line 24: We can't have the unique=True here although it's supposed to be unique. The reason is that it's not actually 100% unique. A rare case of duplicate hash can happen, although very tiny. So, we try to enforce this uniqueness in the python side to catch and handle this rare case, but not to leave it to Postgres.

qcfractal/alembic/versions/cd3d3e2915ce_add_indices_to_access_and_base_result.py
```diff
@@ -0,0 +1,34 @@
...
+down_revision = "c05d63683601"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_index("access_type", table_name="access_log")
+    op.create_index(op.f("ix_access_log_access_date"), "access_log", ["access_date"], unique=False)
+    op.create_index(op.f("ix_access_log_access_type"), "access_log", ["access_type"], unique=False)
```
doaa-altarawy, line 23: I think `unique=False` is the default, so no change on access_type index

bennybp: No change, but the name has changed (the previous index named `access_type` was actually an index on `access_date`. So this renames that, plus adds an index on date (probably needed for the dashboard). The unique=False comes from alembic autogeneration

qcfractal/alembic/versions/cd3d3e2915ce_add_indices_to_access_and_base_result.py
bennybp: This is tricky. We aren't actually checking for that case (since that would be a bit expensive): https://github.com/MolSSI/QCFractal/blob/6d4fd08fcd84ac51262c75e18985c7d868df90e2/qcfractal/storage_sockets/sqlalchemy_socket.py#L723

If we wanted, we could still check for that case and warn the user, but still keep uniqueness in the database. I would expect it to be exceedingly rare, since molecule hashing uses sha1.

https://github.com/MolSSI/QCElemental/blob/ec9d754414824e129729f477e6077a14d859dc85/qcelemental/models/molecule.py#L758

PR #668, opened 2021-03-18
Make molecules not unique by hash again

bennybp:
## Description

It turns out that due to some other issues, the molecule table can have duplicate molecules. This is confirmed in the main MolSSI database, which has >400 duplicates. This of course breaks the migration in v0.15.4

This is a serious issue, which also leads to other consistency issues. But that is something for another day

So for now what we can do is this:

1.) Comment out that part of the previous migration (`c05d`)
2.) Add a migration that deletes the old index, and adds a new one with `unique=False` (`6b07`)

Assume this commit forms release v0.15.5. These modifications should work in the following two scenarios

1.) User upgraded (successfully) to v0.15.4 prior to this commit/release. Their migration for `c05d` made the molecule hash unique. The new migration `6b07` will delete that unique index and make a new one with `unique=False`

2.) User did not upgrade to v0.15.4. The updated old migration `c05d` will not touch the molecule table. The new migration `6b07` will delete the existing  (non-unique) index, and just recreate it.

## Status

- [X] Code base linted
- [ ] Ready to go


bennybp:
This still isn't completely functional. There are duplicate `base_result.hash_index` as well. I have to remove that migration the same as I did for molecule :(

bennybp:
After these changes, it will now successfully migrate the molssi database

PR #669, opened 2021-03-31
Adds Google Tag Manager Container to QCFractal Docs for GA4

SinaMostafanejad:
Adds Google Tag Manager's container for Google Analytics 4 to the layout.html template to propagate through all documentation pages allowing for tracking user data for marketing and communications purposes.


bennybp:
Will need to be updated for v0.50 (and new versions of GA?)

Issue #674, opened 2021-05-18
`FractalClient.query_results` gives dict for query with no use of `include`


dotsdl:
**Describe the bug**

The `FractalClient.query_results` method advertises that [a dict will be returned if `include` is specified](https://github.com/MolSSI/QCFractal/blob/be181409d446d83c7cc2ec8c5f789ea94503d3b8/qcfractal/interface/client.py#L651-L703), which corresponds to pulling only some columns from the corresponding tables in the database and returning them via the REST API. 

However, for the `'OpenFF Full Optimization Benchmark 1'` `OptimizationDataset`, we observe a dict returned for each result even *without* specifying `include`:

```python

client = FractalClient()

optds = client.get_collection("OptimizationDataset", 'OpenFF Full Optimization Benchmark 1')
optds.status()

opt = optds.df['default'].iloc[0]

# asking for the corresponding gradient point calculations gives a list of dicts
results = client.query_results(opt.trajectory)

# asking for a single gradient point calculation also yields a list of dicts (just the one)
result = client.query_results(opt.trajectory[0])
```

This is unexpected, and it is not clear why this occurs.

dotsdl:
From discussion with @bennybp, this is due to changes in some QCElemental models since this dataset was created. He has a migration coming soon that will map old keys to the new ones where present in datasets such as this. Will close this issue when this is out.

PR #675, opened 2021-05-18
Small shim in `FractalClient.query_results` for old Result models

dotsdl:
Addresses #674.

There are some datasets in the database with old QCElemental result models that fail validation when queries. These are retruned by `FractalClient._automodel_request` as dictionaries, even without `include` specified. This shim handles these cases as gracefully as we can at this time until the old results are migrated to the new result model.

## Status

- [ ] Code base linted
- [ ] Ready to go

### DIFF
--- a/qcfractal/interface/client.py
+++ b/qcfractal/interface/client.py
@@ -720,7 +720,10 @@ class FractalClient(object):
         # Add references back to the client
         if not include:
             for result in response.data:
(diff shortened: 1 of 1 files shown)


dotsdl:
This PR will quickly [become irrelevant](https://github.com/MolSSI/QCFractal/issues/674#issuecomment-843629545) given upcoming migration work.

PR #676, opened 2021-05-24
Make fields of provenance not null, and fix other validation issues

bennybp:
## Description
This PR migrates some old data in the database. This old data was causing validation errors.

Errors fixed:

- Molecule where connectivity is an empty list (should either be None or a list with length > 1) (See https://github.com/MolSSI/QCArchiveExamples/issues/16)
- Provenance fields that were None/null (should be empty string)
- Result properties with `mp2_total_correlation_energy` (should be `mp2_correlation_energy`)

This has been thoroughly tested against a copy of the production database to ensure integrity (ie, it doesn't change data unless it needs to be changed)

## Changelog description
Provide data migration for old data that no longer passes validation

## Status

- [X] Code base linted
- [X] Ready to go

### DIFF
--- a/qcfractal/alembic/versions/469ece903d76_migrate_provenance_to_not_null.py
+++ b/qcfractal/alembic/versions/469ece903d76_migrate_provenance_to_not_null.py
@@ -1,0 +1,61 @@
+"""migrate provenance to not null
+
+Revision ID: 469ece903d76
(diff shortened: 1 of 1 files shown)


bennybp:
Failed tests are unrelated to this PR

PR #678, opened 2021-06-02
Fix pandas/byte array conversion in views (and other CI issues)

bennybp:
## Description
Fixes two issues that have been plaguing CI for a while

- HDF5 stores fields as byte arrays. We need to convert these back to strings when used as indices in pandas dataframes
- Conda's weird dependency resolution will install different versions of `dask` and `distributed`. So pin them for now and move on with life.

## Changelog description

- Fix byte array/string conversion between HDF5/pandas in view code
- Pin version of dask to prevent conda from installing incompatible versions

## Status

- [X] Code base linted
- [X] Ready to go

### DIFF
--- a/qcfractal/interface/collections/dataset_view.py
+++ b/qcfractal/interface/collections/dataset_view.py
@@ -214,8 +214,10 @@ class HDF5View(DatasetView):
             with self._read_file() as f:
                 entry_group = f["entry"]
                 self._index = pd.DataFrame({"index": entry_group["entry"][()]})
+                self._index["index"] = self._index["index"].str.decode("utf-8")
                 self._index["_h5idx"] = range(len(self._index))
                 self._index.set_index("index", inplace=True)
+
         if subset is None:
             return self._index.reset_index()
         else:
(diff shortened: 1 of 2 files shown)


SinaMostafanejad (approved):
Thanks for putting these changes in a consistent and simple way.

qcfractal/interface/collections/dataset_view.py
```diff
@@ -233,7 +235,15 @@ def get_entries(self, subset: Optional[List[str]] = None) -> pd.DataFrame:
                     raise ValueError(
                         f"Unknown entry class ({entry_group.attrs['model']}) while " f"reading HDF5 entries."
                     )
+
                 self._entries = pd.DataFrame({field: entry_group[field][()] for field in fields})
+
+                # HDF5 stores these as byte arrays. But we use strings in pandas...
+                self._entries["name"] = self._entries["name"].str.decode("utf=8")
```
SinaMostafanejad, line 242: utf-8? with '-' instead of '='?

qcfractal/interface/collections/dataset_view.py
```diff
@@ -233,7 +235,15 @@ def get_entries(self, subset: Optional[List[str]] = None) -> pd.DataFrame:
...
+
+                if entry_group.attrs["model"] == "ReactionEntry":
+                    self._entries["stoichiometry"] = self._entries["stoichiometry"].str.decode("utf=8")
```
SinaMostafanejad, line 245: utf-8? with '-' instead of '='?

bennybp: See PR #679

PR #680, opened 2021-06-05
Fix incorrect reporting in collection status

bennybp:
## Description

Fix incorrect reporting in some collections. If a status is specified to the `status()` function, the number of records corresponding to that status was incorrectly calculated. It was calculated in such a way that only if all specifications were that status for a record would it be counted. This led to a mismatch between `get_status()` and `get_status(status='COMPLETE')`

Before:
before

After:
after

## Changelog description

Fixes incorrect status reporting in collections

## Status

- [X] Code base linted
- [X] Ready to go


dotsdl (approved):
Looks great @bennybp! Thank you for the quick identification and fix!

PR #684, opened 2021-09-16
Fix #683 

### LINKED ISSUE #683: Compute keywords have inconsistent alias case sensitvity
**Describe the bug**
Datasets store keywords under an alias which is lowered before being stored, however when adding compute tasks for a dataset and providing the original alias name you get an error as it is not lowered before being searched.

**To Reproduce**
As we can not add compute to the public QCArchive I instead use the internal function to look for a keyword alias which as already been stored.

```python
from qcportal import FractalClient
client = FractalClient()
ds = client.get_collection("dataset", "OpenFF Theory Benchmarking Single Point Energies v1.0")
ds._default_parameters(program="psi4", method="B3LYP-D3BJ", basis="DEF2-TZVP", keywords="B3LYP-D3BJ/DEF2-TZVP")
```
> KeyError: "KeywordSet alias 'B3LYP-D3BJ/DEF2-TZVP' not found for program 'psi4'."

now check for the alias using lower case 
```python
ds._default_parameters(program="psi4", method="B3LYP-D3BJ", basis="DEF2-TZVP", keywords="B3LYP-D3BJ/DEF2-TZVP".lower())
```
> ('B3LYP-D3BJ/def2-tzvp-b3lyp-d3bj/def2-tzvp',
 {'driver': 'energy',
  'program': 'psi4',
  'method': 'B3LYP-D3BJ',
  'basis': 'DEF2-TZVP',
  'keywords': '2'},
 {'driver': 'energy',
  'program': 'psi4',
  'method': 'B3LYP-D3BJ',
  'basis': 'DEF2-TZVP',
  'keywords': 'b3lyp-d3bj/def2-tzvp'})

**Expected behavior**
The keyword alias should be lowered [here](https://github.com/MolSSI/QCPortal/blob/38eed6425efaf23f4dbe3ecf49bcaca57e8d7918/qcportal/collections/dataset.py#L1035) when adding compute.

**Additional context**

jthorton:
## Description
This should fix 683 by lowering the keyword alias before searching for the default parameters and adding compute.

- [ ] Code base linted
- [ ] Ready to go

### DIFF
--- a/qcfractal/interface/collections/dataset.py
+++ b/qcfractal/interface/collections/dataset.py
@@ -1032,6 +1032,7 @@ class Dataset(Collection):
                 keywords_alias = self.data.default_keywords[program]
                 keywords = self.data.alias_keywords[program][keywords_alias]
         else:
+            keywords = keywords.lower()
             if (program not in self.data.alias_keywords) or (keywords not in self.data.alias_keywords[program]):
                 raise KeyError("KeywordSet alias '{}' not found for program '{}'.".format(keywords, program))
 


bennybp:
Thanks for the quick fix.

Does it make sense to have case-sensitive keyword aliases in the future? I personally don't think so but maybe you have a use case?

jthorton:
No problem always happy to contribute to an awesome project!

>Does it make sense to have case-sensitive keyword aliases in the future? I personally don't think so but maybe you have a use case?

I also don't think so, lowering is fine for us. Maybe the docstring for add keywords should also be changed to communicate that the alias is always lowered?

bennybp:
Ok I will keep that in mind in the new versions. Thanks!

Issue #688, opened 2021-10-03
wb97m-d3bj, b97-d3bj methods fail because of splitting into functional+dispersion correction calculations


pavankum:
**Describe the bug**
For `b97-d3bj` and `wb97m-d3bj` methods, an energy calculation seems to be divided into `b97`, `wb97m` plus the `d3bj` dispersion correction, and this results in psi4 throwing an input_error saying method doesn't exist since there are no functionals `b97` and `wb97m` separately. Not sure where the split happens since qcengine works fine with the above functionals. 

**To Reproduce**
Not sure how to reproduce this. On [this single point energies dataset](https://github.com/openforcefield/qca-dataset-submission/pull/227), all runs under those two specs fail. Here is one error message along with the input it takes, where we can see in the qcspec under model the method is `wb97m`, which is invalid
 
Code to retrieve the error
```
import lzma

from qcfractal.interface import FractalClient

client = FractalClient.from_file()
ds = client.get_collection('Dataset', 'OpenFF Theory Benchmarking Single Point Energies v1.0')

spec_name = 'wb97m-d3bj/dzvp'
method = 'wb97m-d3bj'
basis = 'dzvp'
recs = ds.get_records(method=method, basis=basis, keywords=spec_name)

indx = len(recs)
print("Record ID: ", recs[indx - 1].iloc[0].record.id, "Status: ", recs[indx - 1].iloc[0].record.status)
print(client.query_tasks(base_result=int(recs[indx - 1].iloc[0].record.id)))
kv = client.query_kvstore(recs[indx - 1].iloc[0].record.error)
out = lzma.decompress(kv[list(kv.keys())[0]].data).decode()
print("-----------")
print(bytes(str(out), 'utf-8').decode("unicode_escape"))```
```
Output of this code snippet 

```
Record ID:  76485400 Status:  RecordStatusEnum.error
[TaskRecord(id='12656653', spec=PythonComputeSpec(function='qcengine.compute', args=[{'id': '76485400', 'schema_name': 'qcschema_input', 'schema_version': 1, 'molecule': {'schema_name': 'qcschema_molecule', 'schema_version': 2, 'validated': True, 'symbols': array(['C', 'C', 'C', 'C', 'C', 'C', 'C', 'C', 'C', 'C', 'C', 'C', 'N',
       'O', 'O', 'S', 'H', 'H', 'H', 'H', 'H', 'H', 'H', 'H', 'H', 'H'],
[...]

PR #691, opened 2021-10-11
Hack fix for d3 calculations which shouldn't be split in the planner

bennybp:
## Description

Fixes issue #688

These functionals shouldn't be split in the composition planner, as the first part isn't available in psi4 as a plain functional.

This is a temporary fix. The next major version will probably do away with the composition planner altogether.

## Changelog description

Skip composition planner for certain functionals

## Status

- [X] Code base linted
- [x] Ready to go

### DIFF
--- a/qcfractal/interface/collections/collection_utils.py
+++ b/qcfractal/interface/collections/collection_utils.py
@@ -99,6 +99,10 @@ def composition_planner(program=None, method=None, basis=None, driver=None, keyw
 
     base = {"program": program, "method": method, "basis": basis, "driver": driver, "keywords": keywords}
 
+    # HACK: shortcut these. See issue #688
+    if method.lower() in ["b97-d3bj", "b97m-d3bj", "wb97m-d3bj"]:
+        return [base]
+
     if ("-d3" in method.lower()) and ("dftd3" != program.lower()) and ("hessian" != driver.lower()):
         dftd3keys = {"program": "dftd3", "method": method, "basis": None, "driver": driver, "keywords": None}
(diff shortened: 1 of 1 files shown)


bennybp:
The composition planner is called both for computing and accessing:

https://github.com/MolSSI/QCFractal/blob/8a291bc5b9063d474045c712eb4b4569acf871a3/qcfractal/interface/collections/dataset.py#L1136

https://github.com/MolSSI/QCFractal/blob/8a291bc5b9063d474045c712eb4b4569acf871a3/qcfractal/interface/collections/dataset.py#L1223

I guess there may still be a problem with existing records? But with the way datasets are currently written that shouldn't be an issue. The previously submitted ones will disappear from the dataset (because the dataset doesn't explicitly link to individual records)

dotsdl:
> The composition planner is called both for computing and accessing:
> 
> https://github.com/MolSSI/QCFractal/blob/8a291bc5b9063d474045c712eb4b4569acf871a3/qcfractal/interface/collections/dataset.py#L1136
> 
> https://github.com/MolSSI/QCFractal/blob/8a291bc5b9063d474045c712eb4b4569acf871a3/qcfractal/interface/collections/dataset.py#L1223
> 
> I guess there may still be a problem with existing records? But with the way datasets are currently written that shouldn't be an issue. The previously submitted ones will disappear from the dataset (because the dataset doesn't explicitly link to individual records)

Ah perfect, you're right, missed the usage in `Dataset._compute`. With this change use of a new `qcportal` from a new QCFractal release should generate the right set of records when called. Can basically try to resubmit the dataset and this *should* just work.

PR #694, opened 2021-10-22
QCElemental 0.23.0 update

jthorton:
## Description
This PR should make QCFractal compatible with qcelemental=0.23.0, we now use the json encoding on dicts when adding or updating results which should remove all non-json serializable objects. 

## Changelog description
Update QCFractal to work with QCElemental=0.23.0

## Status

- [x] Code base linted
- [ ] Ready to go

### DIFF
--- a/devtools/conda-envs/openff.yaml
+++ b/devtools/conda-envs/openff.yaml
@@ -1,7 +1,7 @@
 name: qcarchive
 channels:
   - defaults
-  - psi4
+  - psi4/label/dev
   - conda-forge
 dependencies:
   - pip
@@ -39,7 +39,7 @@ dependencies:
   - requests-mock
 
 #   Environment specific includes
-  - psi4 >=1.4.1
+  - psi4
   - gau2grid
   - rdkit
   - geometric >=0.9.3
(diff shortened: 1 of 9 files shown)


devtools/conda-envs/openff.yaml
```diff
@@ -1,8 +1,7 @@
 name: qcarchive
 channels:
   - defaults
-  - omnia
-  - psi4/label/dev
```
bennybp, line 5: I think we should leave in psi4/label/dev and test against the current master version. Right @loriab ?
I suppose we can leave the pin below

bennybp:
I have zero idea why this is failing now, and cannot reproduce locally

bennybp:
Ok seems to be something with alembic on the conda defaults channel not pulling in importlib_resources. The reason for that is a conda-related mystery which I can't be bothered to figure out (since it does seem to be specified as a dependency).

If you install alembic from the default channel, and then do `pip install -e .` for qcfractal, it then gets installed because pip is smart enough to actually handle dependencies I guess. (but we do `--no-deps` in the CI)

bennybp:
Ok ready to merge if you are ok with the psi4 version change

jthorton:
@bennybp thanks for fixing this, LGTM!

bennybp (approved):

PR #700, opened 2021-12-02
Replace use of `concurrent.futures` with `multiprocessing` in `ExecutorAdaptor` for hard-termination of tasks

### LINKED ISSUE #699: Are temporary files automatically deleted when SIGINT received?
We're having some trouble where jobs killed by LSF are not cleaning up their temporary files. I notice that [this method](https://github.com/MolSSI/QCFractal/blob/5f00dd06bb34ca912c4055f0cbac60863ea89c7f/qcfractal/cli/cli_utils.py#L80-L102) can register a signal handler and gracefully clean up when a signal is received.

According to [this page](https://pretagteam.com/question/how-much-time-before-sigkill), the LSF job manager first sends a SIGINT, waits a bit, then sends a SIGKILL, and finally a SIGTERM. 

Does QCFractal already gracefully intercept SIGINT and clean up its temporary files (e.g. psi4 tempfiles) and exit? Or would this behavior have to be added?

cc: @dotsdl

dotsdl:
Closes #699.

Swaps out `concurrent.futures.ProcessPoolExecutor` for `multiprocessing.Pool`.

We needed a way to hard-terminate processes without waiting for them to complete. `multiprocessing` gives us this.

## Changelog description

`ExecutorAdaptor` now uses `multiprocessing.Pool` instead of `concurrent.futures.ProcessPoolExecutor`, and hard-terminates tasks on manager shutdown.

## Status

- [x] Code base linted
- [x] Ready to go

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -80,7 +80,7 @@ jobs:
 
       # TODO: verify this works as we expect
       - name: Code coverage
-        uses: codecov/codecov-action@v1
+        uses: codecov/codecov-action@v2
         with:
           file: ./coverage.xml
           fail_ci_if_error: true
--- a/devtools/conda-envs/adapter_dask.yaml
+++ b/devtools/conda-envs/adapter_dask.yaml
@@ -44,5 +44,5 @@ dependencies:
   - dask-jobqueue >=0.5.0
 
 #   QCArchive includes
-  - qcengine >=0.20,<0.21
-  - qcelemental >=0.23,<0.24
+  - qcengine >=0.21
+  - qcelemental >=0.24
--- a/devtools/conda-envs/adapter_dask.yaml
+++ b/devtools/conda-envs/adapter_dask.yaml
@@ -1,7 +1,7 @@
 name: qcarchive
 channels:
-  - conda-forge
   - defaults
+  - conda-forge
 dependencies:
   - pip
 
--- a/devtools/conda-envs/adapter_dask.yaml
+++ b/devtools/conda-envs/adapter_dask.yaml
@@ -1,7 +1,7 @@
 name: qcarchive
 channels:
-  - defaults
   - conda-forge
+  - defaults
 dependencies:
   - pip
 
--- a/devtools/conda-envs/adapter_fireworks.yaml
+++ b/devtools/conda-envs/adapter_fireworks.yaml
@@ -38,8 +38,8 @@ dependencies:
   - requests-mock
 
 #   QCArchive includes
-  - qcengine >=0.20,<0.21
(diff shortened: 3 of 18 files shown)


dotsdl:
One note: for the case of optimizations, which can be long-running tasks that we sought to address here, a gradient calculation with e.g. `psi4` will not immediately terminate upon exit of the manager. This is due to the fact that [killing the processes in the pool](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.Process.terminate) does not kill any child processes, such as those started via `subprocess.Popen`. 

I don't see a clean way to propagate the SIGTERM that `qcengine.compute_procedure` receives in the process pool down through `Psi4Harness.compute` and into [the subprocess wait block](https://github.com/MolSSI/QCEngine/blob/cca931694e50094736d44a957f513fa22588c49a/qcengine/util.py#L478) to terminate the child process. However, the current solution at least eliminates the manager immediately so that file cleanup can occur shortly after. The remaining child process is orphaned and will be handled by PID 1 on the host.

dotsdl:
Edit: I do have a solution that can cleanly terminate the underlying `subprocess.Popen`. Will formulate as a separate PR here and to QCEngine.

dotsdl:
From live testing with a local `qcfractal-manager`, we only need the changes introduced in this PR to propagate `KeyboardInterrupt` into the process pool workers that execute calls to QCEngine `compute` and `compute_procedure`. Sending a SIGINT, SIGHUP, or SIGTERM to the manager cleanly kills the manager and child `psi4` calls down through QCEngine with a process pool, and all scratch files are cleaned up.

@bennybp we are getting CI failures from `psql` with all environments. It is not a fundamental issue with the `conda-forge` `postgresql` package; installing it alone with e.g.:

```
conda create -n postgres -c conda-forge postgresql python=3.7
```
gives a `psql --version` call that works. Is there a dependency in the CI envs that could be causing the error we're seeing in the CI (reproduced locally):

```
psql: symbol lookup error: /home/david/.conda/envs/qcfractal-base/bin/../lib/libreadline.so.8: undefined symbol: tputs
```
?

dotsdl:
Or alternatively, should we be restricting the `postgresql` version in these environments to some max version?

dotsdl:
Not sure what the root cause of the CI failures are now. Superficially they appear unrelated to manager execution, but not certain.

dotsdl:
@bennybp looks like we're good to go here! Thanks again for all your help!

dotsdl:
@bennybp can you review and merge? If we get this and #701 in we should be able to cut a release that we can roll out to production managers.

dotsdl:
Looks like all tests still pass with both QCElemental 0.24 and QCEngine 0.22!

PR #701, opened 2021-12-08
Attempting to run CI with QCEngine 0.21, QCElemental 0.24

dotsdl:
## Status

- [ ] Code base linted
- [ ] Ready to go

### DIFF
--- a/devtools/conda-envs/adapter_dask.yaml
+++ b/devtools/conda-envs/adapter_dask.yaml
@@ -44,5 +44,5 @@ dependencies:
   - dask-jobqueue >=0.5.0
 
 #   QCArchive includes
-  - qcengine >=0.20,<0.21
-  - qcelemental >=0.23,<0.24
+  - qcengine =0.21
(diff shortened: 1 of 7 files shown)


dotsdl:
@bennybp not exactly certain, but in order to use QCEngine 0.21.1, which will include MolSSI/QCEngine#339, I think we'll need to resolve the issues raised by the CI here. Thoughts?

bennybp:
Just a small change needed, and it is now passing. Good to go?

bennybp:
On second thought, maybe we should pin to qcengine >= 0.21 and qcelemental >= 0.24? @loriab says she will be creating a new version of qcengine soon, which wouldn't be pulled in (I think). Or are conda and pip smart enough to pull matching minor versions but different patch levels?

dotsdl:
@bennybp just tried out going with qcengine >= 0.21 and qcelemental >= 0.24; looks like all tests pass!

bennybp (approved):

PR #732, opened 2023-06-24
[next branch] Hadrien's change

hadim:
_Only for visibility purposes, I might force push here often to keep the commits clean._


bennybp:
The  URI parsing logic has been fixed. Using env vars for manager config hasn't been. I am still open to that, however with the release of v0.50, you can open a PR directly to `main` :)

PR #746, opened 2023-09-13
Some small docs & GHA cleanup

bennybp:
## Description
Add manager docs from @dotsdl and change install instructions to use conda-forge

Also, don't push to gh-pages on PRs

## Status

- [X] Code base linted
- [X] Ready to go


bennybp:
The error with `_typeshed` in the docs build is an issue with sphinx 7.2 where they set `TYPE_CHECKING=True`

Issue #548, opened 2020-01-21
Autogenerate Collection lname at the SQL layer


dgasmith:
**Is your feature request related to a problem? Please describe.**
At the moment we generate the Collection lowercase search column in the Python layer which means we need to update two columns upon rename that can become potentially out of sync.

**Describe the solution you'd like**
SQL and SQLAlchemy have a `GENERATE ALWAYS AS` column type which as noted will automatically build the relevant column upon insert.

SQLAlchemy documentation [here](https://docs.sqlalchemy.org/en/13/core/defaults.html#computed-generated-always-as-columns)
Desired SQL generation: `lname VARCHAR GENERATED ALWAYS AS LOWER(name) STORED`

**Additional context**
This will prevent potentially problematic updates and push additional logic into the SQL layer. I believe this is the optimal solution as `(collection_type, lname)` forms our compound unique index for this table. Doing a search by casting the column to lowercase on the fly is unlikely to ever become expensive, but this is still an easy optimization and gains a bit of experience (~250 rows currently in the MolSSI database).

bennybp:
https://docs.sqlalchemy.org/en/20/core/defaults.html#computed-ddl

PR #750, opened 2023-09-15
Use computed column for dataset lname

bennybp:
## Description
Use computed columns for dataset lname. Closes issue #548 

Doing it sql side is safer that remembering to do it python side

## Changelog description
Use postgres computed columns for dataset lowercase name (lname)

## Status
- [X] Code base linted
- [X] Ready to go


bennybp:
Failing tests seem to be due to some changes in geometric

PR #758, opened 2023-09-21
Implement retries for spurious connection issues

### LINKED ISSUE #741: Add use of retries with exponential backoff to `PortalClient` requests
IP networks are not 100% reliable, so some fraction of HTTP requests from the `PortalClient` can fail during normal usage. Adding built in use of retries with exponential backoff (e.g. via [retry](https://pypi.org/project/retry/)) would smooth over these failures for users/compute.

bennybp:
## Description

Requests that result in certain errors due to spurious connection/networking issues should now be automatically retried.

The types of exceptions handled should cover most cases, but I am open to expanding the list if people find ones that I missed.

The request will be tried again after waiting 0.5, 1.0, 2.0, and 4.0 seconds (since the previous request)

Seem reasonable @dotsdl? Closes #741 

## Changelog description

Requests now will be automatically retried in case of connection or networking issues

## Status
- [X] Code base linted
- [X] Ready to go

### DIFF
--- a/qcportal/qcportal/client_base.py
+++ b/qcportal/qcportal/client_base.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import logging
 import os
 import time
 from typing import (
@@ -16,7 +17,6 @@ import pydantic
 import requests
 import yaml
 from pkg_resources import parse_version
-from retry.api import retry_call
 
 from . import __version__
 from .exceptions import AuthenticationFailure
@@ -87,6 +87,8 @@ class PortalClientBase:
             If a Message-of-the-Day is available, display it
         """
 
+        self._logger = logging.getLogger("PortalClientBase")
+
         # For developer use and debugging
         self.debug_requests = False
 
@@ -111,11 +113,14 @@ class PortalClientBase:
 
         self._req_session.headers.update({"User-Agent": f"qcportal/{__version__}"})
 
-        self._timeout = 60
         self.encoding = "application/json"
 
-        # Mode toggle for network error testing, not public facing
-        self._mock_network_error = False
+        self.timeout = 60
+
+        # Handling retries of requests
+        self.retry_max = 5
+        self.retry_delay = 0.5
+        self.retry_backoff = 2
 
         # If no 3rd party verification, quiet urllib
         if self._verify is False:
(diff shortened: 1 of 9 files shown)


dotsdl:
Reviewing now! Thanks for this @bennybp!

dotsdl:
[retry looks kinda dead](https://github.com/invl/retry/issues/36); looks like [backoff](https://github.com/litl/backoff) may be a better bet for avoiding bitrot?

qcportal/qcportal/manager_client.py
```diff
@@ -49,8 +49,13 @@ def __init__(
...
 
     def _update_on_server(self, manager_update: ManagerUpdateBody) -> None:
+        # Don't allow retries - we handle that elsewhere in the manager (by trying again later)
         return self.make_request(
-            "patch", f"compute/v1/managers/{self.manager_name_data.fullname}", None, body=manager_update
+            "patch",
+            f"compute/v1/managers/{self.manager_name_data.fullname}",
+            None,
+            body=manager_update,
+            allow_retries=False,
```
dotsdl, line 58: Why disable retries here? I think it makes sense to have layers of retry logic; this handles the case of simple network failures, which can happen pretty often even in the best case.

qcportal/qcportal/client_base.py
```diff
@@ -303,7 +305,24 @@ def _request(
...
+                    fkwargs={
+                        "request": prep_req,
+                        "verify": self._verify,
+                        "timeout": self._timeout,
+                    },
+                    exceptions=(requests.exceptions.ConnectionError, requests.exceptions.ConnectTimeout),
+                    tries=5,
+                    delay=0.5,
+                    max_delay=5,
+                    backoff=2,
```
dotsdl, line 321: Instead of hardcoding these, perhaps use attributes on `PortalClientBase` configurable on `init`? Probably really only need to expose `tries`, `max_delay`, and `backoff`, but could expose all of these if you want to allow for customizability at runtime.

Also, probably need to set `jitter` as well to avoid many workers retrying all in-sync during a network hiccup.

dotsdl (changes requested):
A few comments, and suggest making the parameters for retries configurable at runtime. I also think that `retry` as a library looks a bit dead, so may be a better bet to use something else, like `backoff`. Can probably still use `backoff`'s decorator directly with configurable paramters as e.g.:

```
backoff.on_predicate(backoff.expo, ...)(func_to_retry)(*args, **kwargs)
```

If adding another dependency seems overkill for this, it's not too hard to implement your own solution with the behavior you want. As an example, we do this in a simple way in [alchemiscale here](https://github.com/openforcefield/alchemiscale/blob/9a20f2016e7e1dd8aac904ab3c7eb61505d421c8/alchemiscale/base/client.py#L129).

qcportal/qcportal/manager_client.py
bennybp: The managers have their own retry logic that works on a longer timeframe, where a manager self-destructs if there have been too many missed contacts. Right now that logic depends on knowing how many heartbeats have been missed, and adding additional time here would throw that off (not by too much in production, but enough to make testing more difficult)

qcportal/qcportal/client_base.py
bennybp: I'm fairly opposed to adding these to the constructor, since a vast, vast majority of people would use the defaults, and this would almost double the number of arguments available. I can do something similar to the `encoding` though and just allow those attributes to be set for a client object.

bennybp:
> If adding another dependency seems overkill for this, it's not too hard to implement your own solution

Let me think about just hardcoding a solution. We are only using it in one place, so it it's not be too bad to implement it myself (using the aforementioned class attributes)

bennybp:
Ok implemented by hand now. It's pretty simple :)

There is only one other place I would consider adding this kind of logic (when dealing with certain database operations). But it will end up looking a lot different, so I'm not worried about duplicating code.

dotsdl:
Giving this another review today!

qcportal/qcportal/manager_client.py
dotsdl: Ah cool, thanks! Makes sense here in testing; in production probably helps some with resilience for managers if the server blips for a bit.

dotsdl (approved):
Looks great @bennybp! This should help users quite a bit!

Issue #756, opened 2023-09-19
Creating a fast way to get dataset size


j-wags:
I've recently needed to look at datasets by size (I'm looking for a small one to integrate into tests). Right now the best way to do it is 

```python
dss = client.list_datasets()
for ds_dict in dss:
    ds = client.get_dataset('singlepoint', ds_dict['dataset_name'])
    print(len([*ds.iterate_records()]), ds_dict['dataset_name'])
```

> 937 OpenFF Optimization Set 1
> 48280 OpenFF VEHICLe Set 1
> 189 OpenFF NCI250K Boron 1
> ...

This eventually works, but it takes a long time. I'd love a quick way to directly query the size (for example, the number of records) in a dataset. 

This isn't horribly urgent - The current method works fine and I only need to do it once - but I can imagine that many new users will stumble over this in the future.

bennybp:
Fixed in #762

PR #762, opened 2023-10-02
Add display of number of records in a dataset

bennybp:
## Description
Adds the number of records in a dataset to the `list_datasets` functionality, and also the ability to request from the server the number of records in a dataset via the dataset model (`ds.record_count`). Requested in issue #756

The record count is computed on the fly. For now, this is fine, but we will have to keep an eye on it as the database size increases, though, as then listing dataset size could take too long. There could be more database trickery we could do.

Example output from `print_dataset_list` (for a small subset of the production database):

```
 id  type                record_count  name
----  ----------------  --------------  -----------------------------------
  48  torsiondrive                   3  SMIRNOFF Coverage Torsion Set 1
  50  optimization                   3  OpenFF Discrepancy Benchmark 1
 155  singlepoint                   24  QM9
 159  singlepoint                    3  OpenFF Discrepancy Benchmark 1
 184  manybody                       0  S22
 207  singlepoint                   24  COMP6 DrugBank
 237  gridoptimization               5  OpenFF Trivalent Nitrogen Set 2
 253  optimization                   3  OpenFF Gen 2 Opt Set 2 Coverage
 309  torsiondrive                  39  OpenFF Theory Benchmarking Set v1.0
```

## Changelog description

Add display of number of records in a dataset, and ability to get number of records in a dataset

## Status
- [X] Code base linted
- [X] Ready to go

### DIFF
--- a/qcfractal/qcfractal/components/dataset_routes.py
+++ b/qcfractal/qcfractal/components/dataset_routes.py
@@ -136,6 +136,13 @@ def get_dataset_detailed_status_v1(dataset_type: str, dataset_id: int):
     return ds_socket.detailed_status(dataset_id)
 
 
+@api_v1.route("/datasets/<string:dataset_type>/<int:dataset_id>/record_count", methods=["GET"])
+@wrap_route("READ")
+def get_dataset_record_count_v1(dataset_type: str, dataset_id: int):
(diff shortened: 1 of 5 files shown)


j-wags:
Awesome - this looks great! Sorry it took me so long, just got to this as you were merging. Thanks for putting this in!

PR #765, opened 2023-10-04
Improve compile_values and related functions

janash:
This PR makes some edits to the `compile_values` method and adds a new dataset method called `get_properties_df`.

The following changes are made to `compile_values`:

* `compile_values` now always returns a multi-level index where the specification is the top leve index.
* `compile_values` now has an option to unpack results from the callable, allowing more than one dataframe column to be built at a time.
* `entry_name` may be a string or list of strings, but now also has a default value in case the user doesn't provide it.

Example use:
```python
import qcportal as ptl

client = ptl.PortalClient("https://qcademo.molssi.org")
dataset = client.get_dataset_by_id(1)

# Case 1 - Single value returned from callable.
df1 = dataset.compile_values(lambda x: x.properties["scf_total_energy"],"scf_total_energy" )

# Case 2 - Tuple of two values returned from callable with unpack True
df2 = dataset.compile_values(lambda x: (x.properties["scf_total_energy"], x.properties["scf_iterations"]), unpack=True)
```

The function `get_properties_df` is added to allow easier compilation of record properties into a dataframe. The function takes a list of properties and returns a multi-level index dataframe similar to the type returned from `compile_values`. Under the hood, the function uses `compile_values`  accessing the `properties` using `.get` in case property doesn't exist for a particular specification. Additionally, before the df is returned to the user any columns containing all `nan` are dropped.

Example use (continuing from example above)
```python
properties = dataset.get_properties_df(["scf_total_energy", "scf dipole", "mp2 dipole"])
```

### DIFF
--- a/qcportal/qcportal/dataset_models.py
+++ b/qcportal/qcportal/dataset_models.py
@@ -1119,19 +1119,61 @@ class BaseDataset(BaseModel):
         value_names: Union[List[str], str] = "value",
         entry_names: Optional[Union[str, Iterable[str]]] = None,
         specification_names: Optional[Union[str, Iterable[str]]] = None,
-        include: Optional[Iterable[str]] = None,
-        fetch_updated: bool = True,
-        force_refetch: bool = False,
         unpack: bool = False,
     ) -> pd.DataFrame:
+        """
+        Compile values from records into a pandas DataFrame.
+
+        Parameters
+        -----------
+        value_call : Callable
+            Function to call on each record to extract the desired value. Must return a scalar value or 
+            a sequence of values if 'unpack' is set to True.
+
+        value_names : Union[List[str], str], optional
+            Column name(s) for the extracted value(s). If a string is provided and multiple values are 
+            returned by 'value_call', columns are named by appending an index to this string. If a list 
+            of strings is provided, it must match the length of the sequence returned by 'value_call'.
+            Default is "value".
+
+        entry_names : Optional[Union[str, Iterable[str]]], optional
+            Entry names to filter records. If not provided, considers all entries.
+
+        specification_names : Optional[Union[str, Iterable[str]]], optional
+            Specification names to filter records. If not provided, considers all specifications.
+
+        unpack : bool, optional
+            If True, unpack the sequence of values returned by 'value_call' into separate columns.
+            Default is False.
+
+        Returns
+        --------
+        pd.DataFrame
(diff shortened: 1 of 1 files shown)


qcportal/qcportal/dataset_models.py
```diff
@@ -1118,27 +1116,130 @@ def uninvalidate_records(
     def compile_values(
         self,
         value_call: Callable,
-        value_name: str,
+        value_names: Union[List[str], str] = "value",
```
bennybp, line 1119: Having the type be `Union[Iterable[str], str]` would be more generic :)

qcportal/qcportal/dataset_models.py
```diff
@@ -1118,27 +1116,130 @@ def uninvalidate_records(
...
+        else:
+            column_names = [value_names]
+            df = pd.DataFrame(_data_generator(), columns=("entry", "specification", value_names))
+
+        return_val = df.pivot(index="entry", columns="specification", values=column_names)
+
+        # Make specification top level index.
+        return return_val.swaplevel(axis=1)
+
+    def get_properties_df(self, properties_list: List[str]) -> pd.DataFrame:
```
bennybp, line 1212: Also should be `Iterable[str]`

qcportal/qcportal/dataset_models.py
```diff
@@ -1118,27 +1116,130 @@ def uninvalidate_records(
...
-        force_refetch: bool = False,
+        unpack: bool = False,
     ) -> pd.DataFrame:
-        def _data_generator():
+        """
+        Compile values from records into a pandas DataFrame.
+
+        Parameters
+        -----------
+        value_call : Callable
```
bennybp, line 1129: We don't use the types in the docstrings, since sphinx can now get the types from the function arguments itself.

bennybp (commented):
Just some cosmetic changes that make it consistent with everything else. Otherwise good job! :)

qcportal/qcportal/dataset_models.py
bennybp: Actually, `Sequence`, not `Iterable`.

qcportal/qcportal/dataset_models.py
bennybp: `Sequence`, not `Iterable`, since order matters

bennybp:
Looks good! Thanks!

PR #773, opened 2023-10-14
Improve manager logging of task information

bennybp:
## Description
Improve printing of task status and information in compute manager logs

Information now is printed as tables and now include `record_id`, which is distinct from `task_id`.
In general, `task_id` is rarely useful, and often confused with `record_id`, so we print both explicitly.

Now, two tables are printed. The first is when computations are found to be finished, and includes executor, walltime, and error information.

```
  task id    record id  executor    walltime    status
---------  -----------  ----------  ----------  --------------------
        3            4  local       00:00:2.74  success
        4            5  local       00:00:2.74  success
        5            6  local       00:00:2.52  error: unknown_error
```

The second displays the status of records being sent back to the server (whether they are accepted, rejected, or deferred due to server downtime/network issues).

```
  task id    record id  status    reason
---------  -----------  --------  -------------------------------------
        3            4  sent
        4            5  sent
        5            6  rejected  Task does not exist in the task queue
```

## Changelog description
Improve manager logging of task information

## Status
- [X] Code base linted
- [x] Ready to go

### DIFF
--- a/qcfractalcompute/qcfractalcompute/compute_manager.py
+++ b/qcfractalcompute/qcfractalcompute/compute_manager.py
@@ -8,9 +8,10 @@ import time
 import traceback
 import uuid
 from collections import defaultdict
-from typing import TYPE_CHECKING, Dict, List
+from typing import TYPE_CHECKING, Dict, List, Tuple
 
 import parsl.executors.high_throughput.interchange
+import tabulate
 from parsl.config import Config as ParslConfig
 from parsl.dataflow.dflow import DataFlowKernel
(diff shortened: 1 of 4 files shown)


bennybp:
Just wondering if you have any comments @j-wags or @dotsdl. Any additional info you would find useful?
You are some of the few people to run a manager :)

j-wags:
This looks really slick! I'd be happy with it. I'll ping @dotsdl directly in case he (like me) didn't see the notification.

dotsdl (approved):
Looks great @bennybp! Thanks for adding these!

Issue #740, opened 2023-09-05
[Next] Local data persistence


chrisiacovella:
This issue is to sketch out some ideas and start a discussion related to retrieving and saving datasets.  This follows from some prior discussion during working group meetings. 

### Local caching of records:  

When accessing records from the archive, it would be very helpful to be able to store this data locally in a cache.  It seems like this could come in two distinct flavors: 

#### Automatic caching.  
I'm looking at the current source code and  there appears to be some framework already in place (but maybe not yet implemented?)  that relies upon DBM for the automatic caching.  If implemented, this would allow QCPortal to check the local cache to see if a given record from a specified server has already been retrieved, and if so, use the local version. This would certainly be very beneficial since it would mean that for many users,  rerunning a python script or restarting a notebook kernel would not require re-downloading data.  However, the actual performance will depend upon the amount of memory allocated to the cache and the size of a given dataset a user is working with.    

#### User-defined caching.  

This would povide the same basic functionality as the automatic cacheing, but allowing a user to define the location to store a database, where by default, the cache does not have a maximum size limit.  This would be beneficial to users that are working with, say, entire datasets.  For example, if say, working with the QM9 dataset, I would only like to basically download the records once and be able to store them locally for ease of access later; I don't want to have to worry about the dataset records being purged (due to downloading other data from qcarchive) or just simply having the dataset being larger than the default memory allocation.  In my own work, I've implemented a simple wrapper around the calls to QCPortal where each record is saved into an SQLdict database and this has been very helpful, especially in cases where I lose connection to the database.

### Ability to download entire datasets:  

Some of the datasets in the older version included HDF5 files (that could be downloaded either via the portal or from zenodo).  This allowed an entire dataset to be downloaded very efficiently.  As an example, it would take about 5 minutes to download QM9 in the hdf5 format (~160 mb when gzipped) for ~133K records;  fetching these records one at a time (using the new code) took > 12 hours.  Having a way to download an entire dataset in one file would be very helpful. 
 - A few related notes:  to even download this required wrapping the calls to qcportal in  try/except statements to automatically reconnect, as periodically I would lose connection.  I was able to speed this up to about 15 minutes by using concurrent.futures to multithread the per-record fetches.  If a single downloaded file is not possible for a dataset (e.g., given that datasets may be changing), it would be good to have some way to efficiently download all the records at a single time, saved to some local cached database.  Either way, calling "get_record" in a loop at this point is not very efficient and could be a big stumbling block for users.

[...]

PR #802, opened 2024-01-12
Implement client-side dataset caching

bennybp:
## Description
Previously, dataset information was not cached locally at all. So rerunning a script, or just calling `client.get_dataset` again could require re-fetching all data, even if it had been fetched before.

This PR implements this caching. All storage of records is now in an SQLite database, either in a file or in memory. Some care has been taken in trying to keep the cache up-to-date as much as possible, but I am sure there are still loopholes. This includes records writing themselves back to the cache when they have been updated with additional data (for example, fetching molecules or trajectories).

There are a few ways to use:
* Set the `cache_dir` parameter when creating a client. This will then automatically create SQLite files for each dataset, and re-use them as long as the same `cache_dir` is used in subsequent client construction.
* The PortalClient has a `dataset_from_cache` function where you can pass a file directly (ie, downloaded out-of-band).
* There is a free `dataset_from_cache` function in `dataset_models.py` that works similarly, but will result in an offline dataset object completely disconnected from any server.

This is purely a client-side change, so this branch will work with the currently-deployed MolSSI QCArchive servers.

There is still some polishing to be done (and docs), but I am looking for feedback and any bugs before merging.

See #740 

Todos and missing features:
- [ ] `refresh_cache` needs to be finished
- [ ] Function to delete the cache(s)
- [ ] Any sort of size restriction is not obeyed
- [ ] Docs, of course

## Changelog description

Implement client-side caching of datasets

## Status
- [x] Code base linted
- [x] Ready to go


bennybp:
After testing this, the way records are cached (with all their child records) is not going to work. It works for singlepoints, but torsiondrives are way too big. So this is going to need a bit more work

I think the solution is to store individual records (without children) in a separate table, and store foreign keys in the current `record_data` table. Probably need to rename some of these to something like `dataset_records` or something.

j-wags:
I've been playing around with this today, and it's great! Very intuitive. Two comments along with their importance out of 10 (I wouldn't consider either one blocking).

* (5/10) "cache" makes me think that this will start overwriting itself in some conditions. I'd like to have a mode that's like "I have infinite disk space, don't limit the cache size". Is this the way it works currently/could this mode be added? If not, could the behavior be documented?
* (2/10) It's a little incongruous that the API has me specify `cache_dir`, but then to load things using `dataset_from_cache` I need to provide a file path _inside_ that cache dir, with no API point that tells me the path (as in, I have to go to my file system and `ls` around to get the path to the cache file).


The code that I'm playing with is:

```python
import qcportal

client = qcportal.PortalClient("https://api.qcarchive.molssi.org:443", cache_dir="./cache2")
ds = client.get_dataset("torsiondrive", "XtalPi Shared Fragments TorsiondriveDataset v1.0")

# the next two lines didn't immediately do what I wanted, so I ran the loops below
#ds.fetch_entries()
#ds.fetch_records(include=["optimizations"], force_refetch=True) 

for entry in ds.iterate_entries():
    entry
for record in ds.iterate_records():
    for angle, opt in record[2].minimum_optimizations.items():
        opt.final_molecule
```

Then in a separate interpreter:
```python
from qcportal import dataset_models
ds2 = dataset_models.dataset_from_cache("./cache2/api.qcarchive.molssi.org_443/dataset_378.sqlite")
from openff.qcsubmit.results import TorsionDriveResultCollection
tdrc = TorsionDriveResultCollection.from_datasets([ds2])
tdrc
```

> `TorsionDriveResultCollection(entries={'local': [TorsionDriveResult(type='torsion', record_id=119138412, cmiles='[H:12][C:1](=[C:3]1[C:4](=[O:9])[N:8]([C@@:6]([C:5](=[O:10])[N:7]1[H:18])([H:17])[O:11][C:2]([H:14])([H:15])[H:16])[H:19])[H:13]', inchi_key='BFHIBVQMZBLHGM-OCSBBNMYNA-N'), TorsionDriveResult(type='torsion', ...`

Success!!

bennybp:
Glad it's working so far!

>     * (5/10) "cache" makes me think that this will start overwriting itself in some conditions. I'd like to have a mode that's like "I have infinite disk space, don't limit the cache size". Is this the way it works currently/could this mode be added? If not, could the behavior be documented?

At the moment, there is no limit to the cache (basically if you set the max size to `None`). I haven't added the logic for finite cache sizes yet. But yes, docs always need to be written.

>     * (2/10) It's a little incongruous that the API has me specify `cache_dir`, but then to load things using `dataset_from_cache` I need to provide a file path _inside_ that cache dir, with no API point that tells me the path (as in, I have to go to my file system and `ls` around to get the path to the cache file).

If you create a client with the same `cache_dir` then it should automatically find the existing cache files when you use `get_dataset`. At least that is the intent. The `dataset_from_cache` is more for if you want to pass around/download the cache file separately (kind of like the old 'views').

bennybp:
I'm going to go ahead and merge this. There's still some tasks to be done before the next release, but it seems to be working well. 

The main reason is I have another feature being built on top of this and leaving this open makes it a bit complicated.

PR #813, opened 2024-04-07
Fix unique index on base dataset

bennybp:
## Description

Fixes the unique index on the base dataset table.

For some reason, some instances of QCFractal are missing this index, while others have it. This migration should be safe in both instances

## Changelog description

Fixes unique index in base dataset table

## Status
- [X] Code base linted
- [X] Ready to go


bennybp:
Python 3.9 failures are due to networkx pins on conda-forge packages being incorrect. The newest version of networkx is actually incompatible with python 3.9, but that's not reflected in the c-f recipe: https://github.com/conda-forge/networkx-feedstock/pull/57

PR #827, opened 2024-05-01
Lazy load zstandard

mattwthompson:
## Description

This PR defers `import zstandard` until it's needed. Some [weird packaging issues](https://github.com/conda-forge/qcfractal-feedstock/issues/43) cause a conflict that prevents any QCPortal import. I'm not sure when these compression/caching modules are used, i.e. by common user operations or only by admins. If they're ever not needed, this avoids an unnecessary crash.

## Changelog description

Lazy load `zstandard`

## Status
- [x] Code base linted
- [x] Ready to go

### DIFF
--- a/qcportal/qcportal/cache.py
+++ b/qcportal/qcportal/cache.py
@@ -20,7 +20,6 @@ try:
 except ImportError:
     import pydantic
     from pydantic import BaseModel, Extra, validator, PrivateAttr, Field
-import zstandard
 
 from .serialization import serialize, deserialize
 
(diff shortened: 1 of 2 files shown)


bennybp:
Playing around with the zstd stuff, I see they have an environment variable for how the backend gets loaded. It seems to help in my case. https://python-zstandard.readthedocs.io/en/latest/api_usage.html

```bash
$ PYTHON_ZSTANDARD_IMPORT_POLICY='cffi_fallback' python3 -c 'from openeye import oechem; import zstandard;print("ok")'
ok
```

So openeye bundles a zstd dynamic library with it? I see a `./libs/python3-linux-x64-g++10.x/libzstd.so.1` in the openeye package.  I think using `cffi` would then make `zstandard` imports use the installed library that's part of the `zstd` package, but I'm not positive and no idea if that would cause problems.

mattwthompson:
I wish I found that a while back - that's exactly what I need!

PR #854, opened 2024-10-30
Extras field of records should not be None/NULL

bennybp:
## Description

The `extras` field of `base_record` (therefore all records) should probably not be null. It's more consistent that it's a dict, even if it's empty.

The original idea is that `extras` is only populated on record completion. But
  * At some point, we might make these settable by the submitter
  * We may eventually move to an EAV table, which would always result in an empty dict anyway

This inconsistency has caused problems for some users as well: https://github.com/openforcefield/openff-bespokefit/issues/369 https://github.com/openforcefield/openff-bespokefit/pull/370

Relatively easy fix - make the db column not null, and the pydantic field not optional. A small validator is added to halp newer clients use older servers.

This is backwards compatible - new clients can use old servers, and old clients can still use new servers.

## Status
- [X] Code base linted
- [X] Ready to go


j-wags:
Awesome, thanks Ben!

PR #855, opened 2024-11-08
Lazy-load pandas

mattwthompson:
## Description

Currently, a little less than half of the time taken to import anything from QCPortal is spend dealing with Pandas. I think most use cases don't involve [its direct use](https://github.com/search?q=repo%3AMolSSI%2FQCFractal%20import%20pandas&type=code), so it can be loaded later when needed. This improvement should propagate to downstream packages which want to i.e. use `PortalClient` or build off of record models but don't need Pandas.

Before:

```shell
$ hyperfine \                                                                    14:31:18  ☁  4ebcd88f ☀
    'python -c "from qcportal import PortalClient"' \
    --prepare \
    'python -c "from qcportal import PortalClient"'
Benchmark 1: python -c "from qcportal import PortalClient"
  Time (mean ± σ):     621.8 ms ±  10.0 ms    [User: 514.2 ms, System: 96.5 ms]
  Range (min … max):   608.4 ms … 642.9 ms    10 runs

```

```shell
$ python -X importtime -c "from qcportal import PortalClient" 2> tuna.log && tuna tuna.log
```

image

With these changes:

```shell
$ fc                                                                     14:27:45  ☁  lazy-load-pandas ☀
hyperfine \
    'python -c "from qcportal import PortalClient"' \
    --prepare \
    'python -c "from qcportal import PortalClient"'
Benchmark 1: python -c "from qcportal import PortalClient"
  Time (mean ± σ):     375.1 ms ±   7.5 ms    [User: 318.7 ms, System: 50.1 ms]
  Range (min … max):   368.1 ms … 394.3 ms    10 runs
```
## Status
- [ ] Code base linted
- [ ] Ready to go

### DIFF
--- a/qcportal/qcportal/dataset_models.py
+++ b/qcportal/qcportal/dataset_models.py
@@ -19,8 +19,6 @@ from typing import (
     Mapping,
 )
 
-import pandas as pd
-
 try:
     import pydantic.v1 as pydantic
     from pydantic.v1 import BaseModel, Extra, validator, PrivateAttr, Field
@@ -40,6 +38,7 @@ from qcportal.cache import DatasetCache, read_dataset_metadata, get_records_with
 
 if TYPE_CHECKING:
     from qcportal.client import PortalClient
+    from pandas import DataFrame
 
 
 class Citation(BaseModel):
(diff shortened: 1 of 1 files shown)


bennybp:
Sounds good to me.

I think the CI errors are from conda resolving some older pytest version with python 3.8. Since python 3.8 is now EOL, I can probably take that out of the CI. Would you agree?

mattwthompson:
Sure, I've made an attempt in https://github.com/MolSSI/QCFractal/pull/856

bennybp:
Thanks! This is great, I'm always looking for performance increases

PR #856, opened 2024-11-12
Drop Python 3.8

mattwthompson:
## Description

This PR drops Python 3.8, which is past its EOL, from testing.

## Status
- [ ] Code base linted
- [ ] Ready to go


bennybp:
Thanks! That should be all that's needed.

I think it would still technically work with 3.8 but we just can't waste time testing it

PR #860, opened 2024-12-05
Adding entry attribute modification.

sjayellis:
## Description
This PR aims to add the ability to modify the attributes of an entry within a dataset.
Similar to renaming an entry, this takes a mapping of entry names to the updated attribute values. The current implementation does a full replacement of the attributes of an entry.

## Status
- [x] Code base linted
- [ ] Ready to go


bennybp (approved):
Looks great! Thanks!

PR #882, opened 2025-01-16
Qcportal jupyterhub

sjayellis:
## Description
This PR adds `Dockerfile.jupyterhub`, a dockerfile for hosting a Jupyterhub with QCPortal, along with a configuration file and a custom authenticator that uses QCArchive to authenticate a user.

Users have directories created in home, so it needs to be mounted to allow any created notebooks to persist.

## Status
- [x] Code base linted
- [ ] Ready to go

### DIFF
--- a/Dockerfile.jupyterhub
+++ b/Dockerfile.jupyterhub
@@ -1,0 +1,21 @@
+FROM ubuntu:22.04
+
+ENV DEBIAN_FRONTEND=noninteractive
+
+EXPOSE 8000
+
+RUN apt-get upgrade -y && apt-get update -y && apt-get install -yqq --no-install-recommends python3-pip && pip3 install --upgrade pip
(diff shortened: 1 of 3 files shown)


bennybp:
I've got it all up and running on our server, and it seems to be basically working well. A few things to improve on (not urgently, though). Otherwise, great work!

- [ ] Error handling - incorrect username/password results in 500 Server error
- [ ] Handling of admin users (if the user's role is admin on the server, it should be admin on the hub)
- [ ] Environment variables not set in the spawned servers
- [ ] Possible to change the branding? I seem to remember this being kinda sketchy

bennybp:
I've got some more ideas for this, but I will build off of this in a separate branch.

Now that the qcfractal server has proper session cookie support, it might be easier to use those (among other things)

PR #909, opened 2025-03-13
Make encode_to_json work recursively

bennybp:
## Description

Improves the `encode_to_json` function in `qcportal.serialization` so that it works recursively. This will allow for converting entries and specifications to plain python objects suitable for later `json.dump`.

@jaclark5 

## Status
- [X] Code base linted
- [X] Ready to go

### DIFF
--- a/qcportal/qcportal/serialization.py
+++ b/qcportal/qcportal/serialization.py
@@ -36,7 +36,7 @@ def _msgpack_decode(obj: Any) -> Any:
     return obj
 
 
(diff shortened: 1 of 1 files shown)


jaclark5:
I decided that I like `encode_to_json` better because I can format the json serialization with the indentation level that I want.

bennybp:
> I decided that I like `encode_to_json` better because I can format the json serialization with the indentation level that I want.

There is also the possibility of passing `qcportal.serialization._JSONEncoder` to a plain `json.dump`. But pretty much all of this does the same thing :)

PR #910, opened 2025-03-13
Add external scaffold to/from json

jaclark5:
## Description
This module will allow users to write and read a json file that can be reformed into an initial dataset **before** submission. This means that records are not present.

## Status
- [X] Code base linted
- [x] Ready to go


qcportal/qcportal/external/scaffold.py
```diff
@@ -0,0 +1,94 @@
...
+    elif extension[-1] == "bz2" and extension[-2] == "json":
+        with bz2.open(filename, "rt", encoding="utf-8") as f:
+            ds_dict = json.load(f)
+    else:
+        raise ValueError(f"File extension must be json or json.bz2, not {extension[-2]}.{extension[-1]}")
+
+    ds = client.add_dataset(**ds_dict["metadata"])
+
+    for _, entry in ds_dict["entries"].items():
+        ds.add_entry(**entry)
```
bennybp, line 89: Kind of surprised this works. I guess it's just from keeping names consistent :)

It would be safer (and faster?) to build entry object(s) and use ds.add_entries(obj_list).

Every dataset has it's own entry type, but you can get it with `ds._entry_type`. So this might work (going from memory)

```
entry_type = ds._entry_type
....
new_entry = entry_type(**entry)
```

I think specifications will be ok to do your current way, though. Those are consistent across dataset types.

jaclark5: The `add_entry` seems to wrap each entry type appropriately. You're right, it cuts the time in half to use.

jaclark5:
Blocked by #915 and awaiting new `background_add_entries`.

jaclark5:
@bennybp I implemented `background_add_entries` and my tests passed! Ready for review!

jaclark5:
I forgot that I used deepdiff to ensure the datasets were the same, what do you think if I were to add that to the testing dependencies?

bennybp:
Thanks! This will be a great feature. I just added the deepdiff deps to the conda envs (we might want to use it other places as well).

PR #939, opened 2025-04-29
Add getting dataset status by compute tag

bennybp:
## Description

Adds a function (client & server) to get the status of a dataset grouped by tag.

Tested with some larger datasets on a test server and it seems fast enough (typically less than 3 seconds or so)

(if there's any more, I imagine we should make one general function)

Requested by @jaclark5 

## Status
- [X] Code base linted
- [X] Ready to go

### DIFF
--- a/qcfractal/qcfractal/components/dataset_routes.py
+++ b/qcfractal/qcfractal/components/dataset_routes.py
@@ -138,6 +138,13 @@ def get_dataset_detailed_status_v1(dataset_type: str, dataset_id: int):
     return ds_socket.detailed_status(dataset_id)
 
 
+@api_v1.route("/datasets/<string:dataset_type>/<int:dataset_id>/status_by_tag", methods=["GET"])
+@wrap_global_route("datasets", "read")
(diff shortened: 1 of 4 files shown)


jaclark5:
This looks great! It seems I can't test it live until it's merged and on the servers?

So it would return something like:
```python
response = [
    ("tag1", "running", 5),
    ("tag1", "waiting", 400),
    ("tag2", "error", 300),
]
```

Would it be possible for it to return the record ids in each tuple? Either that or could `ds.iterate_records()` have a keyword for compute_tag?

bennybp:
The former might be better. But do you want record ids or entry/specification strings?

jaclark5:
Probably the entry/specification since it's faster to search by entries, it just thought record ids would be easier since the former would result in a tuple like `(tag, status, count, [(entry1, spec1), (entry2, spec2),....)])`.

jaclark5:
The next step I would want to take with this information is get Molecule properties from the difference categories. Having entries would be easiest since it's fast to call those from the dataset. If I had record ids I'd have to call them through the client to get entry names and then call those from the dataset, so entry names and specs are probably better for what I'm looking for.

PR #951, opened 2025-06-10
Add ability to append a dataset with scaffold

jaclark5:
## Description
This PR will detect if a dataset already exists and append it.

## Status
- [x] Code base linted
- [x] Ready to go

### DIFF
--- a/qcportal/qcportal/external/scaffold.py
+++ b/qcportal/qcportal/external/scaffold.py
@@ -10,6 +10,7 @@ This module will allow users to write and read a json file that can be reformed 
 
 import json
 import bz2
+import warnings
 
 from ..serialization import encode_to_json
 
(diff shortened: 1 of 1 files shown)


qcportal/qcportal/external/scaffold.py
```diff
@@ -81,17 +84,30 @@ def from_json(filename, client):
...
 
     entries = []
     entry_type = ds._entry_type
-    for _, entry in ds_dict["entries"].items():
+    for entry_name, entry in ds_dict["entries"].items():
         if "local_results" in entry:
             del entry["local_results"]
-        entries.append(entry_type(**entry))
+            
+        if entry_name in ds.entry_names:
```
bennybp, line 106: I think this should be `if entry_name not in ds.entry_names` ? I was just hunting down some pytest warnings :)

PR #952, opened 2025-06-29
Use spawn rather than fork for multiprocessing

bennybp:
## Description

I had been using `fork` due to its speed, however warnings are starting to pop up due to the use of mixing threads with fork-based multiprocessing.

So this changes the multiprocessing contexts to spawn. This also moves the functions targeted by `multiprocessing.Process` to a separate module so that they can be reused by the CLI code.

Tests are all passing locally, although with a bit of slowdown.

## Status
- [X] Code base linted
- [X] Ready to go

### DIFF
--- a/.github/workflows/core_tests.yml
+++ b/.github/workflows/core_tests.yml
@@ -53,5 +53,7 @@ jobs:
         run: |
           pytest --runslow --client-encoding="${{ matrix.encoding }}" \
                  --log-level=DEBUG \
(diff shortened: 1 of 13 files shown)


bennybp:
This seems to have surfaced some inefficiencies. Since `spawn` starts a new interpreter, all imports and stuff (adding default internal jobs, etc) have to be done again. Closing for now while I work out some improvements there.

PR #967, opened 2025-07-23
Change how child records are stored (don't use PrivateAttr)

bennybp:
## Description

This changes the way child records are stored internally inside records. Before, they were stored as PrivateAttr, which means they were never serialized if you used pydantic's `dict()` function.

However, for saving/transferring/ingesting complete records, we want those child records in place. So this PR makes those child records regular fields.

However, we DON'T want those child records when writing to cache files (where they are handled differently). So a separate function was added for that.

## Status
- [X] Code base linted
- [X] Ready to go

### DIFF
--- a/qcfractal/qcfractal/components/optimization/test_record_client.py
+++ b/qcfractal/qcfractal/components/optimization/test_record_client.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import itertools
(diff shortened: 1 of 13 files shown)


bennybp:
I've given this a test with reading existing dataset view files, and it works for me. So as far as I can tell this is backwards compatible (as long as you are using the public API).

PR #1004, opened 2026-05-27
Pydantic v2 and Python 3.14 Compatibility

bennybp:
## Description

At long last, full pydantic v2 compatibity through the whole qcarchive stack. Pretty much every file is touched, but changes are straightforward. Goal is complete backwards compatibility, but keep an eye out for bugs.

We will still use qcschema v1 from qcelemental until v2 of the schema is finalized. For the most part, that would only affect compute workers when that change is made.

I've also moved towards some more modern python constructs (using `|` for unions of types rather than `Union`). But that's not universal yet.

Tests pass locally, but I need to take one more pass at how molecules are handled server-side, as that's a sensitive area.

## Status
- [ ] Code base linted
- [ ] Ready to go


qcarchivetesting/conda-envs/fulltest_server.yaml
```diff
@@ -17,7 +17,7 @@ dependencies:
   - pydantic-settings
   - zstandard
   - apsw>=3.42
-  - qcelemental>=0.30,<0.70a0
+  - qcelemental>=0.50.1,<0.70a0
```
loriab:
```suggestion
  - qcelemental>=0.50.2,<0.70a0
```

qcarchivetesting/conda-envs/fulltest_server.yaml
```diff
@@ -39,5 +39,5 @@ dependencies:
 
   # QCFractal Services
   - torsiondrive
-  - qcmanybody
+  - qcmanybody>=0.7.1
```
loriab, line 42: in conda, qcel will impose the qcmb constraint

qcfractal/qcfractal/components/manybody/record_socket.py
```diff
@@ -163,7 +163,8 @@ def initialize_service(self, session: Session, service_orm: ServiceQueueORM) ->
         table_rows = []
 
         for mol_batch in chunk_iterable(qcm.iterate_molecules(), 400):
-            to_add = [x[2] for x in mol_batch]
+            # TODO - remove after v2 manybody
```
loriab, line 166: ready to delete or does this refer to the qcf manybody service

PR #1009, opened 2026-06-01
qcfractalcompute: tolerate non-JSON stdout from xtb in run_conda_subp…

stvogt:
xtb on systems with near-degenerate frontier orbitals prints
non-JSON output (STDA-like excitation analysis) to stdout BEFORE
the qcengine result JSON. Scan back for the last well-formed
 JSON object. Remove when upstream merges the fix

## Status
- [ ] Code base linted
- [ ] Ready to go


bennybp:
I wonder if wrapping qcengine.compute/compute_procedure in qcengine_compute.py with redirect_stdout could solve this once and for all?   I think this issue pops up occasionally 

Edit: You might be able to use this: https://github.com/MolSSI/QCFractal/blob/63f62e5b1648643536076e0902fee01fadf8b63b/qcportal/qcportal/utils.py#L389

like so:

https://github.com/MolSSI/QCFractal/blob/63f62e5b1648643536076e0902fee01fadf8b63b/qcfractal/qcfractal/components/neb/record_socket.py#L198

Issue #1017, opened 2026-06-12
List of roles


jnkrakowski:
please update your docs to have a list of user roles

PR #1019, opened 2026-06-26
Added roles to the docs

cgbriggs99:
## Description

The roles were missing from documentation. I added them to make it easier to find. Previously, users would have to look in the configuration file for the role permissions to find which roles they were allowed to use. Solves #1017 

## Status
- [x] Code base linted
- [x] Ready to go


bennybp:
Thanks! I am working on various docs now so we can incorporate this (although these docs will change).

